PR #163, opened 2014-12-07
Switch from obsolete PIL to Pillow fork

nirgal:
I'm really not sure about that change. Does it work for you?

Here, enable-4.3.0.egg-info/requires.txt gets "PIL" in it, but I do not have any PIL.egginfo, only a Pillow-2.6.1.egg-info.


corranwebster:
Closing this, because this is well and truly resolved by #154 at this point. Thanks @nirgal for the work.

PR #169, opened 2015-02-06
[WIP] Fixes and updates for wxPython 3.0.x compatibility

RobinD42:


jwiggins:
Looking at that failure... Did Travis change underneath us again?

jwiggins:
So, this passes tests now that master has been fixed. Is this still a WIP @RobinD42?

RobinD42:
It's only a WIP in that there may be more wx3.0-related issues or enhancements that I haven't had time to find yet.  This particular change doesn't need to wait for that however.

itziakos:
travis-ci currently checks only on Qt not wx. I think that we should first add a build test using wx before we start with the transition.

corranwebster:
@RobinD42 Tests now run against Wx: it would be good to pull in the current master and then see how things stand, and possibly merge this if everything works OK.

RobinD42:
It looks like I deleted the wrong instance of my original enable repo, so I lost the branch.  I'll resubmit in a new PR.

PR #171, opened 2015-03-20
Fix Travis PIL Requirements

### LINKED ISSUE #170: Travis-ci is failing on master
I think that this needs to be fixed asap so that we can be sure that there no hidden/shadowed errors in PRs

corranwebster:
Use `pip` without dependencies to avoid Pillow install.

Hopefully fixes #170


jwiggins:
This fixes the PIL problem, but now appears to be having a Pyglet problem on Python 2.6... :disappointed:

jwiggins:
:+1: Thanks for the fix!

PR #172, opened 2015-03-20
Force usage of Pyglet 1.1.4

jwiggins:
This may or may not work. I'm just opening a PR to force Travis into action.


jwiggins:
It worked! :tada:

PR #168, opened 2015-01-20
Simple Button Tool

corranwebster:
Basically a common tool that can be used whenever we want something to be "clickable" or "togglable".  Probably of significant utility in Chaco.

Has tests and documentation.

Ready for review.

PR #173, opened 2015-03-20
Replace multi-context with-statements with multiple with-statements

jwiggins:
This slipped in accidentally with #168


corranwebster:
LGTM

PR #177, opened 2015-03-25
fixes for recent swig

jjstickel:
Recent version of swig (3.0.5) broke the build of enable. This pull request fixes the issue. See https://github.com/swig/swig/issues/221#issuecomment-54469375

### DIFF
--- a/kiva/agg/src/affine_matrix.i
+++ b/kiva/agg/src/affine_matrix.i
@@ -157,7 +157,7 @@ namespace agg24
     };
 };
 
-%pythoncode {
+%pythoncode %{
 def is_sequence(arg):
     try:
         len(arg)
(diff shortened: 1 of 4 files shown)


itziakos:
:+1:

jwiggins:
Ahh... Thanks for the patch!

Unfortunately, Travis-CI is only telling us that it doesn't break on older versions of swig:
```
Setting up swig2.0 (2.0.4+really2.0.4-4ubuntu2) ...
Setting up swig (2.0.4+really2.0.4-4ubuntu2) ...
```

@itziakos : Did you test this?

itziakos:
Not locally, I thought that travis was checking that version. Sorry.

jwiggins:
Ahh... OK. I `grin`'d the current code. These four files are the only place where `%pythoncode` is being used in a slightly bad (but allowed by older swig) way. Based on that, there should be zero downside to merging this.

jwiggins:
Thanks @jjstickel!

PR #178, opened 2015-03-27
Cleanup setup and requirements

itziakos:
This PR tries to fix the requirements for kiva and enable and at the same time clean-up the travis-ci build

### DIFF
--- a/enable/__init__.py
+++ b/enable/__init__.py
@@ -7,6 +7,7 @@ import sys
 from ._version import full_version as __version__
 
 __requires__ = [
+    'traits',
     'traitsui',
     'PIL',
     'kiwisolver',
--- a/enable/__init__.py
+++ b/enable/__init__.py
@@ -4,8 +4,7 @@
     Part of the Enable project of the Enthought Tool Suite.
 """
 import sys
-
-from kiva._version import full_version as __version__
+from ._version import full_version as __version__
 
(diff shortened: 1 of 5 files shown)


itziakos:
Please note that now the PR tests take only 7 minutes instead of 12-14.

enable/__init__.py
```diff
@@ -3,10 +3,11 @@
 """ A multi-platform object drawing library.
     Part of the Enable project of the Enthought Tool Suite.
 """
-
-from kiva._version import full_version as __version__
+import sys
```
jwiggins: Not used here

jwiggins:
Other than that unused import I'm :+1: on this PR

jwiggins:
Go ahead and merge as soon as Travis gives the OK. I'm going offline for a bit.

itziakos:
Ok we now pass the setup.py import and we arrive at a new issue which now is the result of the interaction between setuptools and numpy.distutils. It looks that if any packages that is in the `install_requires` list it will result in an out  of memory error when that required packages are installed.

So, that is why commit 14e11492 fails, because it will cause the required packages to be downloaded by the combined setuptools/numpy.distutils setup.py. I have been able to replicate the behaviour locally.

itziakos:
So it looks that the last version of setuptools that work is 9.1

PR #181, opened 2015-04-01
Update gh-pages to 4.5.0

itziakos:


itziakos:
@corranwebster , @jwiggins, can you have a quick look?

corranwebster:
Looks reasonable to me.

What was the reason for adding to the `.gitignore`?

itziakos:
When switching between the branch where the documentation was build and the gh-pages branch. You end up with added folders in the status report. It is just confusing and it would be also wrong to commit these files by mistake.

jwiggins:
LGTM

PR #191, opened 2015-04-22
BUG: fix missing framework when building kiva.quartz.

cournape:
This is required to fix `kiva.quartz` when built on os x 10.7


itziakos:
:+1:

PR #195, opened 2015-05-08
Multiple build environments

itziakos:
This PRs adds support for multiple build environments.  The test coverage changes are not expected to change since there are not many gui toolkit specific tests


itziakos:
Added support for pillow. There is a strange problem with nose and expectedFailure decorated methods. Need to investigate.

itziakos:
This PR is ready for review.

travis_setup.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Simlinks for PIL compilation
+sudo ln -s /usr/lib/`uname -i`-linux-gnu/libfreetype.so /usr/lib/
+sudo ln -s /usr/lib/`uname -i`-linux-gnu/libjpeg.so /usr/lib/
+sudo ln -s /usr/lib/`uname -i`-linux-gnu/libpng.so /usr/lib/
+sudo ln -s /usr/lib/`uname -i`-linux-gnu/libz.so /usr/lib/
+
+# setup gui
+if [ "$ETS_TOOLKIT" = "wx" ]; then
+    echo "HERE"
```
jwiggins: I'm guessing this isn't needed any longer?

travis_setup.sh
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+# compile cairo
+wget -nv http://cairographics.org/releases/py2cairo-1.10.0.tar.bz2
+tar -xf py2cairo-1.10.0.tar.bz2
+cd py2cairo-1.10.0
+./waf configure
+./waf build
+sudo ./waf install
+cd ..
```
jwiggins: Use `pushd`/`popd` instead of `cd`

jwiggins:
Any reason why the setup script can't be named with a leading `.`? That's how the install scripts were named before. I noticed several commits where you kinda bounced around trying to find something that Travis would accept.

Other than some pretty minor feedback, I'm :+1: on this. Hopefully this will be a bit more stable than the earlier accommodations made for PIL/pillow. Oh, and actually testing WX is a nice touch too :wink:

itziakos:
> Any reason why the setup script can't be named with a leading .? That's how the install scripts were named before. I noticed several commits where you kinda bounced around trying to find something that Travis would accept.

it looked to me weird that the scripts where hidden, so I changed that.

itziakos:
> Other than some pretty minor feedback, I'm :+1: on this. Hopefully this will be a bit more stable than the earlier accommodations made for PIL/pillow. Oh, and actually testing WX is a nice touch too :wink:

In general, I think it is a better build setup. Please note however that it does not look that we are testing much or WX or QT code of enable. We need to increase the test coverage.

jwiggins:
WTF. `pushd`/`popd` doesn't work when you execute a script with `sh`? Perhaps `/bin/sh` is not bash on Travis?

WarrenWeckesser:
Ubuntu uses `dash` for `/bin/sh`, and `dash` does not has `pushd` and `popd`.

itziakos:
Ok, I reverted back, it is not that bad to use `cd`

jwiggins:
Thanks @itziakos! :shipit:

Issue #197, opened 2015-05-20
Installing  4.5.1 from pypi fails


itziakos:
Installing 4.5.1 using pip on an empty enviroment without traits, traitsui and other depedencies will fail with the following error

```python
pip install enable-4.5.1.tar.gz --user
Processing ./enable-4.5.1.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "/tmp/pip-l_EevW-build/setup.py", line 194, in <module>
        write_version_py(filename='enable/_version.py')
      File "/tmp/pip-l_EevW-build/setup.py", line 100, in write_version_py
        raise ImportError("Unable to import git_revision. Try removing "
    ImportError: Unable to import git_revision. Try removing kiva/_version.py and the build directory before building.
    
    ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-l_EevW-build
```

itziakos:
The bug is in the enable setup.py (https://github.com/enthought/enable/blob/master/setup.py#L95). We import kiva._version which is wrong it should be enable._version. since enable is the package that defined the version. 

The fix is simple, but I am currently looking for a workaround to help people until the new release arrives.

itziakos:
Without traits the following `ImportError` takes place and thus `import kiva._version` fails in the setup.py

```python
Traceback (most recent call last):
      File "kiva/__init__.py", line 21, in <module>
        from fonttools import Font
      File "kiva/fonttools/__init__.py", line 1, in <module>
        from font import Font, str_to_font
      File "kiva/fonttools/font.py", line 11, in <module>
        from .font_manager import FontProperties, fontManager
      File "kiva/fonttools/font_manager.py", line 64, in <module>
        from traits.etsconfig.api import ETSConfig
    ImportError: No module named traits.etsconfig.api
```
So in order to install the 4.5.1 version from pypi with pip on needs to make sure that `traits` is installed separatly and not as part of a requirements file.

itziakos:
fixed in PR #199

ajford:
I'm still seeing this issue when installing Enable 4.6.0 

```
$ pip install enable
Collecting enable
  Using cached enable-4.6.0.tar.bz2
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-f9rMgt/enable/setup.py", line 225, in <module>
[...]

PR #198, opened 2015-05-20
Fix issue 197

itziakos:
This PR fixes issue #197 by using `enable._version.py` as the file to look for version information when installing from source.


itziakos:
oops wrong branch

PR #202, opened 2015-06-11
Update kiva opengl tests

itziakos:
This PR removes the `expectedFailure` decorator from the `test_text_clip` method in the kiva opengl graphics context tests

### DIFF
--- a/kiva/tests/test_gl_drawing.py
+++ b/kiva/tests/test_gl_drawing.py
@@ -33,11 +33,6 @@ class TestGLDrawing(DrawingImageTester, unittest.TestCase):
         # FIXME: overriding test since it segfaults
         DrawingImageTester.test_star_clip(self)
 
-    @unittest.expectedFailure
-    def test_text_clip(self):
-        # gl graphics context does not clip text properly (#165).
-        DrawingImageTester.test_text_clip(self)
-
     @contextlib.contextmanager
     def draw_and_check(self):
         from pyglet.image.codecs.png import PNGImageEncoder


kiva/tests/test_gl_drawing.py
```diff
@@ -33,11 +33,6 @@ def test_star_clip(self):
         # FIXME: overriding test since it segfaults
         DrawingImageTester.test_star_clip(self)
 
-    @unittest.expectedFailure
-    def test_text_clip(self):
-        # gl graphics context does not clip text properly (#165).
-        DrawingImageTester.test_text_clip(self)
-
```
jwiggins: I think you need to skip this test instead of remove it from this derived class.

itziakos: I only skip tests that segfault. The strange thing is that it fails here but passes in #201

jwiggins: Indeed... Also, I'm guessing that it has passed once or twice in previous PRs too.

itziakos:
closing this PR, as not valid

jwiggins:
Does something still need to be done here to make the test more deterministic?

itziakos:
> Does something still need to be done here to make the test more deterministic?

The tests are very basic and thus not so robust. In most cases we check to see if anything `red` has been written in a file. However, fixing this is not easy and on the other hand the intermittent failure is not so common.

At the very least we need to create a new issue, so that the problem is known

PR #205, opened 2015-08-20
Fix scrolling problem with Qt native scrollbar

corranwebster:
Scrolling was in the reverse direction for vertical scrollbars (ie. when scrollbar was at top, view was at bottom and vice-versa).  This fixes this issue.

Testing is difficult for this - have just spent an unproductive hour trying to get something working.  We probably want a little set of utility routines.  I'll leave it up to the PR reviewer as to whether we should try to add tests before merging.

### DIFF
--- a/enable/tests/qt4/scrollbar_test.py
+++ b/enable/tests/qt4/scrollbar_test.py
@@ -2,11 +2,14 @@
 from contextlib import contextmanager
 import unittest
 
+from traitsui.tests._tools import skip_if_not_qt4
+
 from enable.container import Container
 from enable.qt4.scrollbar import NativeScrollBar
 from enable.window import Window
 
 
+@skip_if_not_qt4
 class ScrollBarTest(unittest.TestCase):
 
     def setUp(self):
--- a/enable/tests/qt4/scrollbar_test.py
+++ b/enable/tests/qt4/scrollbar_test.py
@@ -1,0 +1,131 @@
+
+from contextlib import contextmanager
+import unittest
+
+from enable.container import Container
+from enable.qt4.scrollbar import NativeScrollBar
+from enable.window import Window
+
+
+class ScrollBarTest(unittest.TestCase):
+
+    def setUp(self):
+        self.container = Container(position=[0, 0], bounds=[600, 600])
+        self.window = Window(None, size=(600, 600), component=self.container)
+
+    @contextmanager
+    def setup_window(self, window):
+        window.control.show()
+        try:
+            yield
+        finally:
+            window.control.destroy()
+
+    @contextmanager
+    def setup_scrollbar(self, scrollbar, window, enable_range, value):
+        scrollbar._draw_mainlayer(self, window._gc)
+        try:
+            yield
+        finally:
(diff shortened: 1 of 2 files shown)


jwiggins:
So the `value` variables here are coming from a `GraphicsContext` coordinate system? If so, utility routines are a very good idea. This conversion comes up all the time in enable.

itziakos:
I think that using the  https://github.com/enthought/pyface/blob/master/pyface/ui/qt4/util/modal_dialog_tester.py on should be able to verify that the scrobars have the right value.

corranwebster:
@jwiggins In the case I'm looking at, the value is coming from a `GraphicsContext`, but it doesn't have to - it can be any arbitrary quantity being controlled by the scrollbar.  Enable expects the position of a vertical scrollbar to go bottom to top, but Qt expects top down, but the coordinates are arbitrary.

Nevertheless, I've pulled the computation out into something re-usable and hopefully clearer.

corranwebster:
Hrm... this probably isn't right if `low` isn't 0.

corranwebster:
Added a test and corrected for non-zero min values.

I think that this is ready to merge.

jwiggins:
Thanks Corran! I'll take a look later during the inevitable Travis pauses.

corranwebster:
Hrmmm... need to skip the new test when running under wx.

enable/tests/qt4/scrollbar_test.py
```diff
@@ -0,0 +1,134 @@
...
+    @contextmanager
+    def setup_window(self, window):
+        window.control.show()
+        try:
+            yield
+        finally:
+            window.control.destroy()
+
+    @contextmanager
+    def setup_scrollbar(self, scrollbar, window, enable_range, value):
```
jwiggins: `enable_range` and `value` are being ignored here.

enable/tests/qt4/scrollbar_test.py
```diff
@@ -0,0 +1,134 @@
+
+from contextlib import contextmanager
+import unittest
+
+from traitsui.tests._tools import skip_if_not_qt4
+
+from enable.container import Container
+from enable.qt4.scrollbar import NativeScrollBar
```
jwiggins: This import is failing on the null toolkit test

jwiggins: WX tests too

jwiggins:
I think it might still fail when it actually runs the tests. You should look at the previous run with Qt.

corranwebster:
Hmmm... yes, looks like the window isn't being created, but the tests are running fine on my machine.  Not sure what's going on.

corranwebster:
A bit more digging, it looks like `_size` isn't being set because the `_paint` method doesn't get called in Travis.  Trying to use the event loop to get the window set up properly.

jwiggins:
Sounds good. Once Travis is satisfied, I'm ready to merge this.

corranwebster:
With fixes for unexpected success from master, the tests are now passing, so I think this is good.

jwiggins:
LGTM!

PR #206, opened 2015-08-21
Containerize Travis CI Builds

corranwebster:
In preparation for Python 3 stuff.


corranwebster:
Well, that's not going to work since it calls out to `travis_setup.sh` which `sudo`'s a bunch of stuff.

jwiggins:
The `sudo` commands are all for installing dependencies. If there was a clean way of installing everything in a venv, you could sidestep the need for root privs :wink:

jwiggins:
Or... that. Fingers crossed!

corranwebster:
As expected, PIL seems unhappy. :(

corranwebster:
Why do we need to continue to support PIL again?  Would it be OK to just to CI with Pillow going forward?

corranwebster:
Current state - pycairo is being painful, not sure about PIL.

.travis.yml
```diff
@@ -1,7 +1,26 @@
...
+    - python-wxtools
+    - ccache
+    - cmake
+    - swig
+    - zlib1g-dev
+    - libpng-dev
+    - libfreetype6-dev
+    - libcairo2-dev
+    - python-cairo
+    - python-cairo-dev
```
itziakos: do we need the cairo dev?

dev_requirements_pil.txt
```diff
@@ -1,8 +1,11 @@
 --allow-external PIL
 --allow-unverified PIL
+--allow-external pycairo
```
itziakos: if we have pycairo then we do not need to install it here, unless we install a newer version correct?

itziakos:
I would suggest that this PR waits for #208 before merging. I think it would be easier that way

corranwebster:
I agree on #208, but would like to prove out the containerized building in the meantime.

itziakos:
May I also suggest to start using codecov instead of coveralls? 
```
after_success:
  - pip install codecov
  - codecov
```

.travis.yml
```diff
@@ -1,7 +1,26 @@
...
+    - python-pip
+    - python-numpy
+    - python-wxtools
+    - ccache
+    - cmake
+    - swig
+    - zlib1g-dev
+    - libpng-dev
+    - libfreetype6-dev
+    - libcairo2-dev
```
itziakos: We should also not need to install this, we should let `python-cairo` bring any dependencies it needs.

jwiggins:
#208 is merged

corranwebster:
OK, I think that this is done.

itziakos:
:+1:

jwiggins:
:+1: Looks like a nice performance boost

PR #208, opened 2015-08-28
Better support for Pillow

itziakos:
This PR replaces the instances of `import Image` to `from PIL import Image`. This will allow to use Pillow and stiil be backward compatible with using PIL. The PR also resorects some tests from integrationtests to test the GraphicContext array and adds some basic tests for the `kiva.image.Image` class


jwiggins:
Are there any concerns about only testing against Pillow?

Other than that, LGTM.

corranwebster:
As long as Pillow supports the same versions of Python that we want to support, I'm fine with Pillow-only testing (and with that an implicit deprecation of PIL for the longer-term).

itziakos:
is this ok to merge?

jwiggins:
Yes

PR #137, opened 2014-06-05
Python 3 compatibility and Travis integration for it.

pradyunsg:
This PR adds Python 3 compatibility and Travis integration of the same to Enable. Passes all tests.

PR #209, opened 2015-09-11
Python 3 Compatibility

corranwebster:
Updated version of #137


corranwebster:
Looks like kiwisolver isn't Python 3 compatible?

corranwebster:
Yep, Kiwi solver doesn't support Python 3 yet, and probably won't soon.  I'm guessing we can skip the constraints-based layout tests in the short-term, but the lack of the feature for Python 3 is a bit of a disappointment.

@jwiggins Thoughts on ways forward?

.travis.yml
```diff
@@ -43,8 +42,10 @@ install:
   # nose is already installed
   # numpy is already installed
   - pip install -r travis-ci-requirements
-  - pip install coveralls
+  - if [[ $TRAVIS_PYTHON_VERSION == "2.7_with_system_site_packages" ]]; then pip install kiwisolver; fi
   - python setup.py develop
+before_script:
+  - export NOSE_WITH_COVERAGE=True
```
itziakos: This should not be necessary sinc ewe are not using the nose coverage plugin.

.travis.yml
```diff
@@ -43,8 +42,10 @@ install:
   # nose is already installed
   # numpy is already installed
   - pip install -r travis-ci-requirements
-  - pip install coveralls
+  - if [[ $TRAVIS_PYTHON_VERSION == "2.7_with_system_site_packages" ]]; then pip install kiwisolver; fi
```
itziakos: you can use the pip enviroment markers (https://pip.pypa.io/en/latest/reference/pip_install.html#requirement-specifiers)

itziakos:
> Yep, Kiwi solver doesn't support Python 3 yet, and probably won't soon. I'm guessing we can skip the constraints-based layout tests in the short-term, but the lack of the feature for Python 3 is a bit of a disappointment.

I thought that the constrained layout was optional but it looks that it is not really (see coordinate_box.py)

jwiggins:
@corranwebster: Uh.. go back to casuarius? Or add Python 3 support to kiwisolver?

Failing that, it might be possible to refactor things in such a way that constraints-based layout is actually optional.

itziakos:
> Failing that, it might be possible to refactor things in such a way that constraints-based layout is actually optional.

I hacked something up to make it optional but you guys might want to take a look (it looks a little ugly).

jwiggins:
Yeah, that's probably a bit too hacky.

itziakos:
Tests now pass, of somebody can help with a cleaner way to get kiwisolver optional the PR is good to go.

jwiggins:
https://github.com/jwiggins/kiwi/tree/py3k

corranwebster:
@jwiggins :clap:

jwiggins:
I can't take too much credit there. It's based on an earlier port by @dwillmer.

jwiggins:
PR opened for kiwi (note that I renamed my branch): nucleic/kiwi#13

itziakos:
@jwiggins even with python3 support for `kiwisolver` we still have the problem that the `kiwisolver` depedency is not optional.

corranwebster:
@jwiggins and @itziakos Would it make sense to vendorize Kiwisolver inside enable.layout.*

itziakos:
> @jwiggins and @itziakos Would it make sense to vendorize Kiwisolver inside enable.layout.*

I think that vendorizing kiwisolver to solve the non-optional relationship with enable is wrong.

jwiggins:
@itziakos: I agree that constraint-based layout should be optional.

*I disagree that such an arrangement should be implemented here.* This PR is for Python 3.

itziakos:
> I disagree that such an arrangement should be implemented here. This PR is for Python 3.

That is fine with me. I have created a new issue #212 for the task of making kiwisolver truly optional. However, please note that #212 at the moment due to the fact that kiwisolver is not python 3 compatible #212 or nucleic/kiwi#13 are blocking this PR.

jwiggins:
nucleic/kiwi#13 has now been merged! This PR can move forward.

corranwebster:
Everything is passing.  The native scrollbar tests are passing, but a bit dodgy---but then the native scrollbar is a bit dodgy.  I'll open a ticket for that, I don't think it's because this PR is bad.

@jwiggins @itziakos Any show-stoppers that would prevent merging this?

itziakos:
@corranwebster in Qt you need to run the event loop in order to destroy widgets using process events will not work as expected (something like https://github.com/enthought/pyface/blob/master/pyface/ui/qt4/util/gui_test_assistant.py#L79)

corranwebster:
There appears to be a Heisenbug in the `viewport_test_case.py` test cases under Python 3.  Simply re-running can get the tests to pass.  I can occasionally get a replication when running in Python 3.

itziakos:
> There appears to be a Heisenbug in the viewport_test_case.py test cases under Python 3. Simply re-running can get the tests to pass. I can occasionally get a replication locally when running in Python 3.

It is possible that a number of object deletions and other events are leaking between tests and that can result in unsuspected behavior and timeouts down the line. I would open an issue about the random failure. We need to cleanup the testing infra and make sure that tests do not leak widgets in the Qt ar Wx application singletons

itziakos:
@corranwebster do you still want to keep the workaround on making kiwisolver optional (see coordinate_box)? As far as I understand now that kiwisolver supports python3 we can delegate making kiwisolver optional at a later time.

corranwebster:
It's replicable just running that test case alone; so its not leakage from other tests; although it could be leakage from within the one test.

corranwebster:
I'm OK with making it optional being in here even if kiwisolver now supports Python 3.  We can revisit later if needed.

jwiggins:
@corranwebster: :+1: from my end

corranwebster:
Looks like it is an initialisation order issue for the Heisenbug: the order in which traits get set at start-up time seems variable under Python 3, and if you get the wrong order then you can end up with bad results.

I will merge this, and open an issue for the bug, as it's not a 5 minute fix, by the look of it.

corranwebster:
See #228 for issue.

PR #214, opened 2015-09-29
Add docs and examples to the source distribution

### LINKED ISSUE #213: pypi packages is missing docs and examples
Could you please add back the examples folder and the docs to the pypi package.

jwiggins:
Fixes #213

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,5 @@
 include kiva/agg/agg.i
-include chaco/tests/data/PngSuite/*.png
-include chaco/tests/data/PngSuite/LICENSE.txt
+include docs/Makefile
+include docs/kiva/agg/notes
+recursive-include docs *.py *.rst *.txt *.css *.png *.ico *.doc
+recursive-include examples *.py *.txt *.gif *.jpg *.enaml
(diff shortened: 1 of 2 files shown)


jonathanrocher:
I tried to build the html docs yesterday from the sdist I generated, and failed with the following error because of the way `conf.py` looks for `enable`'s `__init__`:
```
$ make html
sphinx-build -b html -d build/doctrees   source build/html
Running Sphinx v1.2.3

Exception occurred:
  File "../../enable/__init__.py", line 6, in <module>
ImportError: cannot import name full_version
The full traceback has been saved in /var/folders/ff/z56qhb_936g8xx3rzm1_1r800000gp/T/sphinx-err-gn5R7X.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://bitbucket.org/birkenfeld/sphinx/issues/>. Thanks!
make: *** [html] Error 1
```
We may want to fix that in `conf.py`...

jwiggins:
Nice catch @jonathanrocher. I've changed `conf.py` over to `enable/_version.py` and it appears to work.

Issue #204, opened 2015-06-11
Intermittent failure seen in Kiva GL drawing test


jwiggins:
#201 initially failed its CI check when the `TestGLDrawing.test_text_clip` test fail (confusingly, by passing when failure is the expectation). The test failure is intermittent, so tweaks to either the test, or the GL backend might be needed to address the issue.

A fix was attempted in #202, but abandoned.

jwiggins:
GL backend has been removed.

PR #215, opened 2015-10-17
fix support for Pillow 3.0.x

itziakos:
The new Pillow 3.0.x has the `tostring` and `fromstring` methods from `Image` raise an exception. This PR fixed enable the support for Pillow 3.0.x

in more details:
- a compatibility module is added to allow using new and old versions of Pillow
- Wx and Qt backends are checked against Pillow 3.0.0
-  The null.image backend is check against Pillow<3.0.0 and Pillow 3.0.0 on python 2.7 and only on Pillow<3.0.0 on python 2.6

### DIFF
--- a/kiva/compat.py
+++ b/kiva/compat.py
@@ -10,14 +10,14 @@ PILLOW_VERSION = getattr(PIL, 'PILLOW_VERSION', PIL.VERSION)
 
 
 def piltostring(image):
-    if PILLOW_VERSION.starts_with('3'):
+    if PILLOW_VERSION.startswith('3'):
         return image.tobytes()
     else:
         return image.tostring()
 
 
 def pilfromstring(*args, **kwargs):
-    if PILLOW_VERSION.starts_with('3'):
+    if PILLOW_VERSION.startswith('3'):
         return Image.frombytes(*args, **kwargs)
     else:
         return Image.tostring(*args, **kwargs)
--- a/kiva/compat.py
+++ b/kiva/compat.py
@@ -3,21 +3,20 @@ Compatibility module to help support various versions of PIL and
 Pillow.
 
 """
-import PIL
 from PIL import Image
 
-PILLOW_VERSION = getattr(PIL, 'PILLOW_VERSION', PIL.VERSION)
+HAS_FROM_BYTES = hasattr(Image, 'frombytes')
 
 
 def piltostring(image):
-    if PILLOW_VERSION.startswith('3'):
+    if hasattr(image, 'tobytes'):
         return image.tobytes()
     else:
         return image.tostring()
 
 
 def pilfromstring(*args, **kwargs):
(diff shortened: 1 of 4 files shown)


itziakos:
This PR is ready for review

kiva/compat.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+import PIL
+from PIL import Image
+
+PILLOW_VERSION = getattr(PIL, 'PILLOW_VERSION', PIL.VERSION)
+
+
+def piltostring(image):
+    if PILLOW_VERSION.startswith('3'):
+        return image.tobytes()
```
rkern: Can we just check for the existence of `tobytes()` instead?

rkern: I am going to forget to set the environment variable, like, 100% of the time.

itziakos: Done, code does not depend on the PIL/Pillow version

itziakos: oops, that was wrong (sorry no coffee yet). So the issue is that the `tostring` is there but now raises
as a result we cannot just check for the attribute. Even better, the exception is `Exception` so a try except is going to be useless. So we have to check the Pillow version

*Note*: There is no enviroment variable

rkern: Sorry, no coffee on this side either, I saw the `env` in the `.travis.yml` and thought that was being read here. Now I see that it's just controlling the dependencies that get installed for the matrix of tests. Yes, this is fine. What I was suggesting probably would have also worked, though: test for the presence of `tobytes()`, not the absence of `tostring()`.

rkern: Actually, I think I still have a preference for checking for `tobytes()` instead of the version string. When Pillow goes to `4.0`, we still want this to work.

itziakos: good point

kiva/compat.py
```diff
@@ -0,0 +1,22 @@
+"""
+Compatibility module to help support various versions of PIL and
+Pillow.
+
+"""
+from PIL import Image
+
+HAS_FROM_BYTES = hasattr(Image, 'frombytes')
```
jwiggins: Any particular reason why you're defining this constant here?

jwiggins:
Minor code quibble. Other than that, LGTM. Merge once Travis is happy. ("unexpected success".. Kinda silly when you think about it)

kiva/compat.py
itziakos: Since Image is a class from PIL/Pillow I thought that there is no reason to check every time if the frombytes exists, so I thought having a constant with the result was a good idea.

jwiggins: Ahh, I see now. That's fine.

jwiggins:
The earlier test failure is already covered by issue #204. Any objections to merging?

rkern:
None from me.

PR #216, opened 2015-12-17
Refactoring conditional directives for alternative if statements.

wmarquesr:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.
- [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
- [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/kiva/agg/freetype2/src/psaux/t1decode.c
+++ b/kiva/agg/freetype2/src/psaux/t1decode.c
@@ -199,6 +199,7 @@
 #ifdef FT_CONFIG_OPTION_INCREMENTAL
     T1_Face      face  = (T1_Face)decoder->builder.face;
 #endif
+    int decoder_glyph_names;
 
 
     if ( decoder->seac )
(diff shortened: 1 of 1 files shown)


corranwebster:
Thanks for the contribution!

The code here is from an old, vendorized copy of freetype that is used for font handling, so it's probably more appropriate to contribute fixed to the freetype project, unless this is fixing a particular issue that you are running into with Kiva, Enable or Chaco, in which case we're more than willing to consider it.

wmarquesr:
Ok, but you can say what you think about this kind of refactoring?

corranwebster:
I'd be happy to approve PRs with this sort of clean-up for any code in Enable, as long as the code is human-written (ie. not generated by SWIG) and not vendorised (ie. not in the Fonttools or Agg subdirectories).

jwiggins:
I would add one caveat concerning Agg. As far as I know that code is not being actively maintained by an outside project. Enable basically owns its version (for better or worse).

wmarquesr:
So why the PR do not acepted yet ?

PR #223, opened 2016-04-28
Fix Tests

corranwebster:
Looks like PyParsing ~~2.0.4~~ 2.1.0 broke Savage.  Also latest version of Hypothesis doesn't work on Python 2.6, so I've downgraded both.

### DIFF
--- a/travis-ci-requirements
+++ b/travis-ci-requirements
@@ -1,7 +1,7 @@
 numpy
 coverage
 pillow
-pyparsing
+pyparsing==1.5.6
 PyPDF2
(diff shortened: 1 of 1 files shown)


corranwebster:
Not sure why coverage is complaining - seems to be a new test they've added.

@jwiggins @jonathanrocher @itziakos or @cfarrow you have a moment to check the PR.

itziakos:
@corranwebster codecov now supports configuration using the yml file and since we do not have one default configuration will be used. I will spend some time tonight and add a simple configuration

jwiggins:
Changes look fine. Codecov is another story, it appears...

PR #230, opened 2016-05-19
Fix for build issues

### LINKED ISSUE #229: python setup.py install fails to install kiva.agg
And possibly other things as well.  May be related to #98.  As a consequence, `pip install git+http://github.com/enthought/enable.git#egg=enable` also fails.  Workaround is to use develop instead of install (as we have to do in Chaco build setup right now), but that's not a real solution.

corranwebster:
Should fix #229


jwiggins:
Uhh... you're reverting a change that you made: https://github.com/enthought/enable/commit/8f3ff1bf839c89468c7a9efd061bce62d54f6fdb

Do you remember why you made that change to begin with?

jwiggins:
Well... I think the Travis failure answered the "why".

Odd. :unamused:

corranwebster:
Very odd - it looks like `pip install` wants one thing and `setup.py develop` wants the other.

Mind you, `setup.py develop` is dodgy when combined with `2to3`.

corranwebster:
OK, so problem is that if you don't do the builder swap then the integration tests are not updated to Python 3 - but doing the builder swap is not the right thing to do, I think.  I'd say that probably the best option is to make the integration tests work on both 2 and 3 with the same codebase.

corranwebster:
And the code in the integration tests is horrible, so I've folded them into the Kiva test suites instead.  They may need some tweaking to run under Python 3.

corranwebster:
OK - back to builds working.  @jwiggins or @itziakos if you have a chance to double-check this it would be good.  I haven't checked that `pip install .` works properly yet, and won't have a chance until this evening.

jwiggins:
I just tested `pip install .` on a fresh Python 3.4 virtualenv. It worked. I'm :+1: on this PR

corranwebster:
Thank you.

PR #232, opened 2016-05-26
Don't assume window is MockWindow when dealing with pointer position

### LINKED ISSUE #211: `EnableTestAssistant` only works with mock window.
The test assistant is supposed to work with both real and mock windows:
https://github.com/enthought/enable/blob/master/enable/testing.py#L253

However, a recent change assumes that the window is a mock one:
https://github.com/enthought/enable/blob/master/enable/testing.py#L259
(introduced in enthought/enable@bd86c7c).

This breaks tests that passed a real window explicitly.

corranwebster:
Fixes #211.

### DIFF
--- a/enable/testing.py
+++ b/enable/testing.py
@@ -47,7 +47,9 @@ class EnableTestAssistant(KivaTestAssistant):
         window : AbstractWindow, optional
             The enable AbstractWindow to associate with the event. Default
             is to create a mock class instance. If the window has a mouse
-            owner then that interactor is used.
+            owner then that interactor is used.  If a non-Mocked window is
+            used, then either the test should not require the use of
+            get_pointer_position() or should use a mock for the method.
 
         alt_down : boolean, optional
             The button is pressed while `alt` is down. Default value is False.
@@ -231,7 +233,9 @@ class EnableTestAssistant(KivaTestAssistant):
         window : AbstractWindow, optional
             The enable AbstractWindow to associate with the event. Default
             is to create a mock class instance. If the window has a mouse
-            owner then that interactor is used.
+            owner then that interactor is used.  If a non-Mocked window is
+            used, then either the test should not require the use of
+            get_pointer_position() or should use a mock for the method.
 
         alt_down : boolean, optional
             The button is pressed while `alt` is down. Default value is False.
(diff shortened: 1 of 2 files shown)


enable/testing.py
```diff
@@ -256,7 +256,9 @@ def mouse_move(self, interactor, x, y, window=None,
                                         alt_down=alt_down,
                                         control_down=control_down,
                                         shift_down=shift_down)
-        window.get_pointer_position.return_value = (x, y)
+        if isinstance(window, _MockWindow):
```
itziakos: If one passes a real window how does the `mouse_move` actually happens? When a real window get_pointer_position is called it will not have the right (x, y) value. In sort the call will not fail but the desired affect will probably not happen either.

itziakos: Please also note that it is `get_pointer_position` that should be a mock not the `window`.

corranwebster: The `get_pointer_position` function isn't used by most consumers of the `mouse_move` event - in most cases the event position is sufficient.  But you are right, we should check to see if the method is mocked, not the window.

corranwebster: And needless to say, if you pass a real window in a test, you'll need to mock the `get_pointer_position` yourself if you need that handled.

itziakos: > The get_pointer_position function isn't used by most consumers of the mouse_move

This is True but it is not always clear when this is the case (e.g. using a subclass of a Tool). So the user will end up with a very weird error down the line.

Ideally we should be able to fake the mouse move in any Window (e.g maybe sent the mouse move event to the Qt widget) . However, for the moment, I would suggest to raise a warning if the get_pointer_position is not mocked and update the docstring to mention that for the mouse_move to be properly fakes the operation `get_pointer_position` needs to be a mock instance

itziakos: It would also be nice to create a test using a real Window so that users can see how it is done.

itziakos: opened feature request #233 for supporting a more native way for faking mouse moves in real windows

itziakos:
:+1:

corranwebster:
@itziakos We have a test :)

itziakos:
:+1:

PR #238, opened 2016-08-17
Change points_in_polygon to return boolean array

### LINKED ISSUE #237: Agg points_in_polygon should return bool array
Currently it returns an array of int.  This causes problems with newer numpy versions when using bitwise operations.

See https://github.com/enthought/chaco/issues/296 for example.

corranwebster:
Fixes #237.

Using `unsigned char` for boolean array rather than `int`.

### DIFF
--- a/kiva/agg/src/agg_typemaps.i
+++ b/kiva/agg/src/agg_typemaps.i
@@ -83,7 +83,7 @@
 }
 
 // --------------------------------------------------------------------------
-// Typemaps for (int* results, int Nresults)
+// Typemaps for (unsigned char* results, int Nresults)
 //
 //    For: points_in_polygon
 //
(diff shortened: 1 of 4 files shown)


jwiggins:
Why use `unsigned char` and not `bool`?

corranwebster:
Are we guaranteed `bool*` is an array of bytes?  My knowledge of C++ isn't deep enough to know if that's certain.

jvkersch:
> Are we guaranteed bool* is an array of bytes?

I heard you guys talking about this at lunch and I was curious myself, so I looked it up in the C++ standard, and it's not guaranteed.

jwiggins:
Right... I can't find a link to the relevant part of the C++ standard, but I did find this: http://en.cppreference.com/w/cpp/language/types

So no, it's not necessarily an array of bytes. Is this then dictated by what the Numpy C-API expects (even if obscured by SWIG)?

jwiggins:
I say use whatever Numpy expects. That's probably `unsigned char`. If so, LGTM.

corranwebster:
Yep - it's because numpy expects a bool array to be an array of bytes in memory with 0 and 1 values. So signed vs. unsigned shouldn't matter, I guess, but unsigned makes it clearer that we're just getting a chunk of memory.

PR #241, opened 2016-10-10
FIX: affine transformation trs_factor

pberkes:
This is #61 , resubmitted from my personal fork after solving merge conflicts
and flake8 problems.

The calculation to find the transformation parameters
for a translation/rotation/scale transformation were
in fact for a translation/scale/rotation one.

The test only passed because of the particular combination of parameters used.


jwiggins:
Awesome! Thanks for getting this cleaned up!

PR #246, opened 2016-10-26
Kiva explorer demo

corranwebster:
Add a really simple traitsui view for playing with Kiva.

screen shot 2016-10-26 at 9 30 24 am

### DIFF
--- a/examples/kiva_explorer.py
+++ b/examples/kiva_explorer.py
@@ -42,6 +42,7 @@ with gc:
 
 
 class ScriptedComponent(Component):
+    """ An Enable component that draws its mainlayer from a script """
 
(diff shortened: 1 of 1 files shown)


jwiggins:
:heart_eyes:

Bless you, sir. This is exactly what I need right now.

I should probably push my pyagg backend branch, huh?

jwiggins:
Just a quick read from my end, but LGTM

corranwebster:
You're welcome!  And since this is non-critical code, I shall merge it.

eric-jones:
It would be useful to have something that printed out the rendering time on this... (per render and frames per second)

PR #247, opened 2016-10-26
celiagg Kiva backend

jwiggins:
This requires https://github.com/celiagg/celiagg (or `pip install celiagg`!) to be built and installed.

This works fairly well now. We should discuss how to proceed.

### DIFF
--- a/kiva/pyagg.py
+++ b/kiva/pyagg.py
@@ -409,9 +409,8 @@ class GraphicsContext(object):
 
             Region should be a 4-tuple or a sequence.
         """
-        x1, y1, x2, y2 = x, y, x + w, y + h
         tx, ty = self.transform.tx, self.transform.ty
-        self.canvas_state.clip_box = agg.Rect(tx+x1, ty+y1, tx+x2, ty+y2)
+        self.canvas_state.clip_box = agg.Rect(tx+x, ty+y, w, h)
 
     def clip_to_rects(self, rects):
         """ Clip context to a collection of rectangles
@@ -715,9 +714,7 @@ class GraphicsContext(object):
 
         XXX: This is currently broken for some reason
         """
-        # set up drawing state and function
-        shape = agg.ShapeAtPoints(self.path.path, points)
-
+        shape = agg.ShapeAtPoints(path.path, points)
         self.canvas_state.drawing_mode = draw_modes[mode]
         self.gc.draw_shape(shape, self.transform, self.stroke_paint,
                            self.fill_paint, self.canvas_state)
@@ -797,7 +794,7 @@ class CompiledPath(object):
 
     def add_path(self, other_path):
         if isinstance(other_path, CompiledPath):
-            self.path.addPath(other_path.path)
+            self.path.add_path(other_path.path)
 
     def close_path(self):
         self.path.close()
--- a/kiva/pyagg.py
+++ b/kiva/pyagg.py
@@ -622,10 +622,9 @@ class GraphicsContext(object):
         if self.font is None:
             raise RuntimeError("get_text_extent called before setting a font!")
 
-        x, y = self.text_pos
-        width = self.font.width(text)
-        height = self.font.height
-        return x, y, x + width, y + height
+        x1, x2 = 0.0, self.font.width(text)
+        y1, y2 = 0.0, self.font.height
+        return x1, y1, x2, y2
 
     def get_full_text_extent(self, text):
         """ Backwards compatibility API over .get_text_extent() for Enable
--- a/kiva/pyagg.py
+++ b/kiva/pyagg.py
@@ -1,0 +1,776 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2016, Enthought, Inc
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in enthought/LICENSE.txt and may be redistributed only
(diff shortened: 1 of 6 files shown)


kiva/pyagg.py
```diff
@@ -0,0 +1,725 @@
...
+    'pad': agg.GradientSpread.SpreadPad,
+    'repeat': agg.GradientSpread.SpreadRepeat,
+    'reflect': agg.GradientSpread.SpreadReflect,
+}
+StateBundle = namedtuple('StateBundle',
+                         ['state', 'path', 'stroke', 'fill', 'transform',
+                          'text_transform', 'font'])
+
+
+class GraphicsContext(object):
```
corranwebster: It might make sense to inherit from `AbstractGraphicsContext` or `basecore2d.GraphicsContextBase`.  Although at this point it may be more hassle than its worth.

corranwebster:
Yay!  Will try this out at some point soon.

kiva/pyagg.py
```diff
@@ -0,0 +1,725 @@
...
+    def set_font_size(self, size):
+        msg = "set_font_size not implemented on pyagg yet."
+        raise NotImplementedError(msg)
+
+    def set_character_spacing(self, spacing):
+        msg = "set_character_spacing not implemented on pyagg yet."
+        raise NotImplementedError(msg)
+
+    def set_text_drawing_mode(self):
+        """ XXX: What is this for??
```
corranwebster: This is how to draw text: filled, outline, etc.  See `constants.py` for the options.

kiva/pyagg.py
corranwebster: Ah, I see that you've registered it with the ABC below.  Ignore the above, then.

kiva/pyagg.py
jwiggins: Ahhh... You can probably see that I based this on the QPainter backend. Also, based on what I've read in the AGG backend, this isn't possible to implement there (fonts are raster only).

Do you know of a backend which implements this correctly? Quartz?

corranwebster: Yes, if pyagg has vector font support that should cover it if you can specify the render method (fill, stroke, etc.) for the vector path of the text.

jwiggins: Implemented!

jwiggins:
I removed some previously committed changes to the examples... Force pushed. Mind the gap.

This is almost usable now. 😄

jwiggins:
This is now functionally complete. Bugs included at _no extra cost to you_!

jwiggins:
I've fixed all the bugs I could find. This actually works pretty well with all of the Chaco demos now!

corranwebster:
Playing around with this right now.  I noticed that there are some artifacts when using `gc.draw_image` and not aligned to exact pixel edges. Eg.
```
gc.draw_image(a, (200, 400, 100, 100))
```
looks like  this:
screen shot 2016-12-22 at 2 29 13 pm
(notice line at right border) where if I force image to draw precisely on pixel boundaries, eg.
```
gc.draw_image(a, (200.5, 400.5, 100, 100))
```
there is no such problem:
screen shot 2016-12-22 at 2 32 15 pm

corranwebster:
Not sure if I should be putting these into the celiagg repo or here?

There is an issue that coloring isn't working for fonts.  In the default `kiva_explorer.py` script, the fonts should be coloured like this:
screen shot 2016-12-22 at 2 42 49 pm
but instead are just rendererd in black
screen shot 2016-12-22 at 2 42 18 pm
(although FWIW the text layout is better)

corranwebster:
To answer the question about how to proceed.  There are two ways we could do this:
- the way that this is structured now, where we have the current Agg backend and the new Celiagg backend and you access them with `ETS_TOOLKIT=....celiagg` or whatever.  We would need to be careful about use of the current `Image` graphics context and code shouldn't assume the existence of Agg or Celiagg for rendering images (maybe lean more heavily on Pillow for that stuff).
- we replace the Agg backend with the Celiagg backend everywhere.  We'd need Celiagg to be reasonably feature-comparable to the current backend before doing this.
I'm not sure which is best.  I'd be happy to see the back of the current Agg wrappers, but they work for the most part, so it may make sense to have a more gradual transition

jwiggins:
The font color issue is an interesting one. If you try kiva_explorer.py with the Quartz backend, you will see that it matches celiagg. Basically, some backends use stroke color for text, others use fill color. I kinda punted and followed Quartz, but honestly it's not hard to change.

corranwebster:
Hmmm... The `qpainter` backend also uses stroke color for text.  Given that changing those is likely to be difficult (QPainter is just reflecting the underlying behaviour of Qt, I imagine Quartz is similar) I'm willing to live with using `set_stroke_color` instead of `set_fill_color` for celiagg.

Making sure we are consistent in the change throughout the dependent codebases will be a little finicky.

corranwebster:
Thought a bit more about the way forward.  I'm happy for this to be merged as-is so that we can work on further improvements based on this.  We can revisit what to do with the current agg backend as things go forward.

jwiggins:
I'd like to explore the image drawing issue a little bit first, but I suspect that it lies completely within celiagg. We shall see.

After that, yes, I think this should just be merged since it's not disruptive on its own. Then we can begin to discuss the more disruptive changes.

jwiggins:
Okie dokie. I'm pulling the trigger. Thanks for the feedback Corran!

PR #250, opened 2016-11-05
Move points_in_polygon to Cython

jwiggins:
Since I'm maneuvering to eliminate the agg extension from kiva, I need to do something with `points_in_polygon`. I believe the only reason it was part of that extension was to cut down on the number of extension modules being built. Now it lives in its own extension.

**NOTE**: I don't want to merge this until after 4.6.0 is released.


corranwebster:
Generally a good idea, although I haven't tried this out.

corranwebster:
This is a backwards-incompatible change which will require updating some stuff in Chaco (and at least one other internal project I know of).

Possibly this should be kept in kiva (although not in kiva.agg), rather than moved to enable?  I'm in two minds.

jwiggins:
I'm planning for `kiva.agg` to go away. If the import has to change, why not move it into enable at the same time? It doesn't seem drawing related, so that's my argument for not leaving it in kiva.

corranwebster:
While this is currently used for hit detection and point selection by Chaco, this is the sort of routine that can be useful for higher-level drawing routine optimization (eg. avoiding drawing points which are obscured).

In other words, although it's not about drawing, it is about pure geometry, which seems a more Kiva-y thing than an Enable-y thing.

jwiggins:
Agreed. I've moved the code back to kiva, but still outside of agg.

jwiggins:
Thanks for reviewing!

rkern:
@jwiggins Can you please expose it back in `kiva.agg` with a deprecation warning? This is part of the public API and used in Chaco, so now all of the Chaco PR Travis runs are failing. Migrating Chaco isn't hard, but then Chaco would be relying on an unreleased version of Enable. The deprecation period makes the coordination easier.

jwiggins:
Sure thing

PR #251, opened 2016-11-05
Don't use a GraphicsContext for storage in the Image component

jwiggins:
@corranwebster: You are the author of the `Image` component. Do you recall what your motivation was for putting the image data into a `GraphicsContext` and passing that when drawing? Why not just draw directly from an array?

I'd like to phase out certain usage of AGG so that it can be removed. This is one of the spots that needs attention.

### DIFF
--- a/enable/primitives/image.py
+++ b/enable/primitives/image.py
@@ -11,11 +11,10 @@
 """
 
 # Enthought library imports
-from traits.api import Array, Enum, Instance, Property, cached_property
+from traits.api import Array, Enum, Property, cached_property
 
 # Local imports
 from enable.component import Component
-from kiva.image import GraphicsContext
 
 
 class Image(Component):
@@ -37,8 +36,8 @@ class Image(Component):
     #: the size-hint for constraints-based layout
     layout_size_hint = Property(data, depends_on="data")
 
-    #: the image as an Image GC
-    _image = Property(Instance(GraphicsContext), depends_on="data")
+    #: the image as a C-contiguous ndarray
+    _image = Property(Array(shape=(None, None, None)), depends_on="data")
 
     @classmethod
     def from_file(cls, filename, **traits):
@@ -56,9 +55,8 @@ class Image(Component):
     def _draw_mainlayer(self, gc, view_bounds=None, mode="normal"):
         """ Draws the image. """
         with gc:
-            gc.draw_image(
-                self._image, (self.x, self.y, self.width, self.height)
-            )
+            rect = (self.x, self.y, self.width, self.height)
+            gc.draw_image(self._image, rect)
 
     @cached_property
     def _get_format(self):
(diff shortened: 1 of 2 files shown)


corranwebster:
No particular reason that I can think of - it was implemented in haste in reaction to a Chaco image plot being used for the purposes of rendering a logo.

Only reason I can see from looking at the code is perhaps because of graphics context bitmap format issues.  It might be worth a check to see if it works correctly with something like the QPainter backend which I think is `argb32` format by default.  And I guess that images with alpha and with no alpha visually look to be correct in more general settings.

In any case if this appears to work correctly, the `format` property is no longer needed, I think - it doesn't seem to be used.

jwiggins:
Ah, OK. Thanks for the background. I'll test this a bit more extensively to see if anything falls out.

Thanks for the catch concerning `format`. It actually does seem like important information to keep around. Perhaps that's why you implemented it with `GraphicsContext` before. Maybe we need to think about modifying the kiva API? (:fearful:)

corranwebster:
If we're thinking of modifying the Kiva API, now would be the time to do it, particularly given that the formal API definition is currently descriptive (these are the methods that the backends seemed to all have), rather than prescriptive.  I'm not sure how deeply we want to go into different pixel formats (and related things like premultiplied alphas), but there may be a case for it.

As far as image handling goes, there are a fair number of places in the current codebase (eg. a fair bit in the Chaco image plot, for example) where it is assumed that we have the Agg backend available to do some image processing (usually for re-scaling or messing with alpha channels, rarely for anything that actually uses the Kiva drawing code directly).  It may be that we are better off relying on Pillow for this sort of stuff than assuming that Agg is available.

rahulporuri (approved):
The changes look innocent/good to me and the `Image._image` property is an internal implementation detail - changing which should not break any code. If yall think this is a good idea, let's go ahead with this and see what happens when we release the rc1 into the wild.

> I'd like to phase out certain usage of AGG so that it can be removed.

I'd like to know what "certain usage" refers to here though. It'll be useful having this documented somewhere (even if its just a PR) instead of nowhere.

jwiggins:
> I'd like to know what "certain usage" refers to here though.

Basically, I thought that `from kiva.image import GraphicsContext` in the pre-change version of this PR was problematic because it was a direct import of a `GraphicsContext` implementation rather than going through the `ETS_TOOLKIT`-mediated `enable.kiva_graphics_context` module (thus creating an implicit dependency on the quirks of the `kiva.agg` implmentation).

That said, I think it's complete overkill to wrap a pixmap with a whole `GraphicsContext` just so it can be blitted into another context.

corranwebster:
I think this change is OK.  I have a vague recollection about bit orders being an issue, but any bugs with wrong colors for some Kiva backends is really an issue with that backend's draw_image method.

FWIW I am also OK with using a Pillow `Image` as the internal storage if that makes things easier.

jwiggins:
I like the idea of using pillow `Image` throughout for image storage. It's neutral and contains the all important colorspace information needed to render things correctly.

jwiggins:
Merging this now. We can move forward with makeing PIL `Image` instances a common interchange format in a future PR

Issue #253, opened 2016-12-02
Bad URL in README


corranwebster:
Sadly github.enthought.com doesn't exist any more.  Unfortunately this is front and center of the PyPI page and the github readme.

jwiggins:
So... 4.6.1 then?

corranwebster:
Not immediately, but possibly soon for 4.6.1.  There are some reports that there are issues with the native scrollbar stuff under Windows, PyQt and Python 3.5 causing massive failures in the test suite.

I'll add an issue once I get the problem analyzed, but I think there may be an issue open on the Chaco repo which is related.  It would be good to try to get it fixed for a 4.6.1.

corranwebster:
This issue is resolved.

PR #257, opened 2016-12-03
Don't import kiva/_version.py from setup.py

### LINKED ISSUE #254: Path resolution for _version.py in setup.py
~~The setup.py file assumes it is being run from the same directory as it is in.  Pip doesn't always do this.~~  Gives this error (eg. running in an otherwise clean virtual environment):
```
Processing ./enable-4.6.0.tar.gz
    Complete output from command python setup.py egg_info:
      File "/private/var/folders/qp/t7g3hmrs7tb24jhkq6b3wn_00000gq/T/pip-c_hp5a-build/setup.py", line 225, in <module>
        write_version_py(filename='enable/_version.py')
      File "/private/var/folders/qp/t7g3hmrs7tb24jhkq6b3wn_00000gq/T/pip-c_hp5a-build/setup.py", line 103, in write_version_py
        raise ImportError("Unable to import git_revision. Try removing "
    ImportError: Unable to import git_revision. Try removing kiva/_version.py and the build directory before building.
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/qp/t7g3hmrs7tb24jhkq6b3wn_00000gq/T/pip-c_hp5a-build/
```

jwiggins:
Fixes #254 and #253

six was already in the `__requires__` list in enable/__init__.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ if 'develop' in sys.argv:
     sys.argv[idx:idx] = ['build_src', '--inplace', 'build_clib'] + compiler + \
         ['build_ext', '--inplace'] + compiler
 
-from os.path import join
+from os.path import dirname, exists, join
 
 # Setuptools must be imported BEFORE numpy.distutils for things to work right!
 import setuptools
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -98,11 +107,11 @@ def write_version_py(filename):
     elif os.path.exists('kiva/_version.py'):
         # must be a source distribution, use existing version file
         try:
-            from kiva._version import git_revision, full_version
-        except ImportError:
-            raise ImportError("Unable to import git_revision. Try removing "
-                              "kiva/_version.py and the build directory "
-                              "before building.")
+            git_revision, full_version = read_version_py('kiva/_version.py')
```
corranwebster: Probably need to be careful here:

* cross-platform path construction
* need to ensure that it works if `setup.py` is not in current working dir (ie. base off of `os.dirname(__file__)`).

jwiggins: Good point. Done.

jwiggins:
I've built the sdist from this branch on both 2.7 and 3.5. It should be good to go, right?

corranwebster:
Yep - sorry should have said it LGTM earlier.

jwiggins:
Better late than never!  😄

PR #199, opened 2015-05-20
Fix issue 197

itziakos:
This PR fixes issue #197 by using enable._version.py as the file to look for version information when installing from source.

PR #260, opened 2016-12-15
Release 4.6.1

jwiggins:
Beginning the patch release...

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -14,6 +14,15 @@ Maintainence
 ------------
 
 
+Enable 4.6.1
+============
+
+Fixes
+-----
+
+* PR #257: Fix a problem encountered when installing from pip
+
+
 Enable 4.6.0
 ============
 
--- a/setup.py
+++ b/setup.py
@@ -37,9 +37,9 @@ from numpy.distutils.misc_util import is_string
 
 MAJOR = 4
 MINOR = 6
-MICRO = 1
+MICRO = 2
 
-IS_RELEASED = True
+IS_RELEASED = False
 
 VERSION = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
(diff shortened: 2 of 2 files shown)


jwiggins:
@corranwebster: I've uploaded the SDIST package to the releases page: https://github.com/enthought/enable/releases/tag/4.6.1

Tested it under OS X 2.7 & 3.5. Docs also appear to build OK.

corranwebster:
Pushed to PyPI.

Not sure that this branch should merge into master.  We probably want a `maint/4.6` branch that we use for further 4.6.x releases, and we can cherry-pick/rebase PRs from master into the maintenance branch as needed.

jwiggins:
Thanks!

I'm hesitant to have long-living branches for each version. Case in point, #199 was a fix for a broken 4.5.0 SDIST, and here we are making a 4.6.1 for the same reason, all because the problem was fixed in a release branch.

Here's what I recommend instead: **All** development and bug fixes should happen in the master branch. PRs should be squash merged for convenient cherry-picking. As an exception, release PRs should be merged normally (so the commit with `IS_RELEASED = True` can be tagged). Then, when a patch release is needed, a new branch should be created from a release tag and fixes cherry-picked from the master branch. If the changes on master since the previous release are fixes/non-breaking, then branching off the release tag should not be necessary.

I feel that this will be easier to get right in the long run, given that this package sees infrequent changes.

cfarrow:
We've used this pattern with packages that are frequently updated, and it works fine as long as everyone knows the scheme. 👍

corranwebster:
Seems fine to me, but we should make the expectations clear (particularly about squashed merges) somewhere obvious.

jwiggins:
`CONTRIBUTING.rst`... I'll type one up soon

PR #264, opened 2016-12-19
Remove vendored fontTools

### LINKED ISSUE #245: Look at fonttools...
... just look at it.

As pointed out in #244, the vendorized version of fonttools inside kiva is quite old. It should either be turned into an external dependency or upgraded to a newer version.

#151 is potentially related to this as well

jwiggins:
Fixes #245

@corranwebster: Tested lightly... I'm curious what a CI run will produce (unit tests passed locally).


corranwebster:
Hmmm... signs look good so far.  Probably worth running Chaco tests against this branch as well.

corranwebster:
Played with this a reasonable bit. Looks to be working.  No regressions in chaco. Looks reasonably good.

The `kiva/setup.py` commands to `add_packages` etc. can be simplified.

The potential for name collisions between `fonttools` and `kiva.fonttools` worries me somewhat, but I don't think it's worth the trouble of renaming to something like `kiva.fonts`.

I'd say feel free to merge once you have the `setup.py` fixed.

jwiggins:
Well... we could go to relative imports throughout if you're really worried about it

jwiggins:
Thanks for testing! I'm merging this now

PR #269, opened 2017-03-08
Ensure integer sizes

rkern:
numpy 1.12 completes the deprecation of floats as shapes. The `GraphicsContext()` constructor permits such floats, so we coerce internally at the last possible moment.

This also fixes the skip logic for an unrelated test so that I'd have a clean test suite.

This should fix the test failure in enthought/chaco#342 in the one Travis environment that is using numpy 1.12.


jwiggins:
LGTM!

PR #276, opened 2017-05-26
Use older version of fonttools

corranwebster:
It looks like fonttools is undergoing rapid change.  Fixing version form December that should be good.


jwiggins:
Nice job figuring this out. Should we pin also the version in the `__requires__` list in `__init__.py`?

PR #279, opened 2017-06-27
Add points_in_polygon to kiva.agg with DeprecationWarning

jwiggins:
@rkern: As requested.

I'm depending a bit on CI here to test this since I was having 2to3 problems locally (in other modules).

PR #280, opened 2017-06-27
Don't use import * to borrow imports from other modules

jwiggins:
Wow... so this unit test was depending on an `import nose` in a module which it was doing an `import *` from.

This fixes some test failures seen in #279


rkern:
LGTM pending relevant Travis checks.

jwiggins:
Thanks. Note that I still expect failures because enthought/traitsui#369 has not yet been merged.

jwiggins:
OK. Only one failure (which was expected) from the CI run. I'm merging. Thanks for reviewing!

PR #282, opened 2017-07-04
Trivial: add fonttools to dependencies in README

jonathanrocher:
It seems like it is now needed to compile enable, so this adds it together with Swig and Cython as required dependency.


corranwebster:
LGTM

PR #287, opened 2017-08-01
Improve mouse wheel support

### LINKED ISSUE #134: Handle horizontal mouse wheel events
Currently all mouse wheel events are considered to be vertical, at least in the Qt backend.

### LINKED ISSUE #286: MouseWheel events incompatible between Wx and Qt
I suspect that the issue is on the Wx side, but the `event.mouse_wheel` values generated by Wx are:

- integer values
- seem to be an order of magnitude greater than the values being generated by the same code in Qt for the same amount of motion

It would be good to have these at least be comparable.

corranwebster:
This PR does three things:

- makes scroll wheel rotation values on Wx floating point values.  Fixes #286. 
- adds support for horizontal mouse wheel events via a new `mouse_wheel_axis` trait on `MouseEvents`.  This fixes #134.
- adds tests of the mouse wheel logic for mouse events on Wx and Qt.

### DIFF
--- a/enable/tests/qt4/mouse_wheel_test_case.py
+++ b/enable/tests/qt4/mouse_wheel_test_case.py
@@ -1,0 +1,75 @@
+
+from unittest import TestCase
+
+from mock import MagicMock
+
+from traits.api import Any
+from traitsui.tests._tools import skip_if_not_qt4
+
+from enable.container import Container
+from enable.base_tool import BaseTool
+from enable.window import Window
+
+
+class MouseEventTool(BaseTool):
+
+    event = Any
+
+    def normal_mouse_wheel(self, event):
+        self.event = event
+
+
+@skip_if_not_qt4
+class MouseWheelTestCase(TestCase):
+
+    def setUp(self):
+
+        # set up Enable components and tools
+        self.container = Container(postion=[0, 0], bounds=[600, 600])
+        self.tool = MouseEventTool(component=self.container)
+        self.container.tools.append(self.tool)
+
+        # set up qt components
+        self.window = Window(
+            None,
+            size=(600, 600),
+            component=self.container
+        )
+
+        # Hack: event processing code skips if window not actually shown by
+        # testing for value of _size
+        self.window._size = (600, 600)
+
+    def test_vertical_mouse_wheel(self):
+        from pyface.qt import QtCore, QtGui
+        from enable.qt4.constants import mouse_wheel_axes
+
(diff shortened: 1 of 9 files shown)


enable/tests/qt4/mouse_wheel_test_case.py
```diff
@@ -0,0 +1,75 @@
...
+        self.window._size = (600, 600)
+
+    def test_vertical_mouse_wheel(self):
+        from pyface.qt import QtCore, QtGui
+        from enable.qt4.constants import mouse_wheel_axes
+
+        # create and mock a mouse wheel event
+        qt_event = QtGui.QWheelEvent(
+            QtCore.QPoint(0, 0), 200, QtCore.Qt.NoButton, QtCore.Qt.NoModifier,
+            mouse_wheel_axes[0]
```
jwiggins: Why not `QtCore.Qt.Vertical`?

enable/tests/wx/mouse_wheel_test_case.py
```diff
@@ -0,0 +1,79 @@
...
+        self.window._size = (600, 600)
+
+    def test_vertical_mouse_wheel(self):
+        import wx
+        from enable.wx.constants import mouse_wheel_axes
+
+        # create and mock a mouse wheel event
+        wx_event = wx.MouseEvent(mouseType=wx.wxEVT_MOUSEWHEEL)
+        wx_event.GetWheelRotation = MagicMock(return_value=200)
+        wx_event.GetWheelAxis = MagicMock(return_value=mouse_wheel_axes[0])
```
jwiggins: And presumably we don't test against really old WX versions, what about `wx.MOUSE_WHEEL_VERTICAL` here?

enable/wx/base_window.py
```diff
@@ -352,26 +354,31 @@ def _create_mouse_event ( self, event ):
             x           = event.GetX()
             y           = event.GetY()
             self._last_mouse_pos = (x, y)
-            mouse_wheel = ((event.GetLinesPerAction() *
+
+            mouse_wheel = ((float(event.GetLinesPerAction()) *
```
jwiggins: So do the calculation in floating point and suddenly you match the magnitude of Qt wheel events? Subtle.

corranwebster: Given the quality of the code, it surprised me more than a bit that it Just Worked...

corranwebster:
Issued addressed, tests passing.  Merging.

Issue #221, opened 2016-04-26
Multiple instances of `Undefined behavior: left shift of negative value`


pankajp:
Compiling enable with UBSan `-fsanitize=undefined` http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html spews the following messages when running the demo app and cycling through the demos.

```
$ python examples/demo/demo.py 
loaded the Generic plugin 
kiva/agg/freetype2/src/base/ftstream.c:413:22: runtime error: left shift of negative value -114
kiva/agg/freetype2/src/sfnt/ttmtx.c:195:24: runtime error: left shift of negative value -1
kiva/agg/freetype2/src/sfnt/ttmtx.c:207:16: runtime error: left shift of negative value -1
kiva/agg/freetype2/src/base/ftstream.c:357:22: runtime error: left shift of negative value -1
kiva/agg/freetype2/src/sfnt/ttcmap.c:900:18: runtime error: left shift of negative value -1
kiva/agg/freetype2/src/truetype/ttgload.c:310:25: runtime error: left shift of negative value -1
kiva/agg/freetype2/src/truetype/ttgload.c:492:21: runtime error: left shift of negative value -2
kiva/agg/freetype2/src/truetype/ttgload.c:528:21: runtime error: left shift of negative value -2
kiva/agg/freetype2/src/truetype/ttgload.c:309:25: runtime error: left shift of negative value -1
kiva/agg/agg-24/include/agg_rasterizer_scanline_aa.h:262:56: runtime error: left shift of negative value -61
kiva/agg/agg-24/include/agg_rasterizer_scanline_aa.h:272:55: runtime error: left shift of negative value -256
kiva/agg/freetype2/src/sfnt/ttkern.c:250:25: runtime error: left shift of negative value -1
kiva/agg/agg-24/include/agg_path_storage_integer.h:45:20: runtime error: left shift of negative value -63
kiva/agg/agg-24/include/agg_path_storage_integer.h:44:20: runtime error: left shift of negative value -5
kiva/agg/agg-24/include/agg_rasterizer_scanline_aa.h:262:56: runtime error: left shift of negative value -128
kiva/agg/agg-24/include/agg_rasterizer_scanline_aa.h:272:55: runtime error: left shift of negative value -128
kiva/agg/freetype2/src/truetype/ttgload.c:312:25: runtime error: left shift of negative value -2
[...]

Issue #97, opened 2013-03-26
Kiva AGG renderer breaks when compiled with gcc 4.7+


jwiggins:
https://mail.enthought.com/pipermail/enthought-dev/2013-February/032139.html

corranwebster:
The suspect image accessor functions above are part of the original agg codebase.  It's interesting to note that the forked version of agg 2.4 at http://sourceforge.net/projects/agg/ (warning: on SourceForge) has removed the `const` nature of these constructors (see http://sourceforge.net/p/agg/svn/94/ for the appropriate change) to avoid problems with dangling references.

jdeschenes:
This issue is now resolved with #288.

jwiggins:
This should probably be closed.

jwiggins:
This is fixed.

PR #288, opened 2017-08-07
Upgrade to newer agg version

jdeschenes:
This is a pretty big diff. I don't know if you guys will want to pursue this option. Essentially, there are several issues in enable that are caused by issues with the current agg version in enable. This could close:

- #283 
- #97 
- #221 
- enthought/chaco#295
- enthought/chaco#315

This pull uses the same version of agg as celiagg fixes those issues. While the diff is rather large, it can be summarized and reviewed quite easily:

- The folder agg-24 is identical to the one in celiagg(minus some examples/docs, they can be added)
- Namespace `agg24` was changed to `agg`(To make sure that the two agg were identical)
- changes to `kiva/agg/src/kiva_graphics_context.h` transform_image_interpolate to make it compile
- Changes signature in  `kiva/agg/src/kiva_graphics_context_base.h`  of `rendering_buffer_ptr` to `rendering_buffer`
- Some unittest skip were removed(To account for the failure)
- Some changes to unittest since the image generation is *slightly* different(It looks like they were improved actually)

I have tested this pretty extensively on my Windows machine and so far, there is only one small issue that I saw, where the image doesn't interpolate correctly. My guess is that it is caused by the change to kiva_graphics_context.h. I am sure you guys could spot quickly the error I made.

As far as I am aware, you guys were probably leaning toward using celiagg instead, but this pull could be a good stepping stone in that direction as they would now be sharing the same library underneath.

Let me know if you are interested in pursuing this.


jwiggins:
I think this is really good idea for the medium term. Moving to celiagg is not as straightforward as it should be.

That said, the history of this branch looks a bit crazy. Can you rebase/redo to get rid of the commits from your `python3_compat` branch?

Also, the `agg24` namespace is important. It's meant to keep enable's agg symbols from colliding with matplotlib's agg symbols when both libraries are used in the same process. Crazy, I know. It's a long term dream (pipedream?) of mine to get both libraries to use celiagg.

Also, it might be better to source this from matplotlib, since that's where I grabbed it from when building celiagg.

jwiggins:
One more thing: You could make this diff a bit smaller by avoiding build scripts within agg (`Makefile.*` and `CMakeLists.txt`). They aren't used here anyhow. The relevant directories are:
* kiva/agg/agg-24/font_freetype
* kiva/agg/agg-24/font_win32_tt
* kiva/agg/agg-24/include
* kiva/agg/agg-24/src

Everything else should be ignored

jdeschenes:
I redid a new branch with the same version of AGG as matplotlib. This is a lot cleaner now.

jdeschenes:
There are some failures in the travis build:

1. One is related to the PyQt Scrollbar
2. The others are related to the image interpolation. I think I will need your help on that one(As mentionned previously)

P.S. I will try to setup the appveyor and travis for mac os soon, to make sure it is being tested correctly.

kiva/agg/src/kiva_graphics_context.h
```diff
@@ -588,7 +588,7 @@ namespace kiva
 
 
             agg24::path_storage img_outline = img.boundary_path(img_mtx);
-            agg24::rendering_buffer* src_buf = img.rendering_buffer_ptr();
```
jwiggins: Right. So everywhere you used `agg_pixfmt` in this method, it appears you should have used `other_format` instead. Did that not compile for you?

I've tested this locally (macOS) and everything appears to be working as it always did.

kiva/agg/tests/affine_matrix_test_case.py
```diff
@@ -86,7 +86,7 @@ def test_multiply(self):
 
     def test_determinant(self):
         orig = array((1.0,2.0,3.0,1.0,4.0,5.0))
-        desired = -0.2
+        desired = -5.0
```
jwiggins: Why would this change?

kiva/agg/src/kiva_graphics_context.h
jwiggins: The unit tests pass when I make this change

jwiggins:
@mdickinson: Once the `kiva/agg/src/kiva_graphics_context.h` changes are in place, would you like to try this to see if #283 is still reproduced?

kiva/agg/src/kiva_graphics_context.h
jdeschenes: That's the first thing I tried but it doesn't compile on windows. I get a template error diarrhea. Here is an excerpt.

    kiva\agg\src\kiva_graphics_context.h(711): note: while compiling class template member function 'kiva::graphics_context<agg24::pixfmt_abgr32>::graphics_context(unsigned char *,int,int,int,kiva::interpolation_e)'
    kiva\agg\agg_wrap.cpp(4177): note: see reference to function template instantiation 'kiva::graphics_context<agg24::pixfmt_abgr32>::graphics_context(unsigned char *,int,int,int,kiva::interpolation_e)' being compiled
    kiva\agg\src\kiva_graphics_context.h(618): error C2664: 'agg24::image_accessor_clip<agg24::pixfmt_rgb24>::image_accessor_clip(agg24::image_accessor_clip<agg24::pixfmt_rgb24> &&)': cannot convert argument 1 from 'agg24::pixfmt_abgr32' to 'agg24::pixfmt_alpha_blend_rgb<agg24::blender_rgb24,agg24::rendering_buffer,3,0> &'

kiva/agg/tests/affine_matrix_test_case.py
jdeschenes: In `agg_trans_perspective.h` they had the method `determinant` which was actually a determinant reciprocal. They fixed the method in the new version and added a new method for the reciprocal.

kiva/agg/src/kiva_graphics_context.h
jdeschenes: Oups, brain fart. Looks like it works just fine. Thanks!

jdeschenes:
The image is working perfectly now!

There is a big issue with the scrolling though. I don't think it is related to my branch.

jdeschenes:
False alarm. Enable doesn't like it if using the mouse wheel while pyqt4 and pyqt5 are installed on the same environment.

kiva/agg/tests/affine_matrix_test_case.py
jwiggins: Crazy. Looks like the matrix inversion code also changed, which explains why none of the other unit tests broke...

jwiggins:
I'm generally :+1: on this, but it would be nice if some others look into whether or not it actually solves the crashing bugs with image rendering. I _hope_ that it does, because that would be awesome.

I'll be on vacation for the next three weeks, so if any Enthought persons feel reasonably confident about this, then merge away (don't forget to squash).

itziakos:
Just a comment about licensing. As far as, I understand these changes have been adapted from the matplotlib fork of agg 2.4. Correct? We should make sure we provide the appropriate attribution.

jvkersch:
@itziakos I'm trying to figure out what the right approach is here in terms of licensing. Judging from the Matplotlib license (https://github.com/matplotlib/matplotlib/blob/master/LICENSE/LICENSE) we should be good to go if we reproduce the Matplotlib license along with the Agg source code (and perhaps add a clarifying note somewhere that the Matplotlib license applies only to Agg sources). Do you agree?

itziakos:
My take on this is that the agg source at the matplotlib repo has a separate license https://github.com/matplotlib/matplotlib/blob/master/extern/agg24-svn/src/copying. I think we should append add a copy of that license in our agg source license and mention in a note that changes added in this PR have been ported from the matplotlib fork of agg (link to matplotlib).

Maybe, we should check with @jwiggins  and @corranwebster, they might have some more experience on such things.

corranwebster:
I'm a little unclear on exactly what the origins of the code in this branch are, but I think it's something like:

- take the matplotlib agg code
- merge in some fixes from the old agg version that we shipped in Kiva
- additional fixes as needed to get the kiva SWIG wrappers to work

If this is the case, then we need to:

- include the file https://github.com/matplotlib/matplotlib/blob/master/extern/agg24-svn/src/copying
- _if_ the matplotlib team made any modifications to the agg code (which is currently unclear to me), then we also need to follow the matplotlib license
- include a BSD license with appropriate contributors for the original fixes + new changes.

Does this seem reasonable/correct?

jdeschenes:
The approach was:

1. Take the matplotlib code
2. Fix the namespacing issue(agg vs agg24)
3. Fix the wrapper

Also, there were extra modifications in the matplotlib repository.

jvkersch:
This fixes https://github.com/enthought/chaco/issues/339 for me (segfault before, nice image plot after), nice work!

As for licensing, since there are changes made by the matplotlib community, it does look like we need to include the matplotlib license as well (along with a brief description of the changes made in this PR). @corranwebster I'm not sure what your 3rd point would look like (BSD license with appropriate contributors...). Can you follow up on this?

corranwebster:
Yes, I agree, looks like we need the matplotlib license.  The 3rd point would just be a copy of the main Enable project LICENSE.txt with appropriate dates (ie. should probably update copyright 2006-17 or something).

I'd also suggest adding a README.txt to enable/kiva/agg with a brief explanation of the provenance of the code.  That should hopefully cover the licensing aspects.

We can possibly do this in a separate PR as long as we are good about doing it immediately upon the merge of this code.

jvkersch:
@jdeschenes Would you be able to take care of the remaining license issues?

jwiggins:
For AGG, are we not already covered by https://github.com/enthought/enable/blob/master/kiva/agg/agg-24/copying?

Matplotlib _did_ make changes to the AGG source code which we are getting here. So I think we now need to include https://github.com/matplotlib/matplotlib/blob/master/LICENSE/LICENSE as well. Where shall it go? Is it time to add a LICENSE directory to enable?

jvkersch:
@jdeschenes @corranwebster @jwiggins I added the licensing info in a separate PR https://github.com/jdeschenes/enable/pull/1 against @jdeschenes' branch. Let me know what you think.

jdeschenes:
It works for me... I just accepted your modifications.

corranwebster:
Licensing LGTM.  Unless anyone has objections, I think that this is good to go.

jwiggins:
Thanks for clearing that up @jvkersch, and thanks to @jdeschenes for the upgrade!

PR #294, opened 2017-10-04
DOC: Update docs for 4.7.0 release

jvkersch:


_static/alabaster.css
```diff
@@ -15,19 +15,55 @@
...
+
+
+
+
+
+
+
+
+
+
```
jwiggins: strange...

PR #295, opened 2017-10-30
MAINT: Use uint8_t from numpy instead of stdint.h

jvkersch:
The Cython module `kiva._cython_speedups` was using the stdint.h header, which is not shipped with VS 2008 (needed for Python 2.7). This replaces it by the equivalent declaration from Numpy.


kiva/_cython_speedups.pyx
```diff
@@ -67,7 +67,7 @@ def points_in_polygon(pts, poly_pts, use_winding=False):
 
     cdef double[:, ::1] pts_view = pts
     cdef double[:, ::1] poly_pts_view = poly_pts
-    cdef uint8_t[::1] results = np.zeros(len(pts), dtype=np.uint8)
+    cdef np.uint8_t[::1] results = np.zeros(len(pts), dtype=np.uint8)
```
jwiggins: Are you sure that shouldn't be `np.npy_uint8`?

kiva/_cython_speedups.pyx
```diff
@@ -1,5 +1,5 @@
-from libc.stdint cimport uint8_t
 import numpy as np
+cimport numpy as np
```
rkern: Style nit: I may be alone in this (certainly the Cython numpy tutorials do it like you do), but I don't like `import numpy as np` and `cimport numpy as np`. I usually `cimport numpy as cnp` to distinguish the later references.

Or in this case, since you are importing only one thing: `from numpy cimport uint8_t`

Either way, LGTM.

kiva/_cython_speedups.pyx
jvkersch: I think that magically transmogrifies to `npy_uint8` somewhere deep under the Cython hood...

jwiggins:
We should probably get AppVeyor hooked up for this project...

jvkersch:
> We should probably get AppVeyor hooked up for this project...

Agreed, and same thing for Chaco. I ran into this issue (and a similar issue for Chaco) while trying to build Python 2.7 eggs on Windows...

kiva/_cython_speedups.pyx
jvkersch: https://github.com/cython/cython/blob/master/Cython/Includes/numpy/__init__.pxd#L750

jwiggins: Indeed. Thanks for clearing that up

jvkersch:
Thanks guys, merging.

PR #299, opened 2017-11-29
BUG: Remove use of C API not available in Python 3.

rkern:
Importing `kiva.quartz.ABCGI` will fail otherwise.


jvkersch:
LGTM. Could you add a test to import the module?

jwiggins:
LGTMT

PR #302, opened 2018-04-13
Qt5 Support

corranwebster:
Now we have a released Pyface with Qt5 support, it makes sense to see what is needed to support Qt5.

This PR:

- [x] Integrates PyQt5 into the CI testing system.
- [x] Fixes mouse wheel support.
- [x] Fixes dangling NativeScrollbar references on qt4.


enable/tests/qt4/mouse_wheel_test_case.py
```diff
@@ -43,31 +43,55 @@ def setUp(self):
...
+                QtCore.Qt.NoModifier, QtCore.Qt.Vertical
+            )
+        else:
+            qt_event = QtGui.QWheelEvent(
+                QtCore.QPointF(0, 0),
+                self.window.control.mapToGlobal(QtCore.QPoint(0, 0)),
+                QtCore.QPoint(0, 200), QtCore.QPoint(0, 200.0/120), 200,
+                QtCore.Qt.Vertical, QtCore.Qt.NoButton, QtCore.Qt.NoModifier,
+                QtCore.Qt.ScrollUpdate
+            )
```
jwiggins: I feel like this code should be in a pyface testing support module... But I don't see any such module in pyface.

corranwebster: Pyface doesn't really do any testing at the level of individual mouse events (things are usually driven at the level of signals and slots), so we haven't had a need to put anything like this into Pyface.

For now it's probably good enough to have it as part of `enable.tests.qt4`; if we find that we are repeating this a lot we might want to create a `enable.qt4.testing.*` utility method.

jwiggins: Sounds good to me

corranwebster:
I think this is good to merge.

kitchoi:
Apparently the two tests on wx are failing (but the job has been allowed to fail on CI):
```
FAIL: test_horizontal_mouse_wheel (enable.tests.wx.mouse_wheel_test_case.MouseWheelTestCase)
  File "/home/travis/.edm/envs/enable-test-2.7-wx-pillow/lib/python2.7/site-packages/enable-4.7.1.dev15-py2.7-linux-x86_64.egg/enable/tests/wx/mouse_wheel_test_case.py", line 82, in test_horizontal_mouse_wheel
    self.assertEqual(self.tool.event.mouse_wheel_delta, (200, 0))
AssertionError: <undefined> != (200, 0)
FAIL: test_vertical_mouse_wheel (enable.tests.wx.mouse_wheel_test_case.MouseWheelTestCase)
  File "/home/travis/.edm/envs/enable-test-2.7-wx-pillow/lib/python2.7/site-packages/enable-4.7.1.dev15-py2.7-linux-x86_64.egg/enable/tests/wx/mouse_wheel_test_case.py", line 63, in test_vertical_mouse_wheel
    self.assertEqual(self.tool.event.mouse_wheel_delta, (0, 200))
AssertionError: <undefined> != (0, 200)
```

The CI job for wx is very close to passing. If we can require wx job to pass, we will be able to catch these more easily.

I tried to fix these by assigning the `mouse_wheel_delta` to the event object where they are defined.  However the values are not the expected value. I get (1.0, 0) for horizontal case, and (0, 1.0) for vertical case.

PR #310, opened 2018-07-06
Remove locale-dependence in Agg text rendering

rkern:
To handle rendering Unicode text strings in the `agg` backend, we explicitly encode as UTF-8 byte strings in the Python layer before passing the encoded `char*` to the C++ API. On Unix platforms currently, it uses locale-dependent functions to decode the bytes back into codepoint integers to pass to the font engine. Under a non-UTF-8 locale, we render mojibake.

This PR vendorizes a liberally-licensed header-only library for explicitly decoding the UTF-8 bytes in the C++ layer.


kiva/agg/src/kiva_graphics_context.h
```diff
@@ -1292,21 +1294,14 @@ namespace kiva
...
-          }
-        wchar_t *p = &p_[0];
-#endif
+
+        // Explicitly decode UTF8 bytes to 32-bit codepoints to feed into the
+        // font API.
+        std::vector<utf8::uint32_t> codepoints;
+        std::vector<utf8::uint32_t>::iterator p;
+        std::string utf8text(text);
+        utf8::utf8to32(utf8text.begin(), utf8text.end(), std::back_inserter(codepoints));
```
jwiggins: I believe you can just construct a `utf8::iterator` and avoid having to copy into `codepoints`. Additionally, if you get the length of `text`, I think you can construct the iterator like so: `utf8::iterator<char*> p(text, text, text+length);`. If that works, it would be nice to avoid the allocations.

rkern:
I'm not too concerned about the exceptions since we are guaranteed that the input is properly-terminated UTF-8 encoded text.

jwiggins:
Sounds good to me. Thanks for the update!

PR #317, opened 2018-09-01
Fix rect drawing with celiagg

jwiggins:
Component borders were a bit fuzzy when drawn with celiagg thanks to a very specific optimization which was made in the kiva agg backend.

This replicates that in the celiagg backend as well.

### DIFF
--- a/kiva/celiagg.py
+++ b/kiva/celiagg.py
@@ -363,7 +363,7 @@ class GraphicsContext(object):
         if (not self.canvas_state.anti_aliased and
                 self.canvas_state.line_width in (0.0, 1.0) and
                 fabs(self.transform.shx) < 1e-3 and
-                fabs(self.transform.shx) < 1e-3):
+                fabs(self.transform.shy) < 1e-3):
             scale_x = self.transform.sx
             scale_y = self.transform.sy
             tx = self.transform.tx
--- a/kiva/celiagg.py
+++ b/kiva/celiagg.py
@@ -11,6 +11,7 @@
 from __future__ import absolute_import, print_function, division
 
 from collections import namedtuple
+from math import fabs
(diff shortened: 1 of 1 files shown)


jwiggins:
This was mentioned by a user in celiagg/celiagg#43

corranwebster:
Hmmm... Chaco _also_ does a half-pixel adjustment here https://github.com/enthought/chaco/blob/master/chaco/plot_graphics_context.py#L28 but I don't know that we get that class unless we specifically ask for it.

I don't like the ad-hoc half-pixel shifts.  Not saying that we shouldn't do it to keep backwards consistency, but it bugs me.

Beyond that the code LGTM.

kiva/celiagg.py
```diff
@@ -353,11 +354,33 @@ def rects(self, rects):
...
+
+        # XXX: kiva::graphics_context<>::_draw_rect_simple() does a VERY
+        # specific optimization for drawing rectangles in certain circumstances
+        # which results in chaco plot borders which are sharp.
+        # This implements that same special case.  - JW 2018/09/01
+        transform = self.transform
+        if (not self.canvas_state.anti_aliased and
+                self.canvas_state.line_width in (0.0, 1.0) and
+                fabs(self.transform.shx) < 1e-3 and
+                fabs(self.transform.shx) < 1e-3):
```
jwiggins: Typo: One of these should be `self.transform.shy`

corranwebster:
I'm happy for this to be merged with the typo corrected.

kiva/celiagg.py
corranwebster:
```suggestion
                fabs(self.transform.shy) < 1e-3):
```

PR #319, opened 2018-11-14
Include licenses and other files in sdists

toddrme2178:
The license requires that the license be included in all copies of the program.  Also include some other documentation files.


jwiggins:
Thanks for the patch!

PR #321, opened 2018-11-21
Don't clobber a passed-in font object

### LINKED ISSUE #320: PDF backend mutates passed in font
[copied from enthought/chaco#428]

**Problem Description**

When using the `PdfPlotGraphicsContext` to export to PDF, the `draw` function executed in the `render_component` function alters the plot font.

**Reproduction Steps:**

Code to reproduce issue:

```python
from traits.api import *
from traitsui.api import *
from chaco.api import *
from chaco.pdf_graphics_context import PdfPlotGraphicsContext


class PlotContainer(HasTraits):

    plot = Instance(Plot)
    save = Button

    def _plot_default(self):
        data = ArrayPlotData(x=[1, 2, 3, 4], y=[2, 3, 5, 8])
        p = Plot(data)
        p.plot(('x', 'y'))
        return p

    def _save_fired(self):

        # Do save stuff ...
        self.plot.do_layout(force=True)
        gc = PdfPlotGraphicsContext(filename='./plot.pdf')
        gc.render_component(self.plot)
        gc.save()

        print('plot saved!')

    traits_view = View(
        VGroup(
            UItem('plot', editor=ComponentEditor()),
            UItem('save'),
        )
    )


if __name__ == '__main__':
    pc = PlotContainer()
    pc.configure_traits()
```

**Expected behavior:**

Font should remain the same after focus is brought back to window, but it does not.

**OS, Python version:**
Occurs on both Ubuntu 18 and windows 7

```
$ python -V
Python 2.7.13 -- Enthought, Inc. (x86_64)

$ edm list | grep chaco
chaco                               4.7.2-1        enthought/free
```

jwiggins:
Fixes #320

### DIFF
--- a/kiva/pdf.py
+++ b/kiva/pdf.py
@@ -623,9 +623,10 @@ class GraphicsContext(GraphicsContextBase):
         """ Sets the font for the current graphics context.
         """
         # TODO: Make this actually do the right thing
-        if font.face_name == "":
-            font.face_name = "Helvetica"
-        self.gc.setFont(font.face_name, font.size)
+        face_name = font.face_name
+        if face_name == "":
+            face_name = "Helvetica"
+        self.gc.setFont(face_name, font.size)
 
     def get_font(self):
         """ Get the current font """


kiva/pdf.py
```diff
@@ -623,9 +623,10 @@ def set_font(self, font):
         """ Sets the font for the current graphics context.
         """
         # TODO: Make this actually do the right thing
```
cfarrow: Is this TODO still relevant? Perhaps the right thing is not choosing Helvetica.

cfarrow:
LGTM, pending tests.

kiva/pdf.py
jwiggins: I have no idea what the context was for this comment... But as I'm in Switzerland right now, I'm pretty sure I'll get kicked out if I use anything but Helvetica (Arial doesn't count!)

PR #323, opened 2018-12-03
Fix bugs and update VTK backend.

prabhuramachandran:
- An earlier merge in 2011 seems to have introduced some incorrect code
  leading to the key events being incorrectly handled.
- Updating to use newer (i.e. VTK >= 6) versions.

### DIFF
--- a/enable/vtk_backend/vtk_window.py
+++ b/enable/vtk_backend/vtk_window.py
@@ -130,9 +130,8 @@ class EnableVTKWindow(AbstractWindow, CoordinateBox):
         self._add_observer(istyle, "MouseWheelForwardEvent", self._vtk_mouse_wheel)
         self._add_observer(istyle, "MouseWheelBackwardEvent", self._vtk_mouse_wheel)
 
-        self._add_observer(istyle, "KeyPressEvent", self._on_key_pressed)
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
@jwiggins -- please take a look, thanks.

jwiggins:
So wait... You reverted the changes made in https://github.com/enthought/enable/commit/fd962fe3c263d07b8401e347bc3c1b924e2f38d3#diff-d668a32389ee62bfb3d18ef18ad6125e and it works now?

prabhuramachandran:
Well, the key events don't complain, I also added a couple of smaller changes to update to more recent VTK's, maybe I should have made two commits?   See line 420 and 440.  

But with this, the chaco examples seem to run with mayavi master and VTK-8.1.1.  Not sure if anything else is broken.

prabhuramachandran:
Sure, so we just need to move the directory to "vtk"?  You want me to do it or should it be done in a separate PR?

prabhuramachandran:
Whats with the tests?  Who is to merge this PR?  I will send the next PR (change backend directory) once this is merged I guess.

corranwebster:
Pyface issue is interesting: the failing line is an addition from an outside dev to force loading of at least something from the underlying Qt API to fix incorrect selection.  I think that needs re-visiting.

PR #325, opened 2018-12-28
Use non-toolkit-specific Timer class

### LINKED ISSUE #324: HoverTool broken with Pyface master
In enthought/pyface#340, the toolkit-specific `timer.timer.Timer` classes were removed, making https://github.com/enthought/enable/blob/6008c7e204741e6833fe6b760b056ae4c0a7f058/enable/tools/hover_tool.py#L152-L153 fail with a `NotImplementedError`.

```
$ ETS_TOOLKIT=qt4; QT_API=pyqt; python -c "from pyface.toolkit import toolkit_object; klass = toolkit_object('timer.timer:Timer'); klass()"
  File "/Users/skern/src/pyface/build/lib/pyface/base_toolkit.py", line 178, in __init__
    raise NotImplementedError(msg % (toolkit, package, name))
NotImplementedError: the qt4 pyface.ui.qt4 backend doesn't implement timer.timer:Timer
```

stevenjkern:
Closes #324

I still need to write some tests for this tool, so that will take a few days at the least.

### DIFF
--- a/enable/tools/hover_tool.py
+++ b/enable/tools/hover_tool.py
@@ -1,3 +1,6 @@
+# (C) Copyright 2008-2019 Enthought, Inc., Austin, TX
+# All rights reserved.
+
 """
 Tool to detect when the user hovers over a specific part of an underlying
 components.
@@ -9,7 +12,8 @@ from __future__ import absolute_import
 from enable.base_tool import BaseTool
 from traits.api import Any, Callable, Enum, Float, Int
 from traits.etsconfig.api import ETSConfig
-from pyface.timer.api import Timer
+from pyface.timer.api import DoLaterTimer
+
 
 # Define a toolkit-specific function for determining the global mouse position
 if ETSConfig.toolkit == 'wx':
@@ -103,15 +107,11 @@ class HoverTool(BaseTool):
             # update xy and restart the timer
             self._start_xy = GetGlobalMousePosition()
             self.restart_hover_timer(event)
-        else:
-            if self._timer:
-                self._timer.Stop()
 
     def restart_hover_timer(self, event):
         if self._timer is None:
             self._create_timer(event)
-        else:
-            self._timer.Start()
+        self._timer.start()
 
     def on_timer(self):
         position = GetGlobalMousePosition()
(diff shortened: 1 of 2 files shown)


corranwebster:
Fix LGTM.

The backend specific code for getting the global mouse position would be much better if it were in `enable.qt.*` and `enable.wx.*` or even better somewhere like `pyface.gui` but that's beyond the scope of this PR.

enable/tools/hover_tool.py
```diff
@@ -149,5 +149,4 @@ def _is_in(self, x, y):
             return any((t, b, r, l))
 
     def _create_timer(self, event):
-        klass = toolkit_object("timer.timer:Timer")
-        self._timer = klass(self.hover_delay, self.on_timer)
+        self._timer = Timer(self.hover_delay, self.on_timer)
```
corranwebster: Actually, on second thoughts,
```suggestion
        self._timer = do_after(self.on_timer, self.hover_delay)
```
is probably slightly preferable (see https://github.com/enthought/pyface/blob/master/pyface/timer/do_later.py#L45), and should work across new and old versions of Pyface timers.

If you do this, you will need to remove the call to `self._timer.Stop()` in `on_timer`

enable/tools/hover_tool.py
```diff
@@ -7,9 +7,9 @@
 
 # Enthought library imports
 from enable.base_tool import BaseTool
-from traits.etsconfig.api import ETSConfig
-from pyface.toolkit import toolkit_object
 from traits.api import Any, Callable, Enum, Float, Int
+from traits.etsconfig.api import ETSConfig
+from pyface.timer.api import Timer
```
corranwebster: And this would become:
```suggestion
from pyface.timer.api import do_after
```

enable/tools/hover_tool.py
stevenjkern: Thanks for all of this guidance, @corranwebster. Unfortunately, the direct replacement with `do_after` didn't work quite as hoped, since the [old version of `do_after`](https://github.com/enthought/pyface/blob/14c26736ebc3e245cd8e6d469d995d8265873f81/pyface/timer/do_later.py#L33-L45) didn't actually return the timer instance, so `self._timer` stayed `None`.

I think it's probably safe to remove the `_timer` attribute entirely and create a new timer on each hover event, but I wasn't sure, so I opted on the side of changing as little as possible and set `self._timer` to be a `DoLaterTimer` instance.

I have a few more test cases I'd like to add to flesh out the coverage on the conditionals.

stevenjkern:
Alright, I feel a lot better about this now. I've increased test coverage to cover just about everything on the tool, including identifying and fixing a bug fix with a hover in the lower-right corner not being picked up (confusion in naming between `"l"` meaning either "lower" or "left"). I think this is ready for another look if you have any bandwidth, @corranwebster.

stevenjkern:
I think I finally have all of my troubles with this PR worked out. @corranwebster, if you have any time available, could you give this another review?

corranwebster:
LGTM

PR #327, opened 2019-01-15
Use correct form of signal linkage for Qt5

silverjam:
In testing Qt5 support the `enable` package throws and error when calling `QtCore.QObject.connect` directly.  Instead use `control.clicked.connect` to connect to the `clicked()` signal.


corranwebster:
Thanks for the PR, and good catch!

PR #330, opened 2019-02-16
Add support for ttc fonts

### LINKED ISSUE #265: Add support for .ttc fonts
After #264 is merged, we can read fonts from `.ttc` (TrueType collection files), but currently we don't have the support within the font manager to do this.  This would increase font availability on OS X, particularly for non-English fonts.

JCorson:
Based on suggestions from @rkern, I think this is all that is needed to add support for ttc font collections.

We ran into this on a project where we needed to support font collections such as meiryo.ttc on Windows. 

Only tested this on Windows , Python 3 so far. 

Closes #265 (I think?)


corranwebster:
Can you add a test, even if it is a just smoke test?

kiva/fonttools/font_manager.py
```diff
@@ -670,6 +671,27 @@ def afmFontProperty(fontpath, font):
...
+    """
+    Extract a list of ttfFontProperty objects from a ttc file
+    """
+    props = []
+    try:
+        collection = TTCollection(str(fpath))
+    except (RuntimeError, TTLibError):
+        verbose.report("Could not open font collection %s" % fpath)
+        return props
+    except UnicodeError:
```
corranwebster: Is this problem with Unicode limitation of fonttools, Windows, and/or the particular Python version you are using?

kiva/fonttools/font_manager.py
```diff
@@ -670,6 +671,27 @@ def afmFontProperty(fontpath, font):
 
     return FontEntry(fontpath, name, style, variant, weight, stretch, size)
 
+def extractTTC(fpath):
+    """
+    Extract a list of ttfFontProperty objects from a ttc file
+    """
+    props = []
+    try:
```
corranwebster: I'd be tempted to put this try-except in `createFontList` below - have this function either return a correct list or fail; but the code in `createFontList` has to execute a loop and so it shouldn't fail on a particular loop and should be what worries about catching errors.

JCorson: Makes sense to me. Fixed in 533401c

kiva/fonttools/font_manager.py
JCorson: No clue. I assume there was some problem with fonttools itself that we are trying to avoid. I see the same check in matplotlib's code. Somewhat moot with 97d9c6a

JCorson:
> Can you add a test, even if it is a just smoke test?

Sure can and thanks for the review @corranwebster. I couldn't find where this module is tested directly, so I added a quick test module. I did some mocking to exercise the logic added here. This much mocking does make me a little wary, especially adding the `.fonts` attribute. We could add some actual font files to test with, but that may be more than you had in mind.

Let me know what you think.

corranwebster:
I'd be fine with adding a small font file to test with (`fonttools` includes some here https://github.com/fonttools/fonttools/tree/master/Tests/ttx/data).  We do a similar thing with some image files here https://github.com/enthought/enable/tree/master/enable/tests/primitives/data/PngSuite for example.

kiva/fonttools/font_manager.py
corranwebster: Ah - it's possible that this is no longer relevant.  We used to package our own version of fonttools, but now we use a 3rd party wrapper.  It's fine to leave in here, but it may be that since that switch we no longer will get the error.

kiva/fonttools/font_manager.py
```diff
@@ -670,6 +671,20 @@ def afmFontProperty(fontpath, font):
...
+    """
+    Extract a list of ttfFontProperty objects from a ttc file
+    """
+    collection = TTCollection(str(fpath))
+
+    props = []
+    for font in collection.fonts:
+        try:
+            props.append(ttfFontProperty(fpath, font))
+        except:
```
corranwebster: Bare `except`s are generally not a good idea unless you are raising immediately.  I think it is fair to say that if you can't read all the fonts from the `.ttc` file successfully it's OK to just fail for the entire file - who knows what else might fail down the line on this file.  So removing the `try` ... `except` completely here is probably the right thing.

JCorson: Yup, but then I think we need the bare `try`... `except` in `createFontList` so that the loop continues (similar to what is done in `createFontList`). At that point, trying to break this into its own function isn't seeming all that worth it. I moved this code into `createFontList` where the bare `try`...`except` may seem less bad... maybe...

JCorson:
@corranwebster Test font file added (swiped from fonttools) and more tests added for exception handling. Should we acknowledge where the test font file came from in a little text doc inside the `tests` folder?

corranwebster:
Yes, please add a file with the provenance, and perhaps put the fonttools `LICENSE.txt` there as well.

kiva/fonttools/font_manager.py
corranwebster: So there are two concerns here:

- first is that you should almost never to a bare `except` unless you are re-raising.  If you need to swallow exceptions you should:
   * do `except Exception` so that you _don't_ catch things like `SystemExit` and `KeyboardInterrupt`
   * log something about the error that occurred to help with debugging (why didn't my fonts load?)
- the second is that if the font file is corrupted enough that you can't read all the font, you're likely to get other issues with the fonts in the file, so I'd be tempted to just ignore everything from that file - treat the whole thing as bad.  The easiest way to do this is just to let the exception propagate and not return anything.

setup.py
```diff
@@ -286,7 +286,8 @@ def run(self):
               '': ['*.zip', '*.svg', 'images/*'],
               'enable': ['tests/primitives/data/PngSuite/*.png'],
               'enable.savage.trait_defs.ui.wx': ['data/*.svg'],
-              'kiva': ['tests/agg/doubleprom_soho_full.jpg'],
+              'kiva': ['tests/agg/doubleprom_soho_full.jpg',
+                       'fonttools/tests/data/TestTTC.ttc'],
```
corranwebster: Also the license file and source.txt (possibly should be doing that in other places as well, but don't worry about that in this PR).

Also make sure that the `MANIFEST.in` is catching these files as well.

kiva/fonttools/font_manager.py
corranwebster: I'll leave the question of further drive-by clean-up to you.  Welcome if you want to do it, but I don't want to divert effort if this PR is on a critical path for your work.

kiva/fonttools/font_manager.py
```diff
@@ -707,14 +710,31 @@ def createFontList(fontfiles, fontext='ttf'):
...
+                if ext.lower() == ".ttc":
+                    collection = TTCollection(str(fpath))
+                    try:
+                        props = []
+                        for font in collection.fonts:
+                            props.append(ttfFontProperty(fpath, font))
+                        fontlist.extend(props)
+                        continue
+                    except Exception:
+                        logger.error(
```
corranwebster: A couple of tricks for logging:

- logging does '%' substitution for you automatically(!)
- if you do `exc_info=True` as well it will dump the exception data to the log, which can also be useful.

so you get:
```
logger.error("Could not covert font to FontEntry for file %s", fpath, exc_info=True)
```

corranwebster:
A couple of things to tighten up but this is looking pretty good at this point.

kiva/fonttools/font_manager.py
JCorson: Sweet! Thanks!

kiva/fonttools/font_manager.py
JCorson: I went ahead and replaced `verbose` with actual logging. I went ahead and did the extra clean up and replaced `verbose` with actual logging. Might as well take care of it now. 😄

kiva/fonttools/tests/test_font_manager.py
```diff
@@ -0,0 +1,46 @@
+import os
+import unittest
+try:
+    from unittest import mock
+except:
```
jwiggins: `except ImportError:`

JCorson: Good catch! Fixed in e60b763

JCorson:
@corranwebster @jwiggins Sorry all, I lost track of this. Do either of you have time for another look?

corranwebster:
LGTM

PR #332, opened 2019-03-18
Add a maximum length to the damaged regions list to curb memory hogging

tn-exafore:
Otherwise the damaged regions list of components that have been shown at some point but are no longer visible and thus do not receive paint events can grow arbitrarily long, consuming large amounts of memory. This memory would be released if the component became visible again but who knows when/if that happens.

In our application we have TraitsUI tabs that contain Chaco plots whose PlotDatas are updated regardless of the current tab, and the plots in tabs that are not currently visible can "leak" several megabytes per minute.

I'm not familiar with the innards of Chaco/Enable, and this is just a quick solution proposal, so if you don't like this idea, feel free to bake a better one.

An example program showing the problem (I'm using pyqt backend but I don't think it matters):

1. Run for a while keeping to the default tab (1). 
2. Once you click on tab 2, it starts leaking memory from tab 1,
3. Click on tab 3, it starts leaking memory from both tabs 1 and 2. 
4. After running for a while that way leaking memory, click on tab 1 or 2, and it will release the "leaked" memory taken by that tab's plot as it gets a paint event 

```
import threading, time, random
import numpy as np

from traits.api import HasTraits, Instance, Str
from traitsui.api import View, Tabbed, Item, UItem
from pyface.api import GUI
from chaco.api import ArrayPlotData, Plot

n_data = 100
y1 = np.random.rand(n_data)
y2 = np.random.rand(n_data)

class LeakyPlot(HasTraits):
    plot1 = Instance(Plot)
    plot_data1 = Instance(ArrayPlotData)

    plot2 = Instance(Plot)
    plot_data2 = Instance(ArrayPlotData)

    tab3 = Str

    traits_view = View(
        Tabbed (
            UItem('plot1', editor=ComponentEditor()),
            UItem('plot2', editor=ComponentEditor()),
            Item("tab3"),
        )
    )

    def __init__(self):
        super(LeakyPlot, self).__init__()

        self.plot_data1 = ArrayPlotData(y=y1)
        self.plot_data2 = ArrayPlotData(y=y2)

        self.plot1 = Plot(self.plot_data1)
        self.plot1.plot("y", type = "line", name="plot", color = "blue")
        self.plot1.range2d.low_setting = ('auto', 0.0)
        self.plot1.range2d.high_setting = ('auto', 1.0)

        self.plot2 = Plot(self.plot_data2)
        self.plot2.plot("y", type = "line", name="plot", color = "red")
        self.plot2.range2d = self.plot1.range2d

    def update_plot(self):
        with lock:
            self.plot_data1.update_data(y=y1)
            self.plot_data2.update_data(y=y2)


def update_data(plot):
    while not quit:
        time.sleep(0.001)
        with lock:
            for i in range(0, len(y1)):
                y1[i] = random.random()
            for i in range(0, len(y2)):
                y2[i] = random.random()
        GUI.invoke_later(plot.update_plot)

quit = False
lock = threading.Lock()

leaky_plot = LeakyPlot()
data_thread = threading.Thread(target=update_data, name="data updater", args=(leaky_plot,))
data_thread.start()

leaky_plot.configure_traits()

quit = True
data_thread.join()
```

### DIFF
--- a/enable/abstract_window.py
+++ b/enable/abstract_window.py
@@ -76,6 +76,8 @@ class AbstractWindow(HasTraits):
     # The regions to update upon redraw
     _update_region = Any
 
+    # When exceeding this, the entire window is marked damaged to save memory
+    MAX_DAMAGED_REGIONS = 100
 
 
     #---------------------------------------------------------------------------
@@ -232,7 +234,8 @@ class AbstractWindow(HasTraits):
         return
 
     def invalidate_draw(self, damaged_regions=None, self_relative=False):
-        if damaged_regions is not None and self._update_region is not None:
+        if damaged_regions is not None and self._update_region is not None \
+                and len(self._update_region) < self.MAX_DAMAGED_REGIONS:
             self._update_region += damaged_regions
         else:
             self._update_region = None


jwiggins:
Thanks for taking a deeper look at this issue! 

I've looked around the code a bit and it _seems_ like the bug is that the `use_damanged_region` is being ignored. Does it also fix the memory leak if you honor the `use_damaged_region` flag in `AbstractWindow.invalidate_draw`?

tn-exafore:
I guess adding a check for the `use_damaged_region` flag in `AbstractWindow.invalidate_draw` would work too, but only if it is set off, and shouldn't in principle both states of the flag be supported and not leak memory?

jwiggins:
Yes, that's a good point. When `use_damaged_region` is true, you will still encounter the problem addressed by this PR.

However I'm starting to think that we should maybe just remove this code altogether. You are likely right that it's a work in progress. Grepping the code for both chaco and enable leads me to believe that `use_damaged_region` _is always false_ (unless user code explicitly sets it to true).

@rkern Do you remember the provenance of this feature? Should we keep the traits for backwards compatibility and remove all the code in methods which references them? (`use_damaged_region` and `_update_region`)

tn-exafore:
added some example code to the PR description that shows the problem (but I think you already got the idea)

rkern:
Yeah, the intention was to avoid having to expensively repaint the entire control every time there was an update to a small part of it. The feature never got completed, but the infrastructure was left in. I'd be okay ripping it out. We've never really missed it. I imagine that it'd be complicated and error-prone to actually use in any case.

corranwebster:
Based on @rkern's comments and offline discussion with @jwiggins I'd be comfortable with removal of the damage region code completely, although that's a more complex task than this PR.

tn-exafore:
I'd of course be happy with that solution, too, removing the non-working damage region code entirely but yeah, out of scope for this PR. With what kind of schedule are you planning to do it?

corranwebster:
Probably not soon, unfortunately.  I'd be happy with this to be merged as a stopgap if that helps - @jwiggins is the maintainer of the codebase, so it is his call.

jwiggins:
Yes, I agree. I'll open an issue for the removal.

jwiggins:
Thanks again @tn-exafore!

tn-exafore:
Happy to help! This as a stopgap and a more thorough cleanup later on sounds like a good plan to me.

Issue #329, opened 2019-01-28
pip install enable results in clang error: unknown argument: '-framework ApplicationServices'


gpanders:
I have installed all the prerequisites as listed in the README, but am getting the following error:

```make
clang++ -bundle -undefined dynamic_lookup -L/usr/local/opt/readline/lib -L/usr/local/opt/readline/lib -L/usr/loca
l/opt/openssl/lib -L/Users/greande/.pyenv/versions/3.6.7/lib -L/usr/local/opt/readline/lib -L/usr/local/opt/readline/
lib -L/usr/local/opt/openssl/lib -L/Users/greande/.pyenv/versions/3.6.7/lib build/temp.macosx-10.13-x86_64-3.6/build/
src.macosx-10.13-x86_64-3.6/kiva/agg/agg_wrap.o -Lbuild/temp.macosx-10.13-x86_64-3.6 -lkiva_src -lagg24_src -lfreetyp
e2_src -o build/lib.macosx-10.13-x86_64-3.6/kiva/agg/_agg.cpython-36m-darwin.so -framework ApplicationServices -frame
work OpenGL -framework Carbon                                                                                       
    clang: error: unknown argument: '-framework ApplicationServices'
    clang: error: unknown argument: '-framework OpenGL'
    error: Command "clang++ -bundle -undefined dynamic_lookup -L/usr/local/opt/readline/lib -L/usr/local/opt/readline
/lib -L/usr/local/opt/openssl/lib -L/Users/greande/.pyenv/versions/3.6.7/lib -L/usr/local/opt/readline/lib -L/usr/loc
al/opt/readline/lib -L/usr/local/opt/openssl/lib -L/Users/greande/.pyenv/versions/3.6.7/lib build/temp.macosx-10.13-x
86_64-3.6/build/src.macosx-10.13-x86_64-3.6/kiva/agg/agg_wrap.o -Lbuild/temp.macosx-10.13-x86_64-3.6 -lkiva_src -lagg
24_src -lfreetype2_src -o build/lib.macosx-10.13-x86_64-3.6/kiva/agg/_agg.cpython-36m-darwin.so -framework Applicatio
nServices -framework OpenGL -framework Carbon" failed with exit status 1                                            
    
    ----------------------------------------
Command "/Users/greande/.pyenv/versions/3.6.7/envs/pytraining/bin/python -u -c "import setuptools, tokenize;__file__=
[...]

PR #335, opened 2019-04-05
Fix macOS framework linker args

silverjam:
Without this change we were seeing build failures that stated `-framework OpenGL` was an unknown compiler option.  So, the theory is, something changed about how parameters were being passed between Python versions, so "-framework OpenGL" would be parsed as to separate entries in an "argv" array for a command launch-- but now "-framework OpenGL" is being passed as  one argument.  This change makes sure `-framework` and `OpenGL` are two separate args.

### DIFF
--- a/kiva/agg/setup.py
+++ b/kiva/agg/setup.py
@@ -139,11 +139,9 @@ def configuration(parent_package='', top_path=None):
     if sys.platform == 'darwin':
         define_macros = [('__DARWIN__', None)]
         macros = [('__DARWIN__', None)]
-        extra_link_args = ['-framework', 'Carbon']
     else:
         define_macros = []
         macros = []
-        extra_link_args = []
 
     kiva_include_dirs = ['src'] + agg_include_dirs
     config.add_library('kiva_src',
@@ -184,12 +182,17 @@ def configuration(parent_package='', top_path=None):
         darwin_frameworks = ['Carbon', 'ApplicationServices', 'OpenGL']
     else:
         darwin_frameworks = ['ApplicationServices', 'OpenGL']    
+
+    darwin_extra_link_args = []
+    for framework in darwin_frameworks:
+        darwin_extra_link_args.extend(['-framework', framework])
+
     darwin_opengl_opts = dict(
             include_dirs = [
               '/System/Library/Frameworks/%s.framework/Versions/A/Headers' % x
               for x in darwin_frameworks],
             define_macros = [('__DARWIN__',None)],
-            extra_link_args = ['-framework %s' % x for x in darwin_frameworks]
+            extra_link_args = darwin_extra_link_args
             )
 
     build_info = {}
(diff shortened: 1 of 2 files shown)


kiva/quartz/setup.py
```diff
@@ -60,6 +63,10 @@ def generate_c_from_cython(extension, build_dir):
                                   ],
                         )
 
+    extra_link_args = []
+    for framework in frameworks:
+        extra_link_args.extend(['-framework', framework])
```
jwiggins: Why the extra step here? `extra_link_args` will have the same value it got above.

silverjam: Pretty sure there was a reason at some point :-| -- but looks like you're absolutely correct, I'll test out removing this bit and see what happens.

silverjam:
@jwiggins removed the redundant block and re-ran `python setup.py bdist_wheel --universal`, which seemed to work fine.

kiva/agg/setup.py
```diff
@@ -139,11 +139,9 @@ def get_ft2_sources(name_info, build_dir):
     if sys.platform == 'darwin':
         define_macros = [('__DARWIN__', None)]
         macros = [('__DARWIN__', None)]
-        extra_link_args = ['-framework', 'Carbon']
```
silverjam: Dropped adding `framework -Carbon` above since it appears to be unnecessary (compile and link works without it), otherwise the change in how we define the list results in different build error:

```
54 warnings generated.
g++ -bundle -undefined dynamic_lookup -L/Users/jason/miniconda3/envs/py37/lib -arch x86_64 -L/Users/jason/miniconda3/envs/py37/lib -arch x86_64 -arch x86_64 build/temp.macosx-10.7-x86_64-3.7/build/src.macosx-10.7-x86_64-3.7/kiva/agg/agg_wrap.o -Lbuild/temp.macosx-10.7-x86_64-3.7 -lkiva_src -lagg24_src -lfreetype2_src -o build/lib.macosx-10.7-x86_64-3.7/kiva/agg/_agg.cpython-37m-darwin.so -framework ApplicationServices -framework OpenGL Carbon
clang: error: no such file or directory: 'Carbon'
error: Command "g++ -bundle -undefined dynamic_lookup -L/Users/jason/miniconda3/envs/py37/lib -arch x86_64 -L/Users/jason/miniconda3/envs/py37/lib -arch x86_64 -arch x86_64 build/temp.macosx-10.7-x86_64-3.7/build/src.macosx-10.7-x86_64-3.7/kiva/agg/agg_wrap.o -Lbuild/temp.macosx-10.7-x86_64-3.7 -lkiva_src -lagg24_src -lfreetype2_src -o build/lib.macosx-10.7-x86_64-3.7/kiva/agg/_agg.cpython-37m-darwin.so -framework ApplicationServices -framework OpenGL Carbon" failed with exit status 1
```

silverjam:
Hi @jwiggins, anything else we should address/check? This PR should resolve #329.

jwiggins:
LGTM. Thanks for the contribution!

PR #336, opened 2019-04-05
Update Cython speed-ups to support Python 3.7

silverjam:
Current contents of `kiva/_cython_speedups.cpp` don't compile on Python 3.7, use Cython 0.29.6 to update the generated code.


jwiggins:
Thanks for the contribution @silverjam!

PR #338, opened 2019-04-17
Fix horizontal orientation support in SVGButton's Qt backend

tn-exafore:
The SVGButton has an orientation parameter but the visual layout in Qt backend doesn't take this into account but always uses a vertical layout.


corranwebster:
LGTM

Thanks for the contribution.

PR #340, opened 2019-05-02
Fix support for Qt5 QWheelEvents whose pixelDelta is zero

tn-exafore:
* pixelDelta is not provided on all platforms (see https://doc.qt.io/qt-5/qwheelevent.html#pixelDelta), so one needs to use angleDelta or scroll events won't work. For example, on Windows or X11. Perhaps pixelDelta is just reliably available on MacOS? There are probably several ways to do this, I think the one proposed here is one of the simplest. The scaling (or lack of it) between angle and pixel deltas hasn't been thought out, but I followed the existing code in this (see also the related qt4 if branch in basewindow).
* also fix the pixelDelta argument in QWheelEvent constructor calls in the related unit test: it's an integer QPoint, not a float
* add a unit test for a wheel event without pixelDelta

### DIFF
--- a/enable/qt4/base_window.py
+++ b/enable/qt4/base_window.py
@@ -430,6 +430,8 @@ class _Window(AbstractWindow):
                     mouse_wheel_delta = (0, delta)
             else:
                 delta = event.pixelDelta()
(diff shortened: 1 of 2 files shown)


jwiggins:
Looks good. Thanks for the fix!

PR #343, opened 2019-05-25
Fix code which triggered DeprecationWarnings

### LINKED ISSUE #339: DeprecationWarning triggered by downstream package tests
We have a project using chaco (and hence enable).  Out test suites trigger the following DeprecationWarning (only included here ones that are from `enable`):
```
/Users/kchoi/ETS/enable/kiva/agg/agg.py:1734: DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead
  img = fromstring(piltostring(pil_img),uint8)
/Users/kchoi/ETS/enable/enable/component.py:1094: DeprecationWarning: use "HasTraits.trait_set" instead
  self.set(bounds=[new_w, new_h], trait_change_notify=notify)
/Users/kchoi/ETS/enable/enable/component.py:1096: DeprecationWarning: use "HasTraits.trait_set" instead
  self.set(position=new_pos, trait_change_notify=notify)
........../Users/kchoi/ETS/enable/kiva/agg/agg.py:1734: DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead
  img = fromstring(piltostring(pil_img),uint8)
```

There may be more deprecation warnings not triggered by our test suite.

Platform: OSX
Python: 3.6
Tested with commit 77b23977ad5486328ec59f6bc14cbbdc8b53d7dd on master

jwiggins:
Fixes #339

### DIFF
--- a/enable/container.py
+++ b/enable/container.py
@@ -199,8 +199,9 @@ class Container(Component):
             # Update our position and the positions of all of our components,
             # but do it quietly
             for component in self._components:
-                component.set(position = [component.x-ll_x, component.y-ll_y],
-                              trait_change_notify = False)
+                component.trait_setq(
+                    position=[component.x-ll_x, component.y-ll_y]
+                )
 
             # Change our position (in our parent's coordinate frame) and
             # update our bounds
--- a/enable/component.py
+++ b/enable/component.py
@@ -1091,9 +1091,9 @@ class Component(CoordinateBox, Interactor):
                 new_pos = self.position[:]
                 new_pos[0] += (old_w - new_w) / 2.0
 
-        self.set(bounds=[new_w, new_h], trait_change_notify=notify)
+        self.trait_set(bounds=[new_w, new_h], trait_change_notify=notify)
         if new_pos:
-            self.set(position=new_pos, trait_change_notify=notify)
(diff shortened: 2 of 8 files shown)


jwiggins:
Note that `piltostring` is just a wrapper around [Image.tobytes](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.tobytes) (or `Image.tostring` when not using Pillow). All of the `fromstring` -> `frombuffer` conversions are passing the output of `piltostring`.

enable/container.py
```diff
@@ -199,8 +199,10 @@ def compact(self):
             # Update our position and the positions of all of our components,
             # but do it quietly
             for component in self._components:
-                component.set(position = [component.x-ll_x, component.y-ll_y],
-                              trait_change_notify = False)
+                component.trait_set(
+                    position=[component.x-ll_x, component.y-ll_y],
+                    trait_change_notify=False
```
jvkersch: How do you feel about replacing `.trait_set(..., trait_change_notify=False)` by `.trait_setq(...)`?

jwiggins: Yeah. That would be a bit better, I think.

PR #349, opened 2019-06-05
Update unittest import

rahulporuri:
import it from the standard library instead of importing it via `traits`.


rahulporuri:
I rebased this branch on top of master to fix the merge conflicts.

PR #367, opened 2019-09-24
Nose captures logs!

rahulporuri:
Nose was capturing logs when running the test suite, which is why we werent seeing the kiva font manager related logs/tracebacks on the enable test runs but we were seeing them in other package test runs!

(I checked the following packages and they make sure that nose doesnt capture logs - traits, traitsui, pyface. chaco doesn't use nose anymore.)


jwiggins:
Can you rebase this on to the latest master? (ie: post #366)

Issue #362, opened 2019-08-19
Avoid import time side-effect from font_manager


kitchoi:
Currently importing `kiva.fonttools.font_manager` has a side effect of loading the system font files.  This makes it difficult to configure logging in tests should the side effect emits log messages.

For a project that depends on enable (e.g. via chaco), running a test suite using `unittest discover` may trigger log messages because FontManager fails to parse one or many system fonts:
```
[00:03:41] Traceback (most recent call last):
[00:03:41]   File "C:\Users\appveyor\.edm\envs\py36-ci\lib\site-packages\kiva\fonttools\font_manager.py", line 1440, in <module>
[00:03:41]     fontManager = pickle_load(_fmcache)
[00:03:41]   File "C:\Users\appveyor\.edm\envs\py36-ci\lib\site-packages\kiva\fonttools\font_manager.py", line 1050, in pickle_load
[00:03:41]     fh = open(filename, 'rb')
[00:03:41] FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\appveyor\\AppData\\Roaming\\Enthought\\kiva\\fontList.cache'
[00:03:41] 
[00:03:41] During handling of the above exception, another exception occurred:
[00:03:41] 
[00:03:41] Traceback (most recent call last):
[00:03:41]   File "C:\Users\appveyor\.edm\envs\py36-ci\lib\site-packages\kiva\fonttools\font_manager.py", line 738, in createFontList
[00:03:41]     prop = ttfFontProperty(fpath, font)
[00:03:41]   File "C:\Users\appveyor\.edm\envs\py36-ci\lib\site-packages\kiva\fonttools\font_manager.py", line 525, in ttfFontProperty
[00:03:41]     name = props[(1, 0, 0, 1)].decode()
[00:03:41] KeyError: (1, 0, 0, 1)
[00:03:41] Could not covert font to FontEntry for file C:\windows\Fonts\LeelawUI.ttf
[00:03:41] Traceback (most recent call last):
[00:03:41]   File "C:\Users\appveyor\.edm\envs\py36-ci\lib\site-packages\kiva\fonttools\font_manager.py", line 1440, in <module>
[00:03:41]     fontManager = pickle_load(_fmcache)
[00:03:41]   File "C:\Users\appveyor\.edm\envs\py36-ci\lib\site-packages\kiva\fonttools\font_manager.py", line 1050, in pickle_load
[00:03:41]     fh = open(filename, 'rb')
[...]

PR #368, opened 2019-09-24
Defer imports from font_manager in font module

rahulporuri:
Deferring these imports means that a number of places where we need to import from `font` don't trigger the import time side-effect in the `font_manager` module.

This PR doesn't attempt to actually fix either of the issues #362 or #363 .


jwiggins:
What is this fixing, if I may ask?

jwiggins:
Which references #362 and #363, neither of which are fixed here. Right?

rahulporuri:
Thanks for the quick reviews John!

PR #370, opened 2019-10-03
Don't touch logging configuration

jwiggins:
`kiva.ps` was messing with the logging configuration in a pretty serious way. Then to add insult to injury, the logger instance is _never used_ anywhere in the module.


jwiggins:
Thanks!

PR #374, opened 2019-11-26
Add opacity to null-toolkit-defined rgba_color_trait

ZachPope531:
If ETSConfig.toolkit is not `wx` or `qt4`, the alpha channel of `rgba_color` is not available. This patch defaults the opacity to 1.0 in that case.
You can reproduce the bug by running 
```
from enable.trait_defs.rgba_color_trait import rgba_color
from traits.etsconfig.api import ETSConfig

ETSConfig.toolkit = ''
print(rgba_color(255)[3])  # Throws IndexError, should be alpha channel
```


corranwebster:
I'm curious about the use case: most of the time Enable/Chaco uses `enable.colors.ColorTrait`or one of its specializations such as `black_color_trait`.  The `RGBAColor` and related classes and values are only used in situations where you expect to have a UI.

rkern:
1. The code in `master` is simply buggy and does not implement what it promises.
2. `RGBAColor` is used elsewhere to represent RGBA colors in models, even when `enable` isn't used (because it's the one color trait that actually implements the alpha channel).
3. Even `enable` needs to run with the `null` TraitsUI/PyFace toolkit in order to make static PNGs, etc.

This particular bug was evoked because the code managed to import this module before importing TraitsUI/PyFace/etc. that actually picked a backend. It was worked around by setting `ETS_TOOLKIT` to enforce the desired toolkit on app start, which is good, but the code in `master` is still buggy.

PR #378, opened 2019-12-13
Pass file object into TTCollection and close when done

### LINKED ISSUE #377: ResourceWarning with TTCollection
I am seeing the following warnings on a project running on Appveyor:

```
...\lib\site-packages\kiva\fonttools\font_manager.py:724: ResourceWarning: unclosed file <_io.BufferedReader name='C:\\Windows\\Fonts\\YuGothR.ttc'>
  collection = TTCollection(six.text_type(fpath)
```

Python: 3.6.5
OS: Windows
Enable version: 4.8.1

JCorson:
As of fonttools 4.1, we can use `TTCollection` as a context manager directly, but we are currently using 3.29 with EDM. However, we can pass an open file object directly into `TTCollection` rather than the file path.

Also includes a tiny refactor to remove duplicated logging error message strings. 

Closes #377

### DIFF
--- a/kiva/fonttools/font_manager.py
+++ b/kiva/fonttools/font_manager.py
@@ -721,7 +721,8 @@ def createFontList(fontfiles, fontext='ttf'):
             _, ext = os.path.splitext(fpath)
             try:
                 if ext.lower() == ".ttc":
-                    with TTCollection(six.text_type(fpath)) as collection:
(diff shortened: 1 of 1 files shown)


corranwebster:
Ignoring Python 2 errors, since we're not supporting Python 2 going forward.

PR #380, opened 2019-12-30
Demo runner (a TraitsUI for enable demos)

### LINKED ISSUE #379: Add demo runner from TraitsUI
Add a demo runner script as used in TraitsUI and Chaco examples.  This should at least work with the `examples/enable` directory and possibly the `examples/svg` directory, and ideally also the `examples/kiva` directory should have the scripts be runnable (although it may not do much in the UI).

For the Kiva scripts the right thing might be to refactor the scripts to have a `draw(gc)` function (or something similar) and a `main` function which creates a `GraphicsContext` and saves it to file.  When run stand-alone, the `__main__` simply calls main, but when run from within `demo.py` it could create an Enable component which uses the `draw` function and that gets displayed in the GUI.  This might require subclassing or specializing parts of `traitsui.extras.demo` from TraitsUI to add functionality for Enable.

midhun-pm:
Fixes #379 
A few examples without a visual component don't give any user feedback when run, will fix them in a different PR.


shoeb-github:
Is it not possible to make `traitsui` an optional dependency using `extras_require=...` in `setup.py`?

rkern:
`enable` has always had a full dependency on `traitsui`.

examples/demo.py
```diff
@@ -0,0 +1,25 @@
...
+#  license included in enthought/LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+#  Author: David C. Morrill
+#  Date:   09/15/2005
+#
+# -------------------------------------------------------------------------
```
midhun-pm: Brought this from traitsui, should this be modified?

rahulporuri: It should. The current copyright header policy for Enthought OSS projects is - 

```
# (C) Copyright 20XX-20XX Enthought, Inc., Austin, TX
# All rights reserved.
```

examples/demo.py
```diff
@@ -0,0 +1,25 @@
...
+
+""" Run the Traits UI demo.
+"""
+
+from __future__ import absolute_import
+
+from traitsui.extras.demo import demo
```
rahulporuri: It looks like this demo machinery in traitsui has the window title "Traits UI Demos" hardcoded. That's definitely what we want/expect but i'm not sure what the best way ahead is - Do we duplicate all of the demo-related functionality from traitsui in enable? Do we simply make the window title an argument to the `demo` function in traitsui (and then use a new release in enable)

examples/enable/basic_draw.py
```diff
@@ -2,8 +2,10 @@
...
 """
-from enable.example_support import DemoFrame, demo_main
-from enable.api import Component, Container, Window
+from traits.api import HasTraits, Instance
+from traitsui.api import Item, View
+
+from enable.api import Component, ComponentEditor
 
 
 class Box(Component):
```
rahulporuri: The `Box` class has a `resizable` attribute set to an empty string. I'm assuming that means that the box is not resizable - can you make this explicit and set it the attr to `False`?

examples/enable/basic_move.py
```diff
@@ -3,10 +3,11 @@
 """
 from __future__ import print_function
```
rahulporuri: I think we're at a place where we can get rid of all of the `__future__` imports and other python 2/3 compatibility code in the examples. What do you think?

examples/enable/basic_draw.py
rahulporuri: The color of the box is defined using rgba values. it would be more readable is we used predefined color constants (which i think exist).

examples/enable/basic_move.py
```diff
@@ -3,10 +3,11 @@
...
-
 from traits.api import Float
-from enable.api import Component, Pointer, Window
+from traits.api import HasTraits, Instance
+from traitsui.api import Item, View
+
+from enable.api import Component, Pointer, ComponentEditor
 
 
 class Box(Component):
```
rahulporuri: same comments about the `resizable` trait as before and about the box colors as before (basic_draw).

examples/enable/canvas_demo.py
```diff
@@ -1,27 +1,28 @@
...
 
-    def _create_window(self):
+    traits_view = View(Item('canvas', editor=ComponentEditor(),
+                            show_label=False, width=200, height=200),
+                       resizable=True, title="Canvas Example")
 
+    def _canvas_default(self):
         canvas = Canvas(bgcolor="lightsteelblue", draw_axes=True)
-        from basic_move import Box
+        from .basic_move import Box
```
rahulporuri: This demo doesn't seem to be working and I suspect that's because of this import - `examples` is not a python package and neither are any of it's sub dirs.

examples/savage/buttons_on_canvas.py
```diff
@@ -3,13 +3,14 @@
 import os.path
```
rahulporuri: clicking on this example from the examples UI window consistently crashes the demo application on my machine (Mac OS + Python 3.6 + PyQt)

the traceback is 

```
% edm run -e enable-test-3.6-pyqt-pillow -- python examples/demo.py
Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt. You must
reimplement QApplication::notify() and catch all exceptions there.

libc++abi.dylib: terminating with uncaught exception of type kiva::$_0
```

rahulporuri:
@midhun-pm you can ignore my comments regarding specific demos. I understand that the intention of this PR was to get a working runner with a few demos. We can handle cleaning up the demo scripts in a later PR/issue.

rahulporuri:
@midhun-pm the CI is failing because I think it tries installing traits from the github source - which doesn't support python 2.7 anymore. I think we can stop testing on Python 2.7 at this point but I would prefer that change be made in a separate PR.

examples/demo.py
corranwebster: Make the window title customizable in the traitsui.extras.demo.

examples/enable/basic_draw.py
corranwebster: The `resizable` trait holds the _directions_ that the trait can be resized, expecting (IIRC) `h` for horizontal and `v` for vertical, `''` means neither direction is resizable.

rahulporuri: ahh. i shouldn't have made comments without looking at what the trait actually does. sorry for the noise. resolving conversation.

examples/enable/basic_draw.py
```diff
@@ -19,16 +21,18 @@ def _draw_mainlayer(self, gc, view_bounds=None, mode="default"):
             gc.fill_path()
 
 
-class MyFrame(DemoFrame):
+class Demo(HasTraits):
```
corranwebster: This, and other similar changes, breaks the ability for the `pyglet` backend to run stand-alone examples (there is no TraitsUI support for pyglet).  Rather than replace `DemoFrame` in all the examples, instead replace the `DemoFrame` implementation here https://github.com/enthought/enable/blob/master/enable/example_support.py#L69 with a TraitsUI class (where `demo_main` just calls `configure_traits`) and refactor so that `create_window` is replaced by an abstract method that creates the top-level `component` and the pyglet `demo_main` is responsible for creating the `Window`.

corranwebster: OTOH, I'm not sure how much we care about the pyglet backend.

examples/enable/basic_move.py
corranwebster: Should do it as part of a separate PR

corranwebster:
There's a lot of repeated code for grabbing images from temporary files and displaying it.  Perhaps that should be turned into common code in kiva or enable - it distracts from the main point of the example.

examples/enable/basic_draw.py
rkern: Let's leave the contents of the demos for another PR.

examples/enable/shapes/run.py
```diff
@@ -1,12 +1,12 @@
...
 # Local imports
 from box import Box
 from circle import Circle
 
+# Enthought library imports.
+from enable.api import Container, Window
+from enable.example_support import DemoFrame, demo_main
+
 
 class MyFrame(DemoFrame):
```
shoeb-github: This still uses `DemoFrame`. Can be fixed by refactor suggested by @corranwebster.

midhun-pm:
Fix is in PR #383

PR #383, opened 2020-01-07
Demo runner for enable

### LINKED ISSUE #379: Add demo runner from TraitsUI

midhun-pm:
Fix #379

### DIFF
--- a/examples/enable/basic_draw.py
+++ b/examples/enable/basic_draw.py
@@ -27,10 +27,6 @@ class Demo(DemoFrame):
         container.add(box)
         return container
 
-    def _create_window(self):
-        return Window(self, -1, component=self._create_component())
-
-
 
 if __name__ == "__main__":
     # Save demo so that it doesn't get garbage collected when run within
--- a/examples/enable/basic_draw.py
+++ b/examples/enable/basic_draw.py
@@ -19,16 +19,20 @@ class Box(Component):
             gc.fill_path()
 
 
-class MyFrame(DemoFrame):
+class Demo(DemoFrame):
 
-    def _create_window(self):
+    def _create_component(self):
         box = Box(bounds=[100.0, 100.0], position=[50.0, 50.0])
         container = Container(bounds=[500, 500])
         container.add(box)
-        return Window(self, -1, component=container)
+        return container
+
+    def _create_window(self):
+        return Window(self, -1, component=self._create_component())
+
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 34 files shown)


examples/enable/basic_draw.py
```diff
@@ -19,16 +19,20 @@ def _draw_mainlayer(self, gc, view_bounds=None, mode="default"):
...
 
-    def _create_window(self):
+    def _create_component(self):
         box = Box(bounds=[100.0, 100.0], position=[50.0, 50.0])
         container = Container(bounds=[500, 500])
         container.add(box)
-        return Window(self, -1, component=container)
+        return container
+
+    def _create_window(self):
```
corranwebster: This looks almost identical in every example: can it be moved to `DemoFrame`? In fact it may only be needed for `pyglet` version of `DemoFrame`.

corranwebster:
I'm getting segfaults on OS X/Python 3.6/pyqt5 as soon as I try and view any demo (the app comes up and shows the tree, but then immediately dies when I select a demo).

This is in an environment created with `python ci/edmtool.py install --runtime=3.6 --toolkit=pyqt5`

I will try pyqt4 to see if it is any different.

Do I need to be on a particular branch of TraitsUI for the demos to work?

corranwebster:
On pyqt4 things work, but I am getting an error `Cannot import Constraints Container` for most of the Kiva demos; for some (like the gradient demo) I am getting can't import `scipy`.

Also, I needed to manually install configobj.

The core problem is that there are optional dependencies for the demos, so we need to have some way to install those.

However that is probably something best dealt with in another PR.

examples/enable/basic_draw.py
midhun-pm: Yes, will do that.

PR #386, opened 2020-01-09
Add optional demo dependencies to setup.py

### LINKED ISSUE #385: Demos have optional dependencies
Some demos have dependencies on additional packages like `scipy` and possibly `kiwisolver`, as well as on `traitsui[demo]`; we should update the `setup.py` to have optional dependencies for demos in the same way we do for TraitsUI; and we should modify the CI scripts to allow those to be optionally installed.

shoeb-github:
closes #385

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -106,7 +106,9 @@ dependencies = {
     "pypdf2",
     "swig",
     # required by some demos
-    "scipy"
+    "chaco",
+    "mayavi",
+    "scipy",
 }
 
 extra_dependencies = {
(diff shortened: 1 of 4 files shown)


ci/edmtool.py
```diff
@@ -105,6 +105,10 @@
     "unittest2",
     "pypdf2",
     "swig",
+    # required by some demos
+    "chaco",
+    "mayavi",
+    "scipy",
```
shoeb-github: I didn't make this optional because the `install` command sets up the dev environment which probably should have all dependencies.

PR #387, opened 2020-02-07
Maint: add Container.components setter.

jonathanrocher:
In Chaco's implementation of the `BasePlotContainer`, a [warning](https://github.com/enthought/chaco/blob/master/chaco/base_plot_container.py#L58) states that getting or setting `plot_components` is deprecated, and components should be used instead. But enable's `Container` class doesn't have a setter for the `components` property. 

This PR corrects that.


jonathanrocher:
Thanks for approving the change @jvkersch . I added a unit test to exercise the setter. Further comments/reviews welcome.

jonathanrocher:
This has been sitting approved for a while, so I will merge soon unless there are more comments.

PR #401, opened 2020-03-26
Made the test examples run on windows and using wxpython4

pbrod:
Fixes #400 Made the test examples run on windows and using wx:
 * Replaced deprecated "wx.EVT_XXX(panel, handler)" with "panel.Bind(wx.EVT_XXX, handler)" in   rgba_color_editor.py
 * Replaced faulty "slider.SetTickFreq( 10, 1 )" with "slider.SetTickFreq(10)" in rgba_color_editor.py
 * Fixed the failing ToolkitEditorFactory.all_facenames method in kiva_font_editor.py
 * Replaced deprecated wx.PyPanel with wx.Panel
 * Insulated the xxxtrait.configure_traits() call inside a "if __name__ == '__main__'" block
 * Added the "demo = XXXtrait()" to each demo-script so it can be discovered and run by the traitsui.extra.demo.demo.
 * Changed "from enable.wx_backend import Window" to "from enable.api import Window"
and moved all imports to the top of the file.
 * Replaced call to wx.SystemSettings_GetMetric with wx.SystemSettings.GetMetric
 * Replaced call to deprecated wx.EVT_XXXX( control, self._on_XXXX) with control.Bind(wx.EVT_ERASE_BACKGROUND, self._on_XXXXX)
 * Fixed a bug in BaseWindow._create_control: If parent is None it is set to wx.Frame(None) in order to avoid crash in the unittests.
 * Replaced wx.WXK_PRIOR and wx.WXK_NEXT (which does not exist on wxpython4) with wx.WXK_PAGEUP and wx.WXK_PAGEDOWN
 * Added missing player_play.svg and player_pause.svg to examples/savage folder (LGPL license)

Fixes #399 replaced:
 * wx.GraphicsRenderer_GetDefaultRenderer with wx.GraphicsRenderer.GetDefaultRenderer
 * wx.GraphicsContext_Create with wx.GraphicsContext.Create

Fixes #396 Made sure rows and columns are integers by replacing "/" with "//" in kiva/agg/freetype2/src/tools/docmaker/tohtml.py

Fixes #395
 * Replaced dangerous default value "transforms=[]" with "transforms=None" in __init__ methods.

Fixes #393 Added self as first argument to AbstractGraphicsContext.draw_image method

Fixed failing examples:
 * examples/kiva/agg/benchmark.py (Float -> float, added missing import of  zeros)
 * examples/kiva/agg/polygon_hit_test.py (point_in_polygon -> points_in_polygon)
[...]

PR #403, opened 2020-03-29
WxPython 4 Fixes

### LINKED ISSUE #399: wx.GraphicsRenderer_GetDefaultRenderer does not exist in wxpython 4
https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/enable/savage/svg/backends/wx/renderer.py#L159

replace with wx.GraphicsRenderer.GetDefaultRenderer

jwiggins:
**NOTE:** This was pulled in from #401 to make review a bit easier.

Resolves #399

* Made `wx.VERSION` comparison more robust
* Made the test examples run on windows and using wx:
  * Replaced deprecated `wx.EVT_XXX(panel, handler)` with `panel.Bind(wx.EVT_XXX, handler)` in  `rgba_color_editor.py`
  * Replaced faulty `slider.SetTickFreq( 10, 1 )` with `slider.SetTickFreq(10)` in `rgba_color_editor.py`
  * Fixed the failing `ToolkitEditorFactory.all_facenames` method in `kiva_font_editor.py`
  * Replaced deprecated `wx.PyPanel` with `wx.Panel`
  * Changed `from enable.wx_backend import Window` to `from enable.api import Window` and moved all imports to the top of the file.
  * Replaced call to `wx.SystemSettings_GetMetric` with `wx.SystemSettings.GetMetric`
  * Replaced call to deprecated `wx.EVT_XXXX( control, self._on_XXXX)` with `control.Bind(wx.EVT_ERASE_BACKGROUND, self._on_XXXXX)`
  * Fixed a bug in `BaseWindow._create_control`:
    If `parent` is None it is set to `wx.Frame(None)` in order to avoid crash in the unit tests.
  * Replaced `wx.WXK_PRIOR` and `wx.WXK_NEXT` (which does not exist on wxpython4)
    with `wx.WXK_PAGEUP` and `wx.WXK_PAGEDOWN`

Fixes #399 replaced:
* `wx.GraphicsRenderer_GetDefaultRenderer` with `wx.GraphicsRenderer.GetDefaultRenderer`
* `wx.GraphicsContext_Create` with `wx.GraphicsContext.Create`

### DIFF
--- a/enable/trait_defs/ui/wx/rgba_color_editor.py
+++ b/enable/trait_defs/ui/wx/rgba_color_editor.py
@@ -16,7 +16,7 @@ the wxPython user interface toolkit.
 import wx
 
 from enable.colors import color_table
-
+from enable.api import Label, Window
 from traits.api import Bool
 from traits.trait_base import SequenceTypes
 
@@ -31,14 +31,11 @@ from traitsui.wx.helper import position_window
 #-------------------------------------------------------------------------------
 
 # Standard color samples:
-color_choices = ( 0, 51, 102, 153, 204, 255 )
-color_samples = [ None ] * 216
-i             = 0
-for r in color_choices:
-    for g in color_choices:
-        for b in color_choices:
-            color_samples[i] = wx.Colour( r, g, b )
-            i += 1
+COLOR_CHOICES = ( 0, 51, 102, 153, 204, 255 )
+COLOR_SAMPLES = tuple([wx.Colour(r, g, b)
+                       for r in COLOR_CHOICES
+                       for g in COLOR_CHOICES
+                       for b in COLOR_CHOICES])
 
 #-------------------------------------------------------------------------------
 #  'ToolkitEditorFactory' class:
@@ -164,8 +161,6 @@ class SimpleColorEditor ( Editor ):
         """ Finishes initializing the editor by creating the underlying toolkit
             widget.
         """
-        from enable.api import Label
-        from enable.wx_backend import Window
         window = Window( parent,
                     component = Label( '', border_size = 1, font = 'modern 9' ) )
         self._swatch = window.component
@@ -314,8 +309,6 @@ class ReadonlyColorEditor ( ReadonlyEditor ):
         """ Finishes initializing the editor by creating the underlying toolkit
             widget.
         """
-        from enable.api import Label
-        from enable.wx_backend import Window
         window = Window( parent,
                     component = Label( '', border_size = 1, font = 'modern 9' ) )
         self._swatch = window.component
@@ -381,12 +374,13 @@ def color_editor_for ( editor, parent, update_handler = None ):
     # Add all of the color choice buttons:
     sizer2 = wx.GridSizer( 0, 12, 0, 0 )
 
-    for i in range( len( color_samples ) ):
+    for color_sample in COLOR_SAMPLES:
         control = wx.Button( panel, -1, '', size = wx.Size( 18, 18 ) )
-        control.SetBackgroundColour( color_samples[i] )
+        control.SetBackgroundColour( color_sample )
         control.update_handler = update_handler
-        wx.EVT_BUTTON( panel, control.GetId(),
-                       swatch_editor.update_object_from_swatch )
+        panel.Bind(wx.EVT_BUTTON,
+                   swatch_editor.update_object_from_swatch,
+                   id=control.GetId())
         sizer2.Add( control )
         editor.set_tooltip( control )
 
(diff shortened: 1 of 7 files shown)


rahulporuri:
@jwiggins can you merge master into this branch?

kitchoi:
https://github.com/enthought/enable/pull/424 has refreshed the CI setup. CI builds since #386 was merged (Jan 2020) may not have been run against the local source.
Note there is also a new release of wxPython 4.1.1 recently (Nov 2020), which may or may not impact the changes in this PR.

enable/trait_defs/ui/wx/rgba_color_editor.py
```diff
@@ -16,7 +16,7 @@
 import wx
 
 from enable.colors import color_table
-
+from enable.api import Label, Window
```
kitchoi, line 19: This could result in import cycles waiting to happen. It may not happen by virtue of ordering (i.e. alphabetical) and importlib caching imported modules. I think we should just do `from enable.label import Label` and `from enable.window import Window` here.
(My understanding is that `api` module is for others using the library and for tests, as tests should mimic the environment used by the library user. The `api` module is not targeting enable itself.

enable/wx/base_window.py
```diff
@@ -215,37 +215,37 @@ def __init__(self, parent, wid=-1, pos=wx.DefaultPosition,
...
+            self.control.Bind(wx.EVT_RIGHT_UP, None)
+            self.control.Bind(wx.EVT_RIGHT_DCLICK, None)
+            self.control.Bind(wx.EVT_MOTION, None)
+            self.control.Bind(wx.EVT_ENTER_WINDOW, None)
+            self.control.Bind(wx.EVT_LEAVE_WINDOW, None)
+            self.control.Bind(wx.EVT_MOUSEWHEEL, None)
+            self.control.Bind(wx.EVT_KEY_DOWN, None)
+            self.control.Bind(wx.EVT_KEY_UP, None)
+            self.control.Bind(wx.EVT_CHAR, None)
+            self.control.Bind(wx.EVT_WINDOW_DESTROY, None)
```
kitchoi, line 248: Does binding to None mean the same as `Unbind`, e.g.?:
```
self.control.Unbind(wx.EVT_WINDOW_DESTROY)
```
`Bind` to `None` looks weird to me, just wanted to mention I am unsure what is right here.

enable/trait_defs/ui/wx/rgba_color_editor.py
jwiggins: Good catch

enable/wx/base_window.py
jwiggins: No idea. I have little to no experience writing Wx code

kiva/trait_defs/ui/wx/kiva_font_editor.py
```diff
@@ -107,20 +108,21 @@ def str_font ( self, font ):
...
         return '%s point %s%s%s%s' % (
                font.size, self.face_name( font ), style, weight, underline )
 
     #---------------------------------------------------------------------------
     #  Returns a list of all available font facenames:
     #---------------------------------------------------------------------------
-
-    def all_facenames ( self ):
+    @staticmethod
+    def all_facenames():
```
kitchoi, line 120: The switch to staticmethod looks unnecessary, it is called by the parent class on an instance.
The parent class has `all_facenames` as an instance method:
https://github.com/enthought/traitsui/blob/ab90190dded4c70b2ba39e9323b6ef424c926dd4/traitsui/wx/font_editor.py#L122

For consistency, this should probably be kept as an instance method even though its current implementation does not use `self.`

(For forward compatibility, it is probably better to keep it as instance method as it is easier to modify an instance method which has retained access to self than a static method which has almost as much access as a module level function.)

enable/wx/base_window.py
kitchoi: wxPython confirms this, so we are good:

> handler – A callable object to be invoked when the event is delivered to self. Pass None to disconnect an event handler.

(from https://wxpython.org/Phoenix/docs/html/wx.EvtHandler.html#wx.EvtHandler.Bind)

kitchoi (commented):
Done reviewing, LGTM, two minor suggestions.

enable/wx/base_window.py
jwiggins: Thanks for investigating

kiva/trait_defs/ui/wx/kiva_font_editor.py
jwiggins: I also removed the `@staticmethod` change on `font_face`

kitchoi (approved):
Thank you!

jwiggins:
Thanks for the review

Issue #415, opened 2020-09-25
Drop Python 2 Support


corranwebster:
The next major release of Enable should drop support for Python 2 and Python 3.5 and lower.

rahulporuri:
@aaronayres35 is this complete now?

aaronayres35:
> @aaronayres35 is this complete now?

I believe so (or at least very close).  I was planning to do one final sweep through to make sure there wasn't anything I missed.  Will do that this afternoon and close the issue after that if that works with you?  @rahulporuri

rahulporuri:
>  Will do that this afternoon and close the issue after that if that works with you?

SGTM.

PR #405, opened 2020-06-09
FIX: Make font style- and stretch- check case insensitive

achabotl:
This makes style and stretch case insensitive in
`kiva.fonttools.font_manager.ttfFontProperty`. The case sensitivity led
to font styles not being properly detected on (at least) macOS.

The checks on filename were already case insensitive.

I'm not sure how to test this. There's a test in `fonttools`, but it would required modifying the test `.ttf` file and I didn't know how to fix it. Help would be appreciated.

https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/kiva/fonttools/tests/test_font_manager.py#L53

Here's a small test sample and an example that it works
```python
from kiva.fonttools import Font
from kiva.image import GraphicsContext
from kiva.constants import ITALIC, BOLD

gc = GraphicsContext((200, 200))

f = Font('', size=24)
print("Top: ", f.findfont())
gc.set_font(f)
gc.set_text_position(30, 150)
gc.show_text("Hello World")

f = Font('Georgia', size=24)
print('2nd', f.findfont())
gc.set_font(f)
gc.set_text_position(30, 110)
gc.show_text("Hello World")

f = Font('Georgia', size=24, style=ITALIC)
print('3rd', f.findfont())
gc.set_font(f)
gc.set_text_position(30, 70)
gc.show_text("Hello World")

f = Font('Georgia', size=24, style=BOLD)
print('4th', f.findfont())
gc.set_font(f)
gc.set_text_position(30, 30)
gc.show_text("Hello World")

gc.save('text.png')
```

text

Fixes: #404
Also addresses part of #391.

PR #424, opened 2020-10-14
Make CI green again and remove Python 3.5 from CI

### LINKED ISSUE #410: Use the latest EDM version (3.0.1)
A new major version of EDM is available and it would be good to update to the latest version - https://assets.enthought.com/downloads/edm/. This should be as simple as updating the environment variables in travis and appveyor config files.

https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/.travis.yml#L23
https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/appveyor.yml#L8

### LINKED ISSUE #423: Test failure on asserting identity between trait lists
It seems that the new test added in #387 wasn't run on CI when it was last build.

I am experiencing issue with the `edmtool.py` install command: It does not seem to install the new source. When I do manage to install the most current source and I run the tests, I get the following test failure:
```
FAIL: test_get_set_components (enable.tests.container_test_case.ContainerTestCase)
  File "/Users/kchoi/.edm/envs/enable-test-3.6-pyqt5-pillow/lib/python3.6/site-packages/enable/tests/container_test_case.py", line 42, in test_get_set_components
    self.assertIs(container.components, new_list)
AssertionError: [<enable.component.Component object at 0x7fa3fe5198e0>, <enable.component.Component object at 0x7fa3fe519990>] is not [<enable.component.Component object at 0x7fa3fe5198e0>, <enable.component.Component object at 0x7fa3fe519990>]
```
However the development environment has been installing master of traits, traitsui, pyface and apptools, so I am not sure if this was a regression from upstream projects. But reading the test, I suspect the test should have used `assertEqual`, because traits List always make a new instance of TraitListObject. I believe this is the relevant trait:
https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/enable/container.py#L81-L82

kitchoi:
This PR's goal is to make CI green again.

Closes #423
Closes #410

I tried my best to limit the scope of this PR, but it was difficult to achieve 😂 .

In summary:
- EDM version is upgraded to use RH7 repository.
- The test failure in #423 is fixed. The test was failing in the first place but the test was not run on CI before.
- Revert half of #386 which caused CI to be run against a released egg of enable instead of local source
- Fix test errors complaining about GuiTestAssistant not being implemented on null toolkit (see this [comment](https://github.com/enthought/enable/pull/405#issuecomment-642715980)).
- Pin dependencies on traits, traitsui, pyface, apptools to released versions.
- Install traits, traitsui, pyface and apptools from EDM, remove Python 3.5 from the CI matrix. (<-- this is a bit opportunistic)

Details:
- First commit makes sure the current source is exercised by CI and we get the same failure in #423
- The subsequent commit tried to resolve missing link for liblzma from libtiff. The issue affects the most recent libtiff egg for RH6 repository. Upgrading EDM in order to use the RH7 repository resolves this issue.
- Avoid #425 by pinning released versions of traits, traitsui, pyface and apptools. Further restrict to TraitsUI<7.1.0. I think the issue should be fixed separately
- Fix the error to do with GuiTestAssistant not being available on null toolkit. It came up in the build for #405 as well. Not sure why it only started occurring then.
- Revert half of #386 which installs dependencies for running demo examples (it brings in chaco, mayavi etc). Those dependencies bring in the released version of enable, and since then enable CI has not been testing the local source. We don't really need these dependencies for running tests nor for developing enable. It seems reasonable not to install it by default.

On this:

> - Install traits, traitsui, pyface and apptools from EDM, remove Python 3.5 from the CI matrix.

This is because many recent releases have not been built for EDM + Python 3.5 any more. Python 3.5 has already passed end-of-life (see #415 too). This step I can revert.

### DIFF
--- a/enable/tests/tools/hover_tool_test_case.py
+++ b/enable/tests/tools/hover_tool_test_case.py
@@ -23,7 +23,7 @@ if GuiTestAssistant.__name__ == "Unimplemented":
 
     # ensure null toolkit has an inheritable GuiTestAssistant
     # Note that without this definition, the test case fails as soon as it
-    # is instantiated, before it can be s
+    # is instantiated, before it can be skipped.
     class GuiTestAssistant(object):
         pass
 
--- a/enable/tests/tools/hover_tool_test_case.py
+++ b/enable/tests/tools/hover_tool_test_case.py
@@ -19,6 +19,13 @@ from traitsui.tests._tools import skip_if_null
 
 
 GuiTestAssistant = toolkit_object('util.gui_test_assistant:GuiTestAssistant')
+if GuiTestAssistant.__name__ == "Unimplemented":
+
+    # ensure null toolkit has an inheritable GuiTestAssistant
+    # Note that without this definition, the test case fails as soon as it
+    # is instantiated, before it can be s
+    class GuiTestAssistant(object):
+        pass
 
 
 LOWER_BOUND = 50
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,8 @@ env:
 
 matrix:
   include:
-  - env: RUNTIME=3.5 TOOLKITS="null pyqt pyqt5" PILLOW='pillow'
   - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5" PILLOW='pillow'
   - env: RUNTIME=3.6 TOOLKITS="wx" PILLOW='pillow'
-  - env: RUNTIME=3.5 TOOLKITS=null PILLOW='pillow<3.0.0'
   - env: RUNTIME=3.6 TOOLKITS=null PILLOW='pillow<3.0.0'
   allow_failures:
   - env: RUNTIME=3.6 TOOLKITS="wx" PILLOW='pillow'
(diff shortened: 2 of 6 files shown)


enable/tests/tools/hover_tool_test_case.py
```diff
@@ -19,6 +19,13 @@
 
 
 GuiTestAssistant = toolkit_object('util.gui_test_assistant:GuiTestAssistant')
+if GuiTestAssistant.__name__ == "Unimplemented":
+
+    # ensure null toolkit has an inheritable GuiTestAssistant
+    # Note that without this definition, the test case fails as soon as it
+    # is instantiated, before it can be s
```
rahulporuri, line 26:
```suggestion
    # is instantiated, before it can be skipped.
```

rahulporuri (approved):
LGTM. Haven't tested the changes to etstool locally though.

rahulporuri:
alsi fixes #410

jwiggins (approved):
LGTM

Issue #425, opened 2020-10-21
Test failure when run with TraitsUI 7.1.0


kitchoi:
Running the test suite with TraitsUI 7.1.0 results in the following failure (and many more like that):
```
======================================================================
ERROR: Failure: ImportError (cannot import name 'skip_if_null')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/home/travis/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/nose/loader.py", line 419, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "/home/travis/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 684, in _load
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/travis/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/enable/tests/test_component_editor.py", line 11, in <module>
    from traitsui.tests._tools import get_dialog_size, skip_if_null
ImportError: cannot import name 'skip_if_null'
```

While this could have been caught with regular cron job (none yet), the fact that enable depends on a private module of an external package and for something so trivial might be itself be worth fixing: I'd propose reimplementing these trivial functions within enable, e.g. `skip_if_null` can be replaced with just `unittest.skipIf(ETSConfig.toolkit == "null")` (assuming the toolkit selection has occurred prior to its evaluation).

PR #428, opened 2020-10-26
Add source click option used on cron jobs to install pyface, traitsui, and traits from source

### LINKED ISSUE #407: Setup cron job to test against dependencies from source
Travis (and possibly appveyor too) allow the creation of cron jobs. Some of the ETS packages now have cron jobs setup to test against dependencies which are installed from source - instead of testing against the released versions.

See examples : 

- https://github.com/enthought/envisage/pull/162
- https://github.com/enthought/traitsui/pull/914
- https://github.com/enthought/pyface/pull/549

aaronayres35:
fixes #407 

This PR makes it so that cron jobs install `pyface`, `traitsui`, and `traits` from source by adding a `--source` click option used in the install function in `edmtool.py`.

This follows what is done in other ets projects (e.g https://github.com/enthought/envisage/blob/8245931f7e3dfcb6513bfb7dc2ae9b10ca6f89a1/etstool.py#L255-L273)

Note that based on the comment
```
# Avoiding traitsui 7.1.0 for now due to private imports
# in test suite (see enthought/enable#425)
"traitsui==7.0.1-1",
```
and the issue #425 I expect that CI will currently fail for these cron jobs, but that is what we want so that these kinds of issues can be detected early.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ env:
   global:
     - INSTALL_EDM_VERSION=3.0.1
       PYTHONUNBUFFERED="1"
+    - TRAVIS_EVENT_TYPE="cron"
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ env:
   global:
     - INSTALL_EDM_VERSION=3.0.1
       PYTHONUNBUFFERED="1"
-    - TRAVIS_EVENT_TYPE="cron"
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,13 @@ before_install:
   - export PATH="${HOME}/edm/bin:/usr/lib/ccache:${PATH}"
   - edm install -y wheel click coverage
 install:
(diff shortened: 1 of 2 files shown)


aaronayres35:
Also, I am not sure how to manually trigger a cron job to ensure this is working correctly?  The code is directly pulled from envisage, but it would still be good to test with `${TRAVIS_EVENT_TYPE} == 'cron'` to make sure everything works as expected

rahulporuri (approved):
LGTM. I tested this locally and it works as expected.

In order to test this on CI, you should simply be able to set the environment variable `TRAVIS_EVENT_TYPE='cron'` in the env var section of the travis config file.

```
env:
    global:
        ...
```

aaronayres35:
CI fails on the cron job due to the errors described in #425 

Also, there is an allowed failure on wx, but it fails with:
```
RuntimeError: Python 3.6, toolkit wx, and pillow pillow not supported by test environments
```
because "wx" is not listed here:
```
supported_combinations = {
    '3.6': {'pyqt', 'pyqt5', 'null'},
}
```

In any case, the cron job seems to work as expected.  I will revert the last commit and merge when CI finishes again.

PR #430, opened 2020-10-26
make python 3.6 defualt runtime for click options instead of 3.5

### LINKED ISSUE #429: default runtime in edmtool.py is 3.5
Currently, the default runtime is python 3.5, see:
https://github.com/enthought/enable/blob/375d52e850295809829df2e5281c10866c745949/ci/edmtool.py#L149

but 3.5 is not even available as a supported combination anymore:
https://github.com/enthought/enable/blob/375d52e850295809829df2e5281c10866c745949/ci/edmtool.py#L91-L93

So trying to run `python ci/edmtool.py install` fails.  Eventually we will drop python <3.6, but at the very least for now the default should be changed to 3.6 as that is what we list as a supported combination already

aaronayres35:
fixes #429 

This PR simply bumps the runtime option default from 3.5 to 3.6 as 3.6 is what is listed in `supported_combinations`

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -54,8 +54,8 @@ using::
 
     python edmtool.py test_all
 
-Currently supported runtime values are ``2.7`` and ``3.5``, and currently
-supported toolkits are ``null``, ``pyqt``, and ``wx``.  Not all
+Currently supported runtime value is `3.6``, and currently
+supported toolkits are ``null``, ``pyqt``, and ``pyqt5``.  Not all
 combinations of toolkits and runtimes will work, but the tasks will fail with
 a clear error if that is the case.
 
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -146,7 +146,7 @@ def cli():
 
 
 @cli.command()
-@click.option('--runtime', default='3.5')
+@click.option('--runtime', default='3.6')
 @click.option('--toolkit', default='null')
 @click.option('--pillow', default='pillow')
 @click.option('--environment', default=None)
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM. Can you also update the docstring at the top of the module? The docstring still refers to `2.7` and `3.5` as the supported runtimes.

aaronayres35:
> LGTM. Can you also update the docstring at the top of the module? The docstring still refers to `2.7` and `3.5` as the supported runtimes.

I also change the docstring to say "currently supported toolkits are ``null``, ``pyqt``, and ``pyqt5``" instead of "currently supported toolkits are ``null``, ``pyqt``, and ``wx``" and `wx` is also not a part of `supported_combinations` atm.

aaronayres35:
> > LGTM. Can you also update the docstring at the top of the module? The docstring still refers to `2.7` and `3.5` as the supported runtimes.
> 
> I also change the docstring to say "currently supported toolkits are `null`, `pyqt`, and `pyqt5`" instead of "currently supported toolkits are `null`, `pyqt`, and `wx`" and `wx` is also not a part of `supported_combinations` atm.

seems wx support was lost here: #384 .  Should we open an issue to re-add wx support/ re-add it to CI?

rahulporuri:
> seems wx support was lost here: #384 . Should we open an issue to re-add wx support/ re-add it to CI?

we definitely should. I think there is a PR open that tries fixing the wx support.

aaronayres35:
> > seems wx support was lost here: #384 . Should we open an issue to re-add wx support/ re-add it to CI?
> 
> we definitely should. I think there is a PR open that tries fixing the wx support.

ah my mistake, it is still run on CI just as an allowed failure.  We still would want to fix that/ add it back to the supported_combinations though

Issue #307, opened 2018-06-19
wx.WXK_NEXT and wx.WXK_PRIOR not defined in wxpython 4.0.1


will133:
In enable/wx/constants.py it seems like the wx.WXK_NEXT and wx.WXK_PRIOR is still used.

corranwebster:
See the comment in your TraitsUI issue.  We are a long way from WxPython 4 support in Enable.

kitchoi:
This is now fixed by #403
Closing.

PR #432, opened 2020-10-26
re-adding wx as a supported combination in edmtool 

aaronayres35:
This PR simply re-adds 'wx' to `supported_combinations` in edmtool.py.  Currently we still have a CI for wx and it is an allowed failure.  However, because wx is not in this list of supported_combinations, the CI just fails immediately with 
```
RuntimeError: Python 3.6, toolkit wx, and pillow pillow not supported by test environments
Command exited with code 1
```
without actually running any tests.  Even if they fail currently it is good for the tests to be exercised with CI to see if new failures are ever introduced.

Also, this PR lifts the installation pattern for wxPython 4 from pyface as it is not yet available in EDM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ addons:
     - libfreetype6-dev
     - libcairo2-dev
     - libglu1-mesa-dev
+    - libsdl2-2.0-0
 
 env:
   global:
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -89,7 +89,7 @@ from contextlib import contextmanager
 import click
 
 supported_combinations = {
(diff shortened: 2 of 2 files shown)


kitchoi:
Thanks @aaronayres35 

Looks like we will need to add `libsdl2-2.0-0` to `.travis.yml`. Some tests are failing with errors like this:
```
ImportError: libSDL2-2.0.so.0: cannot open shared object file: No such file or directory
```
(If we should keep this allow-to-fail job, I hope it can at least fail with something more meaningful to be worth setting it up.)

kitchoi:
Thank you! We can now see https://github.com/enthought/enable/issues/307 in CI.

kitchoi (approved):
LGTM

PR #437, opened 2020-10-28
Use PyQt5 from EDM instead of PyPI 

### LINKED ISSUE #431: Use PyQt5 from EDM instead of PyPI
At the moment, we seem to be installing a specific version of PyQt5 from PyPI - instead of installing it using EDM. Now that PyQt5 is available (has been available) via EDM, we should use EDM instead of PyPI.

aaronayres35:
fixes #431 

This PR simply installs pyqt5 from edm instead of pip.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,18 @@ addons:
     - libpng-dev
     - libfreetype6-dev
     - libcairo2-dev
+    # Qt dependencies
     - libglu1-mesa-dev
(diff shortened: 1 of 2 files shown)


rahulporuri:
Looks like the failure is because of missing apt packages. Can you take a look at the apt packages installed by traitsui/pyface and add those here?

rahulporuri (approved):
LGTM

PR #444, opened 2020-11-11
Add dclick option to test assistant

### LINKED ISSUE #438: Double click testing not possible with EnableTestAssistant
I'm using EnableTestAssistant to test some tools, and I would like to test some double-click behavior. However, from I can tell, it seems that the `mouse_down` function is hard-coded to dispatch `_down`, and there does not appear to be a separate function for testing double-click's specifically.

It seems like a straight-forward addition to add a `dclick` keyword argument to the `mouse_down` function with a default value of `False`, and when `True`, the respective event is fired with the `_dclick` suffix instead of `_down`.

notmatthancock:
- Adds `dclick` option to `EnableTestAssistant.mouse_down`
- Fixes import failure in associated test module
- Adds tests for `EnableTestAssistant.mouse_down`

Closes #438

### DIFF
--- a/enable/testing.py
+++ b/enable/testing.py
@@ -169,8 +169,7 @@ class EnableTestAssistant(KivaTestAssistant):
         return event
 
     def mouse_down(self, interactor, x, y, button='left', window=None,
-                   alt_down=False, control_down=False, shift_down=False,
-                   dclick=False):
+                   alt_down=False, control_down=False, shift_down=False):
         """ Send a mouse button down event to the interactor.
 
         Parameters
@@ -203,8 +202,57 @@ class EnableTestAssistant(KivaTestAssistant):
             The button is pressed while `shift` is down. Default value is
             False.
 
-        dclick : boolean, optional
-            If True, 'dclick' event is dispatched instead of single mouse down
+        Returns
+        -------
+        event : MouseEvent
+            The event instance after it has be processed by the `interactor`.
+
+        """
+        window = self.create_mock_window() if window is None else window
+        event_attributes = {'x': x, 'y': y,
+                            'alt_down': alt_down,
+                            'control_down': control_down,
+                            'shift_down': shift_down,
+                            '{0}_down'.format(button): True,
+                            'window': window}
+        event = self.create_mouse_event(**event_attributes)
+        self._mouse_event_dispatch(interactor, event,
+                                   '{0}_down'.format(button))
+        return event
+
+    def mouse_dclick(self, interactor, x, y, button='left', window=None,
+                     alt_down=False, control_down=False, shift_down=False):
+        """ Send a mouse double-click event to the interactor.
+
+        Parameters
+        ----------
+        interactor : Interactor
+            The interactor (or subclass) where to dispatch the event.
+
+        x : float
+            The x coordinates of the mouse position
+
+        y : float
+            The y coordinates of the mouse position
+
+        button : {'left', 'right'}, optional
+            The mouse button for which to simulate a press (down) action.
+
+        window : AbstractWindow, optional
+            The enable AbstractWindow to associate with the event. Default
+            is to create a mock class instance. If the window has a mouse
+            owner then that interactor is used.
+
+        alt_down : boolean, optional
+            The button is pressed while `alt` is down. Default value is False.
+
+        control_down : boolean, optional
+            The button is pressed while `control` is down. Default value is
+            False.
+
+        shift_down : boolean, optional
+            The button is pressed while `shift` is down. Default value is
+            False.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


enable/tests/test_assistant_test_case.py
```diff
@@ -3,8 +3,9 @@
 except ImportError:
     import mock
 import nose
+from unittest import skipIf
 
-from traitsui.tests._tools import skip_if_null
+from traitsui.tests._tools import is_null
```
kitchoi, line 8: Currently we have pinned CI to TraitsUI 7.0.1 when https://github.com/enthought/enable/issues/425 has yet to be fixed. Importing private module should be avoided as much as possible, especially for something this trivial, it is not worth the dependency. Could you please use `unittest.skipIf(ETSConfig.toolkit == "null")`?

kitchoi: Or revert back to the original for now as this is orthogonal?

notmatthancock: done

kitchoi: Thank you! Sorry for the inconvenience there.

enable/tests/test_assistant_test_case.py
```diff
@@ -23,7 +24,24 @@ def test_mouse_move():
...
+    component.normal_left_down = mock.Mock()
+    test_assistant.mouse_down(component, x=0, y=0)
+    component.normal_left_down.assert_called_once()
+
+
+def test_mouse_dclick():
+    test_assistant = EnableTestAssistant()
+    component = Component(bounds=[100, 200])
+    component.normal_left_dclick = mock.Mock()
+    test_assistant.mouse_down(component, x=0, y=0, dclick=True)
```
kitchoi, line 40: Instead of adding a flag for `mouse_down`, what do you think about having a separate method altogether?

Part of that is because Mouse Down != Mouse Click. Mouse Click includes both pressing the mouse down and releasing it at the same place. With mouse down and mouse release separated, one can, for example, test pressing the mouse somewhere, hold it, and then release it elsewhere. This would be what happens in a drag-and-drop action.
Double clicking a mouse key is a very different command: It involves system settings e.g. how quick two successive clicks should be considered as one double click event.

notmatthancock: Yea, that makes sense.

enable/testing.py
```diff
@@ -220,6 +220,58 @@ def mouse_down(self, interactor, x, y, button='left', window=None,
...
+        event : MouseEvent
+            The event instance after it has be processed by the `interactor`.
+
+        """
+        window = self.create_mock_window() if window is None else window
+        event_attributes = {'x': x, 'y': y,
+                            'alt_down': alt_down,
+                            'control_down': control_down,
+                            'shift_down': shift_down,
+                            '{0}_down'.format(button): True,
```
kitchoi, line 268: It took me a while to understand if this "left_down" or "right_down" was necessary... I think it could potentially be handled by the window to change focus, so it makes sense to keep.

kitchoi (commented):
LGTM.  Thank you for the addition!
(I am not an enable expert so it might be good to have a second pair of eyes on this.)

enable/testing.py
corranwebster: Whether or not it necessary, what matters is whether the generated event has the same features of the events which are generated by the backends, eg. https://github.com/enthought/enable/blob/854778dc8013ae79dbd35057ef191583105f47c7/enable/qt4/base_window.py#L456-L458

This doesn't allow generation of chorded double-click events, but that's probably for the best 😄

enable/tests/test_assistant_test_case.py
```diff
@@ -23,7 +24,24 @@ def test_mouse_move():
...
+
+def test_mouse_dclick():
+    test_assistant = EnableTestAssistant()
+    component = Component(bounds=[100, 200])
+    component.normal_left_dclick = mock.Mock()
+    test_assistant.mouse_dclick(component, x=0, y=0)
+    component.normal_left_dclick.assert_called_once()
+
+
+@skipIf(ETSConfig.toolkit == "null", "Skipping null tookit")
```
corranwebster, line 44: Not happy about this change.

corranwebster (approved):
Approved.

But can we please have a _public_ base-level testing module which has standard implementations for toolkit skip checks.  I am unhappy about the amount of code replication that occurs when we don't have that.

kitchoi:
> But can we please have a public base-level testing module which has standard implementations for toolkit skip checks. I am unhappy about the amount of code replication that occurs when we don't have that.

I'd propose handling this separately - it is orthogonal to this PR.

PR #452, opened 2020-11-29
Expand drag tool comments

### LINKED ISSUE #451: Interplay between DragTool flags not clear
The DragTool has a boolean flag called `end_drag_on_leave` that is True by default. However, the behavior of ending when the drag leaves is only observed when `capture_mouse` is False. This is because mouse leave events are generated by a mouse move event traveling through AbstractWindow and being transformed into mouse leave event; however, if `capture_mouse` is True, then events are dispatched back to the DragTool instead of passing through AbstractWindow. Therefore, `end_drag_on_leave` never works as expected so long as `capture_mouse` is True. This behavior is not clear from the documentation.

notmatthancock:
- Adds additional explanatory comments to DragTool flags
- Changes default value of `end_drag_on_leave` to False because the default value of True doesn't produce the described behavior

Closes #451

### DIFF
--- a/enable/tools/drag_tool.py
+++ b/enable/tools/drag_tool.py
@@ -17,7 +17,9 @@ class DragTool(BaseTool):
     drag_button = Enum("left", "right")
 
     # End the drag operation if the mouse leaves the associated component?
-    end_drag_on_leave = Bool(True)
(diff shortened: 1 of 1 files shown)


jwiggins (approved):
LGTM

PR #453, opened 2020-11-30
Remove EOF markers in python source files

rahulporuri:
This PR removes `EOF` markers in Python source files.


jwiggins (approved):
Such a strange convention...

PR #454, opened 2020-11-30
Make improvements to edmtool

rahulporuri:
This PR makes the following improvements to edmtool

- ensure that `-h` works along with `--help`
- add a new `shell` command

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -153,9 +153,9 @@ github_url_fmt = "git+http://github.com/enthought/{0}.git#egg={0}"
 
 
 # Ensure that "-h" is supported for getting help.
-CONTEXT_SETTINGS = dict(
-    help_option_names=["-h", "--help"],
-)
+CONTEXT_SETTINGS = {
+    "help_option_names": ["-h", "--help"],
+}
 
 
 @click.group(context_settings=CONTEXT_SETTINGS)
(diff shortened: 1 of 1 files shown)


ci/edmtool.py
```diff
@@ -152,7 +152,13 @@
 github_url_fmt = "git+http://github.com/enthought/{0}.git#egg={0}"
 
 
-@click.group()
+# Ensure that "-h" is supported for getting help.
+CONTEXT_SETTINGS = dict(
```
jwiggins, line 156: Nit: No reason why this shouldn't be a dictionary literal

ci/edmtool.py
```diff
@@ -234,6 +240,23 @@ def install(runtime, toolkit, pillow, environment, source):
...
+@click.option('--pillow', default='pillow')
+@click.option('--environment', default=None)
+def shell(runtime, toolkit, pillow, environment):
+    """ Create a shell into the EDM development environment
+    (aka 'activate' it).
+
+    """
+    parameters = get_parameters(runtime, toolkit, pillow, environment)
+    commands = [
+        "edm shell -e {environment}",
```
jwiggins, line 255: Not really your fault, but it seems a little bit smelly to split on whitespace when turning these strings into command lines. There are cases where that will fail.

jwiggins (approved):

PR #392, opened 2020-03-22
Split out the GL backend

jwiggins:
This PR aims to separate the Kiva OpenGL backend completely from the agg backend. This has several benefits:
* OpenGL can be avoided on headless installations with no GL libraries installed
* macOS builds on Catalina and later can skip OpenGL without losing agg
* The agg backend becomes a little bit simpler and (possibly) easier to replace with Celiagg

To that end, I've copied all of the GL-specific code from `kiva.agg` into a new `kiva.gl` subpackage. Included in the copy is a very small subset of agg 2.4 which is needed for affine transform, color, and path representations.

This will begin life as a draft PR, because there are a couple of important steps remaining:
* [x] Removal of the OpenGL code from the `kiva.agg` package
* [x] Testing!

PR #455, opened 2020-11-30
Remove some unnecessary empty return statements

rahulporuri:
This PR removes a number of unnecessary empty return statements from the codebase - from `enable`, from `kiva`, from the test suite and from the examples.

Apologies to the reviewer because this is quite an ugly PR to review. I checked each `return` statement to ensure that the empty statement is not part of conditional logic/early `return` from a function/method. Only empty `return`s at the end of functions/methods are removed.

### DIFF
--- a/enable/viewable.py
+++ b/enable/viewable.py
@@ -16,7 +16,6 @@ class Viewable(HasTraits):
     def request_redraw(self):
         # This overrides the default Component request_redraw by asking
         # all of the views to redraw themselves.
-        return
 
     def draw(self, gc, view_bounds=None, mode="default"):
         if len(self.viewports) > 0:
@@ -24,4 +23,3 @@ class Viewable(HasTraits):
                 view.draw(gc, view_bounds, mode)
         else:
             super(Viewable, self).draw(gc, view_bounds, mode)
-        return
--- a/enable/viewable.py
+++ b/enable/viewable.py
@@ -16,6 +16,7 @@ class Viewable(HasTraits):
     def request_redraw(self):
         # This overrides the default Component request_redraw by asking
         # all of the views to redraw themselves.
+        return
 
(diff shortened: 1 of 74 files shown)


enable/viewable.py
```diff
@@ -16,12 +16,10 @@ class Viewable(HasTraits):
     def request_redraw(self):
         # This overrides the default Component request_redraw by asking
         # all of the views to redraw themselves.
-        return
```
jwiggins, line 19: This causes an `IndentationError`

jwiggins (commented):
It seems like there should be a tool that can do this for us...

enable/viewable.py
```diff
@@ -16,6 +16,7 @@ class Viewable(HasTraits):
     def request_redraw(self):
         # This overrides the default Component request_redraw by asking
         # all of the views to redraw themselves.
+        return
```
jwiggins: You can also just change the comment to a docstring so that the `return` can still be removed.

kiva/gl.py
```diff
@@ -249,7 +249,6 @@ def __touch__(self, key):
                     self.__order__ = self.__order__[1:] + [key]
                 else:
                     self.__order__.append(key)
-        return
```
kitchoi, line 252: No action. Just an observation that this will conflict with #392
(There are probably plenty of opportunity for merge conflicts anyway.)

kitchoi (approved):
LGTM. I reviewed the code surrounding the removal and confirmed (the best I could) that they are all at the end of a function  (not early termination, and not function whose API explicitly requires a None to be returned)

kitchoi:
Since this PR was opened, a number of PRs have gone in, including one that updates the CI configuration to require tests to pass on wx. As a sanity check, it would be good to merge master in here to get another CI run before we merge.

PR #456, opened 2020-12-01
Replace use of tempfile.mktemp with tempfile.NamedTemporaryFile 

### LINKED ISSUE #398: Use of tempfile.mktemp is deprecated and poses a security risk
Replaced the unsafe call
https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/examples/kiva/compiled_path.py#L36

with

```python
with tempfile.NamedTemporaryFile(suffix='.jpg', delete=False) as fid:
    file_path = fid.name
```

aaronayres35:
fixes #398 

This PR replaces all uses of `tempfile.mktemp` as it is deprecated and unsafe.

### DIFF
--- a/examples/kiva/compiled_path.py
+++ b/examples/kiva/compiled_path.py
@@ -33,20 +33,17 @@ def compiled_path():
     gc.set_stroke_color((0, 0, 1, 1))
     gc.draw_path_at_points(locs, path, STROKE)
 
-    with tempfile.NamedTemporaryFile(suffix='.jpg', delete=False) as fid:
-        file_path = fid.name
-
-    gc.save(file_path)
-
-    return file_path
+    with tempfile.NamedTemporaryFile(suffix='.jpg') as fid:
+        gc.save(fid.name)
+        image = Image.from_file(fid.name, resist_width='weak',
+                                resist_height='weak')
+    return image
 
 
 class Demo(DemoFrame):
 
     def _create_component(self):
-        file_path = compiled_path()
-        image = Image.from_file(file_path, resist_width='weak',
-                                resist_height='weak')
+        image = compiled_path()
 
         container = ConstraintsContainer(bounds=[500, 500])
         container.add(image)
--- a/examples/kiva/compiled_path.py
+++ b/examples/kiva/compiled_path.py
@@ -33,7 +33,8 @@ def compiled_path():
     gc.set_stroke_color((0, 0, 1, 1))
     gc.draw_path_at_points(locs, path, STROKE)
 
-    file_path = tempfile.mktemp(suffix='.jpg')
+    with tempfile.NamedTemporaryFile(suffix='.jpg', delete=False) as fid:
+        file_path = fid.name
 
     gc.save(file_path)
 
--- a/examples/kiva/dash.py
+++ b/examples/kiva/dash.py
@@ -34,20 +34,19 @@ def dash(sz=(1000, 1000)):
     gc.close_path()
     gc.draw_path()
     t2 = perf_counter()
-    with tempfile.NamedTemporaryFile(suffix='.bmp', delete=False) as fid:
-        file_path = fid.name
-    gc.save(file_path)
+    with tempfile.NamedTemporaryFile(suffix='.bmp') as fid:
+        gc.save(fid.name)
+        image = Image.from_file(fid.name, resist_width='weak',
+                                resist_height='weak')
     tot_time = t2 - t1
     print('time:', tot_time)
-    return file_path
+    return image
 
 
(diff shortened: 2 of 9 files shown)


examples/kiva/compiled_path.py
```diff
@@ -33,7 +33,8 @@ def compiled_path():
     gc.set_stroke_color((0, 0, 1, 1))
     gc.draw_path_at_points(locs, path, STROKE)
 
-    file_path = tempfile.mktemp(suffix='.jpg')
+    with tempfile.NamedTemporaryFile(suffix='.jpg', delete=False) as fid:
+        file_path = fid.name
```
jwiggins, line 37: This won't work. The file is destroyed after the context goes away. You'll instead need to move the `with` block into `Demo._create_component` so that the file can exist during both the call to `compiled_path` and `Image.from_file`.

examples/kiva/dash.py
```diff
@@ -34,7 +34,8 @@ def dash(sz=(1000, 1000)):
     gc.close_path()
     gc.draw_path()
     t2 = perf_counter()
-    file_path = tempfile.mktemp(suffix='.bmp')
+    with tempfile.NamedTemporaryFile(suffix='.bmp', delete=False) as fid:
+        file_path = fid.name
```
jwiggins, line 38: Same here. There's no reason to use a context manager if you're not treating the file as something which is actually temporary.

jwiggins (commented):
All of these changes seem to ignore the "temporary" part of `NamedTemporaryFile`

examples/kiva/compiled_path.py
aaronayres35: That makes sense.
This would require changing the signature of `compiled_path` correct? Otherwise, if the context manager is in `Demo._create_component`, `compiled_path` would not have access to `fid`/`file_path`.

Also as a tangential question, it seems that rather than using `NamedTemporaryFile` in a context manager, we could use `tempfile.mkstemp` instead, without a context manager?  Also, I see that `Image.from_file` calls `PIL.Image.open(filename)`, but I'm not understanding when the file is intended to be closed/stop existing.  Ref: https://pillow.readthedocs.io/en/stable/reference/open_files.html#image-lifecycle

jwiggins: That's a good question. Looking into `Image.from_file`: https://github.com/enthought/enable/blob/582fb46b624ec6372ffe5e23b3b2d53d582b056e/enable/primitives/image.py#L37-L42

We see that the image is immediately converted to a numpy array. It should be safe to destroy the file after `Image.from_file` returns.

The end state we want to get here is to undo the usage of `tempfile.mktemp` and have examples that don't leave temporary files lying around after they exit. That should be possible by creating files long enough to write an image into them and read the data back. It looks like the same pattern is followed for every one of these examples. You might instead create the `Image` instance in the drawing functions which need the temporary file and then all the file handling can just be in one place.

Something like:
```python
def drawer():
    # ...
    with tempfile.NamedTemporaryFile(suffix='.bmp') as fid:
        gc.save(fid.name)
        image = Image.from_file(fid.name, resist_width='weak',
                                resist_height='weak')
    return image
```

examples/kiva/simple2.py
```diff
@@ -16,7 +16,8 @@ def simple():
     # directly manipulate the underlying Numeric array.
```
jwiggins, line 16: ^ There's still a lingering `tempfile.mktemp` call on line 14. (don't ask me why it's written that way)

examples/kiva/simple2.py
```diff
@@ -11,22 +11,23 @@ def simple():
     gc.set_fill_color((0, 0, 0))
     gc.rect(99, 100, 300, 300)
     gc.draw_path()
-    gc.save(tempfile.mktemp(suffix=".bmp"))
+
+    with tempfile.NamedTemporaryFile(suffix='.bmp') as fid:
+        gc.save(fid.name)
```
jwiggins, line 16: I'm a little curious if this is necessary at all or just left over from debugging/experimentation. Perhaps it's needed for a  side effect on the `bmp_array` attribute?

jwiggins (approved):
Thanks for the updates

examples/kiva/simple2.py
aaronayres35: I'm not seeing anything supporting why it would be needed, and the example seems to run the same with or without it

jwiggins: If the behavior is unchanged without it, I would remove it. It's not useful to have noise like this in example code.

jwiggins (approved):

PR #457, opened 2020-12-01
Use "is" instead of "==" to compare to None

### LINKED ISSUE #397: Use "is" instead of "==" to compare to None
Use "is" instead of "==" to compare to None, because "==" may call a custom "__eq__" method which is slower and error-prone.

"is" is generally preferred when comparing arbitrary objects to singletons like None because it is faster and more predictable. "is" always compares by object identity, whereas what "==" will do depends on the exact type of the operands and even on their ordering.

This recommendation is supported by PEP 8, which explicitly states that "comparisons to singletons like None should always be done with "is" or "is not", never the equality operators.

This issue occurs several places. A quick search for "== None" gives:
In enable:
  * scrolled.py (2 matches)

In kiva:
  * sources.py (2 matches)
  * tohtml.py (3 matches)
  * pdfmetrics.py
  * ps.py
svg.py

aaronayres35:
fixes #397 

This PR simply converts `==` comparisons to `None` to use `is` instead.

### DIFF
--- a/kiva/agg/freetype2/src/tools/docmaker/sources.py
+++ b/kiva/agg/freetype2/src/tools/docmaker/sources.py
@@ -208,7 +208,7 @@ class  SourceBlock:
         self.format    = processor.format
         self.content   = []
 
-        if self.format == None:
+        if self.format is None:
             return
 
         words = []
@@ -297,7 +297,7 @@ class  SourceProcessor:
             if line[-1] == '\012':
                 line = line[0:-1]
 
-            if self.format == None:
+            if self.format is None:
                 self.process_normal_line( line )
             else:
                 if self.format.end.match( line ):
--- a/kiva/agg/freetype2/src/tools/docmaker/sources.py
+++ b/kiva/agg/freetype2/src/tools/docmaker/sources.py
@@ -208,7 +208,7 @@ class  SourceBlock:
         self.format    = processor.format
         self.content   = []
 
-        if self.format is None:
+        if self.format == None:
             return
 
         words = []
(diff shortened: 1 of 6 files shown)


kiva/agg/freetype2/src/tools/docmaker/sources.py
```diff
@@ -208,7 +208,7 @@ def  __init__( self, processor, filename, lineno, lines ):
         self.format    = processor.format
         self.content   = []
 
-        if self.format == None:
+        if self.format is None:
```
jwiggins, line 211: This file is not relevant. It just happens to be in the vendored freetype tree that's here.

kiva/agg/freetype2/src/tools/docmaker/tohtml.py
```diff
@@ -437,7 +437,7 @@ def  index_exit( self ):
         self.index_items = {}
 
     def  index_dump( self, index_filename = None ):
-        if index_filename == None:
+        if index_filename is None:
```
jwiggins, line 440: Same here

jwiggins (approved):

kiva/agg/freetype2/src/tools/docmaker/sources.py
aaronayres35: Ah I see, should I revert this change then, or does it not matter

jwiggins: Yeah, I would leave all the vendored code alone

jwiggins (approved):
Thanks for the updates

PR #461, opened 2020-12-03
Remove all from __future__ imports as we are dropping all python <3.6 support

aaronayres35:
part of #415 

Currently CI is only running on python 3.6 and we plan to drop support for versions <3.6 with the next release.  This PR removes the no longer needed `from __futute__ import _____` statements in the codebase.


jwiggins (approved):
LGTM

PR #462, opened 2020-12-03
Remove try excepts for importing unittest.mock

aaronayres35:
part of #415 
`unittest.mock` became part of the standard library in python 3.3.  Since we are dropping all python versions <3.6 we no longer need to have `try ... except`s for these imports 

Also, as a result of this, `mock` can be removed from the list of dependencies in edmtool.py.

### DIFF
--- a/enable/testing.py
+++ b/enable/testing.py
@@ -1,9 +1,7 @@
 # Copyright (c) 2008-2013 by Enthought, Inc.
 # All rights reserved.
-try:
-    from unittest.mock import Mock
-except ImportError:
-    from mock import Mock
+from unittest.mock import Mock
+
 
 from enable.abstract_window import AbstractWindow
 from enable.events import MouseEvent, KeyEvent, DragEvent
--- a/enable/testing.py
+++ b/enable/testing.py
@@ -2,7 +2,6 @@
 # All rights reserved.
 from unittest.mock import Mock
 
-
 from enable.abstract_window import AbstractWindow
 from enable.events import MouseEvent, KeyEvent, DragEvent
 from kiva.testing import KivaTestAssistant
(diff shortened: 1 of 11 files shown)


enable/testing.py
```diff
@@ -1,9 +1,7 @@
 # Copyright (c) 2008-2013 by Enthought, Inc.
 # All rights reserved.
-try:
-    from unittest.mock import Mock
-except ImportError:
-    from mock import Mock
+from unittest.mock import Mock
+
```
jwiggins, line 4: Nit: Only one blank line needed between import groups

enable/tests/test_assistant_test_case.py
```diff
@@ -1,7 +1,4 @@
-try:
-    from unittest import mock
-except ImportError:
-    import mock
+from unittest import mock
 import nose
```
jwiggins, line 2: Note: `unitest` is std library and `nose` is 3rd party, so there should be a blank line between these imports.

enable/tests/tools/hover_tool_test_case.py
```diff
@@ -4,10 +4,8 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import unittest
-try:
-    from unittest import mock
-except ImportError:
-    import mock
+from unittest import mock
+
```
jwiggins, line 8: Nit: only one blank line

kiva/fonttools/tests/test_font_manager.py
```diff
@@ -1,9 +1,7 @@
 import os
 import unittest
-try:
-    from unittest import mock
-except ImportError:
-    import mock
+from unittest import mock
+
```
jwiggins, line 4: Nit: one blank line

jwiggins (approved):
Just some style quibbles, but otherwise LGTM

jwiggins (approved):
Thanks for the updates

PR #463, opened 2020-12-03
Remove python version checks 

aaronayres35:
part of #415 

This PR removes any explicit python version checks and only keeps the code paths that are specific to python versions we plan to continue to support (python >=3.6).

### DIFF
--- a/enable/tests/primitives/test_image.py
+++ b/enable/tests/primitives/test_image.py
@@ -5,10 +5,7 @@ import sys
 
 import six
 
-if sys.version_info[:2] == (2, 6):
-    import unittest2 as unittest
-else:
-    import unittest
+import unittest
 
 import numpy as np
 from numpy.testing import assert_array_equal
--- a/enable/tests/primitives/test_image.py
+++ b/enable/tests/primitives/test_image.py
@@ -2,10 +2,9 @@
 
 import os
 import sys
+import unittest
 
 import six
-
-import unittest
 
 import numpy as np
 from numpy.testing import assert_array_equal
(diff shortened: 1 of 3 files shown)


enable/tests/primitives/test_image.py
```diff
@@ -5,10 +5,7 @@
 
 import six
 
-if sys.version_info[:2] == (2, 6):
-    import unittest2 as unittest
-else:
-    import unittest
+import unittest
```
jwiggins, line 8: This belongs with the stdlib imports above `six`

kiva/fonttools/font.py
```diff
@@ -86,11 +86,6 @@ def __init__(self, face_name="", size=12, family=SWISS, weight=NORMAL,
             (type(underline) != int) or (not isinstance(face_name, six.string_types)) or \
             (type(encoding) != type(DEFAULT)):
                 raise RuntimeError("Bad value in Font() constructor.")
-        ### HACK:  C++ stuff expects a string (not unicode) for the face_name, so fix
-        ###        if needed.
-        ### Only for python < 3
-        if six.PY2 and isinstance(face_name, six.text_type):
-            face_name = face_name.encode("latin1")
```
jwiggins, line 93: Do we have test coverage for this code?

jwiggins (commented):
A couple comments

kiva/fonttools/font.py
aaronayres35: I ran the test suite under coverage, and the if statement was previously hit by the tests.  But since we only test on python 3.6 now, `six.PY2` is never true.  The `__init__` method is called in tests though yes.

jwiggins: 👍

jwiggins (approved):
Thanks for the update

PR #465, opened 2020-12-03
remove all "coding: UTF-8" headers

aaronayres35:
part of #415 

Python 3 reads source code in UTF-8 by default, so these can all be removed


jwiggins (approved):
LGTM

PR #466, opened 2020-12-03
Remove commented out code

aaronayres35:
This PR simply removes a bunch of commented out code from the code base.  There is certainly more that should be removed still floating around


jwiggins (approved):
😍

PR #469, opened 2020-12-04
TST: Switch on default warning flag for CI test command

### LINKED ISSUE #434: Switch on the default warning flag when running testsuite
See related https://github.com/enthought/envisage/issues/311

Deprecation warnings triggered during test runs are visible, but deprecation warnings triggered prior to when a unittest test case is loaded are by default hidden from the console.

With that, deprecation warnings from trait types definition do not show up from running the current test command, because the warnings occur at import time.

aaronayres35:
fixes #434 

This PR makes the exact same fix that was done in https://github.com/enthought/envisage/pull/326 and https://github.com/enthought/traitsui/pull/1059. It adds the adds the -W default flag.

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -262,10 +262,10 @@ def test(runtime, toolkit, pillow, environment):
     environ = environment_vars.get(toolkit, {}).copy()
     environ['PYTHONUNBUFFERED'] = "1"
     commands = [
-        "edm run -e {environment} -- python -W default -m"
-        "coverage run -m nose.core enable -v --nologcapture",
-        "edm run -e {environment} -- python -W default -m"
-        "coverage run -a -m nose.core kiva -v --nologcapture",
+        ("edm run -e {environment} -- python -W default -m"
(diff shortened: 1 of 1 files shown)


ci/edmtool.py
```diff
@@ -262,8 +262,10 @@ def test(runtime, toolkit, pillow, environment):
     environ = environment_vars.get(toolkit, {}).copy()
     environ['PYTHONUNBUFFERED'] = "1"
     commands = [
-        "edm run -e {environment} -- coverage run -m nose.core enable -v --nologcapture",
-        "edm run -e {environment} -- coverage run -a -m nose.core kiva -v --nologcapture",
+        "edm run -e {environment} -- python -W default -m"
+        "coverage run -m nose.core enable -v --nologcapture",
```
jwiggins, line 266: It might be helpful to wrap these multi-line commands in parentheses so that it's clear they belong together. Otherwise you have to be paying attention to the tiny commas that separate the commands.

jwiggins (approved):

jwiggins (approved):

PR #476, opened 2020-12-08
Add or update README files to mention what are vendored code

kitchoi:
Certain packages in enable are vendored. This information is relevant for development because knowing what are vendored code affects our strategy in changing them. e.g. We may reconsider an external dependency at some point, and we may not do trivial styling changes or renaming because it would make such move harder for little benefits. It also provides context when we change them.

This PR updates the README files in `kiva.agg` and `kiva.fonttools` to make it more obvious what are vendored code, based on offline discussion with @corranwebster.

Upon updating the README for `kiva.fonttools`, I also added matplotlib license file for the package. The same license file can be found in `kiva.agg` but `kiva.fonttools` is independent enough that it would be good to maintain the file independently as well.


kiva/agg/README.rst
```diff
@@ -1,4 +1,17 @@
...
+`Anti-grain geometry <http://www.antigrain.com/>`_ drawing library into a
+Python extension and exposes the Kiva API.
+
+Third-party libraries
+---------------------
+
+agg-24
+~~~~~~
+
+The folder ``agg24`` contains a copy of the Agg-2.4 library.
```
jwiggins, line 14: The folder name is `agg-24`

jwiggins (approved):
Just one comment, otherwise LGTM

jwiggins (approved):

PR #478, opened 2020-12-08
Fix documentation build failing on Python 3

kitchoi:
When I tried to build the documentation with the following commands:
```
# In an environment with enable source installed
pip install sphinx
cd docs
make html
```

I get the following error:
```
  File "/Users/kchoi/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/sphinx/config.py", line 319, in eval_config_file
    execfile_(filename, namespace)
  File "/Users/kchoi/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/sphinx/util/pycompat.py", line 89, in execfile_
    exec(code, _globals)
  File "/Users/kchoi/Work/ETS/enable/docs/source/conf.py", line 42, in <module>
    execfile(os.path.join('..', '..', 'enable', '_version.py'), d)
NameError: name 'execfile' is not defined
```

`execfile` is a Python 2 built-in. This PR fixes the config file so that the config file works on Python 3. We probably want to add a `docs` command in `ci/edmtool.py` like in the other ETS projects to streamline this step, and possibly run this command as part of CI, but that can come in separate PRs/issues.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -9,6 +9,7 @@
 # All configuration values have a default value; values that are commented out
 # serve to show the default value.
 
+import runpy
 import sys, os
 
 # If your extensions are in another directory, add it here. If the directory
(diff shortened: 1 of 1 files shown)


jwiggins (approved):
LGTM

PR #479, opened 2020-12-08
Drop python2 specific imports

### LINKED ISSUE #415: Drop Python 2 Support

aaronayres35:
closes #415 

This PR removes a couple import try excepts that were python 2 specific (e.g. `time.perf_counter` is always available since python 3.3 so we no longer need the except clause)


rahulporuri (approved):
LGTM

PR #489, opened 2020-12-14
Use config for bootstrapping a EDM environment

kitchoi:
Currently CI builds are failing because the default config created by EDM includes a repository to `enthought/commercial` that has been ~deprecated~ (edited: removed). The default config should be fixed in EDM. In the meantime, this PR attempts to workaround the issue by specifying a config file to use.

(Orthogonal: Presumably enable should always provide a config file in all the steps involving EDM... but at the same time, if we had always provided a config, it would be harder for such issue to reveal themselves.)

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -148,6 +148,9 @@ CONTEXT_SETTINGS = {
     "help_option_names": ["-h", "--help"],
 }
 
+# Config file for EDM
+EDM_CONFIG = os.path.join(ROOT, "ci", ".edm.yaml")
+
 
 @click.group(context_settings=CONTEXT_SETTINGS)
 def cli():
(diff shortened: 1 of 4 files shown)


ci/edmtool.py
```diff
@@ -174,7 +177,7 @@ def install(runtime, toolkit, pillow, environment, source):
     # edm commands to setup the development environment
     commands = [
         "edm environments create {environment} --force --version={runtime}",
-        "edm install -y -e {environment} {packages}",
+        "edm --config {edm_config} install -y -e {environment} {packages}",
```
jwiggins, line 180: Looks like the `edm env create ...` command also needs the config

jwiggins:
Didn't realize EDM did this... I'm ready to approve as soon as CI is green

ci/edmtool.py
kitchoi: Well spotted. Thank you.

jwiggins (approved):

PR #284, opened 2017-07-25
Python3 compatibility fixes

jdeschenes:
This is a fix to remove 2to3 when installing enable.

PR #492, opened 2020-12-14
Fix AttributeError on Python 3.9 due to usage of removed API in xml

### LINKED ISSUE #490: Test errors on Python 3.9 due to deprecated usage of ElementTree.getiterator
Running the test suite on Python 3.9 reveals quite a number of errors coming from the same usage of `ElementTree.getiterator`. This is one example:

```
ERROR: test_circle (kiva.tests.test_svg_drawing.TestSVGDrawing)
  File "/Users/kchoi/Work/ETS/enable/kiva/tests/drawing_tester.py", line 58, in test_circle
    self.gc.stroke_path()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/contextlib.py", line 124, in __exit__
    next(self.gen)
  File "/Users/kchoi/Work/ETS/enable/kiva/tests/test_svg_drawing.py", line 20, in draw_and_check
    elements = [element for element in tree.getiterator()]
AttributeError: 'ElementTree' object has no attribute 'getiterator'
```

The `getiterator` method has been deprecated since 3.2
https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getiterator

kitchoi:
Closes #490

The first commit fixes errors that can be seen by running the test suite on Python 3.9.

The second commit makes similar changes, but those changes are not covered by tests.

On the changes that are not covered by test:
For `enable.svg.document.addRectToDocument`. I was trying to write a test that does something like this:
```
    def test_add_rect_to_document_with_clip_path(self):
       # create_svg_text is a function that adds the necessary XML declaration and <svg> root element
        svg_text = create_svg_text(
            '<rect x="11" y="11" clip-path="circle() view-box" />'
        )
        document = document.SVGDocument(etree.parse(StringIO(svg_text)).getroot(), renderer=KivaRenderer())
```
However it still does not reach the code path where I fixed the usage of ElementTree.
It seems that [`if 'clip-path' in node:`](https://github.com/enthought/enable/blob/6b41a429f5cfddc70fa67c3a1025cdd426a98988/enable/savage/svg/document.py#L646) is always false. Previously it was doing `if 'clip-path' in node.keys()` where [`keys()` returns the XML attributes](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.keys), but `node` was thought to be a `dict` during a migration from Python 2 to 3 (see #284).  Upon fixing `in node` to `in node.keys()`, I run into another difficulty trying to open an SVG referenced by the clip-path assuming it was a local file. So I eventually gave up. I am happy to revert the second commit owing to the fact it is untested.

### DIFF
--- a/enable/savage/compliance/viewer.py
+++ b/enable/savage/compliance/viewer.py
@@ -68,7 +68,7 @@ class XMLTree(wx.TreeCtrl):
             node = self.AppendItem(node, txt)
             self.SetPyData(node, element)
         #children
-        for child in element.getchildren():
+        for child in element:
             self.addElementToTree(child, node)
         #attributes
         for key, value in element.items():
--- a/enable/savage/svg/document.py
+++ b/enable/savage/svg/document.py
@@ -653,7 +653,7 @@ class SVGDocument(object):
             clip_path = self.renderer.makePath()
             ops.extend(self.createTransformOpsFromNode(element))
             ops.extend(self.generatePathOps(clip_path))
-            for child in element.getchildren():
+            for child in element:
                 cpath, cops = self.processElement(child)
                 if cpath:
                     clip_path.AddPath(cpath)
--- a/enable/savage/svg/document.py
+++ b/enable/savage/svg/document.py
@@ -302,7 +302,7 @@ class SVGDocument(object):
         attribute it may be given.
         """
         idmap = {}
-        for e in element.getiterator():
+        for e in element.iter():
             id = e.get('id', None)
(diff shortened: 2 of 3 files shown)


jwiggins (approved):
Less broken than before... Let's iterate as needed here.

kitchoi:
Thanks. I will open an issue for the clip-path code path being unreachable.

PR #494, opened 2020-12-14
Contribute examples to etsdemo

### LINKED ISSUE #436: Contribute examples to the etsdemo application
The enable GitHub repository has a number of examples - https://github.com/enthought/enable/tree/master/examples. For the upcoming release of the enable package, it would be useful to contribute said examples to the etsdemo application.

aaronayres35:
fixes #436

This PR contributes the kiva and enable examples to etsdemo.

~It is currently flagged as a WIP, as some examples are broken / have strange behaviors.  However, some of these problems may best be solved in a separate PR.~
EDIT: I have opened issues for any broken demos, and have added comments to this PR for any demos that work when run independently, but not as we want them to when run through etsdemo.

To run the etsdemo app:
```
pip install -e .[demo]
pip install etsdemo[pyside2]  # or a different gui toolkit
etsdemo
``` 

Additionally, some of the kiva examples simply save a file, rather than having a showcased demo.  e,g. `dummy`, `pdf_arcs`, `star_path`.  Not sure what is best approach for these.

I removed the WIP marker, as I think the lingering issues are likely best handled in separate PRs.


enable/__init__.py
```diff
@@ -20,6 +20,6 @@
     __requires__.append('pillow')
 
 __extras_require__ = {
-    'demo': ['chaco', 'mayavi', 'scipy', 'traitsui[demo]'],
+    'demo': ['chaco', 'mayavi', 'scipy', 'traitsui[demo]', 'kiwisolver'],
```
aaronayres35, line 23: This is a bit orthogonal to this PR, but so many examples depend on `ConstraintsContainer` that it felt appropriate to add.  If one does `pip install enable[demo]` and `pip install etsdemo[pyside2]` and they launch the demo app, they would expect to have what they need to see all the demos.  

I believe there are other dependencies as well (e.g. `pyglet` for `pyglet_gl.py`), but they are much less common.  I may end up adding to them to this PR as well?

aaronayres35:
`kiva_explorer` is a great example, but it does not work with the etsdemo application.
I am not sure what to do with it, as it is a bit special

aaronayres35:
CI is failing on the enable added test, but I can't reproduce locally

aaronayres35:
> CI is failing on the enable added test, but I can't reproduce locally

I recreated the environment and I can now reproduce the test failure, however if I run `pip install -e .` I no longer see the error

ci/edmtool.py
```diff
@@ -228,7 +228,7 @@ def install(runtime, toolkit, pillow, environment, source):
     # dependencies or we risk testing against an released enable.
     install_local = (
         "edm run -e {environment} -- "
-        "pip install --force-reinstall --no-dependencies " + ROOT
+        "pip install --force-reinstall --no-dependencies -e ."
```
aaronayres35, line 231: This is definitely not the fix we want.  I just pushed this as a sanity check. I still do not exactly understand why the added test fails without this change.

Independent of this though, we probably want to open a separate issue regarding adding an `editable` click option as has been done for other ets projects.  Further, I am not sure why `enable` was using:
`#: Path to the top-level source directory
ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))`
to do this rather than `.` as is done in the other ets projects

jwiggins (commented):
Just some minor comments. I didn't look into the testing problems at all.

kiva/examples/kiva/kiva_explorer.py
```diff
@@ -0,0 +1,132 @@
...
+            ),
+        ),
+        resizable=True,
+        title='Kiva Explorer',
+    )
+
+
+if __name__ == '__main__':
+    demo = ScriptedComponentView()
+    demo.configure_traits()
```
aaronayres35, line 132: I left original `kiva_explorer.py` sitting in `examples` as it is currently still best run outside of etsdemo.  Further, in the etsdemo application the demo is still not shown here, all that is visible is the code at the moment.

ci/edmtool.py
```diff
@@ -228,7 +228,7 @@ def install(runtime, toolkit, pillow, environment, source):
     # dependencies or we risk testing against an released enable.
     install_local = (
         "edm run -e {environment} -- "
-        "pip install --force-reinstall --no-dependencies -e ."
+        "pip install --force-reinstall --no-dependencies " + ROOT
```
aaronayres35: I'm not exactly sure why, but I'm not seeing the test failure anymore

kitchoi: Previously the test failure reflects the issue that I ran into 😅 : the examples files are moved inside `demo` subdirectory but the `package_data` has not been updated to say `demo/enable/*` etc.
Using `pip install -e .` "fixes" the test failure because it actually makes the package available from source: the package_data is not really relevant in that setup.

After https://github.com/enthought/enable/pull/494/commits/7cecefbdf62fcd26ff5be1a40dcae4cb5f7d8dc8, the `package_data` is fixed to reference the `demo` subdirectory, and that truly fixes the test.

This is in fact one instance where it is valuable to have a CI job where we don't install the source with editable mode. It tests the installed environment more truthfully.

kiva/examples/kiva/kiva_explorer.py
aaronayres35: I think the current status is about the best we are going to get with `kiva_explorer` in etsdemo for now.  Eventually, we may want to rework things so that maybe when you click run the explorer application pops up separately or something like that.  

I think I will delete the `kiva_explorer.py` that is still sitting in the original `examples` directory (actually I will delete `examples` entirely) as it can still be run the same way independent of etsdemo where it sits now.

On that note though, I am not a huge fan of where it sits now because it comes off as just another kiva example, but it is inherently a bit different, and would be nice if that distinction were clearer

aaronayres35:
The `vtk_example.py` now sitting in `enable/examples/demo/enable` does not work with `etsdemo` currently.
For one thing it imports from another example `from basic_move import Box`, but it also is using
```
if __name__=="__main__":
    main()
```
and using `mlab.show()` to show the example.

aaronayres35:
https://github.com/enthought/enable/blob/7b56e5275c707e7ee5749ce4ba141240729f5a26/examples/enable/tools/apptools/undoable_move_tool.py
Is also not working in etsdemo but works when run on its own

ci/edmtool.py
aaronayres35: Ahhh I see, that makes sense now, thank you!

aaronayres35:
https://github.com/enthought/enable/blob/7b56e5275c707e7ee5749ce4ba141240729f5a26/examples/kiva/pyglet_gl.py as well is not etsdemo friendly

aaronayres35:
Note that https://github.com/enthought/enable/blob/7b56e5275c707e7ee5749ce4ba141240729f5a26/examples/kiva/qt4_simple.py fails with pyside2 with this error:
```
$ python qt4_simple.py 
  File "qt4_simple.py", line 56, in <module>
    app = QtGui.QApplication(sys.argv)
RuntimeError: Please destroy the QApplication singleton before creating a new QApplication instance.
```
But runs fine with pyqt5.  However, `etsdemo` can not be run with pyqt5 because of the `QtWebKit` issue.
<details>
```
 $ etsdemo
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/pyface/qt/QtWebKit.py", line 19, in <module>
    from PyQt5.QtWebEngine import *
ModuleNotFoundError: No module named 'PyQt5.QtWebEngine'

  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/tree_editor.py", line 944, in _on_tree_sel_changed
    parent=editor, view=view, kind="subpanel"
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traits/has_traits.py", line 1678, in edit_traits
    args,
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/view.py", line 462, in ui
    ui.ui(parent, kind)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/ui.py", line 244, in ui
    self.rebuild(self, parent)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/toolkit.py", line 147, in ui_subpanel
    ui_panel.ui_subpanel(ui, parent)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 62, in ui_subpanel
    _ui_panel_for(ui, parent, True)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 68, in _ui_panel_for
    ui.control = control = _Panel(ui, parent, is_subpanel).control
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 130, in __init__
    self.control = panel(ui)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 259, in panel
    panel = _GroupPanel(content[0], ui).control
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 617, in __init__
    layout = self._add_items(content, inner)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 883, in _add_items
    ui, object, name, item.tooltip, None
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/editor_factory.py", line 168, in readonly_editor
    description=description,
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/editors/html_editor.py", line 34, in html_editor
    return toolkit_object("html_editor:SimpleEditor")(*args, **traits)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/toolkit.py", line 71, in toolkit_object
    obj = _toolkit(name)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/pyface/base_toolkit.py", line 152, in __call__
    module = import_module(mname, package)
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/traitsui/qt4/html_editor.py", line 23, in <module>
    from pyface.qt import QtCore, QtGui, QtWebKit
  File "/Users/aayres/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/pyface/qt/QtWebKit.py", line 28, in <module>
    from PyQt5.QtWebKit import *
ModuleNotFoundError: No module named 'PyQt5.QtWebKit'
Abort trap: 6
```
</details>

kitchoi (commented):
Some preliminary review comments. I understand this is still a work-in-progress. Happy to review this again.

enable/__init__.py
```diff
@@ -20,6 +20,13 @@
...
 __extras_require__ = {
-    'demo': ['chaco', 'mayavi', 'scipy', 'traitsui[demo]'],
+    'demo': [
+        'chaco',
+        'mayavi',
+        'scipy',
+        'traitsui[demo]',
+        'kiwisolver',
+        'pyglet'
+    ],
```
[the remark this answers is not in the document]
aaronayres35: > Do you think we can take this opportunity to replace "traitsui[demo]" to just "traitsui"?
`traitsui` alone already exists under `__requires__`, so maybe this can be removed entirely?  If it stays it should definitely be replaced by `traitsui[examples]`.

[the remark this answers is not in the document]
kitchoi: > traitsui alone already exists under __requires__, so maybe this can be removed entirely?

Yes you are right. Please just remove `traitsui[demo]` entirely.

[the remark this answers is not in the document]
kitchoi: > rename "demo" to "examples" too

I just realized that this might be confusing as the word `demo` appears in two places. Sorry about that. I meant renaming the "demo" in the *key* of the dictionary, not the "demo" in the `traitsui[*]` value.

kitchoi (approved):
For the ones that don't work well, since we are exposing them anyway, what do you think about adding docstrings to warn about that instead of letting the user finding it out for themselves and getting disappointed?
e.g. https://github.com/enthought/traitsui/blob/98b6593d4e8eabb010982861ae977cc8a9d66f79/traitsui/examples/demo/Standard_Editors/TableEditor_demo.py#L5
It is also useful to add a comment referencing the issue you opened. https://github.com/enthought/traitsui/blob/98b6593d4e8eabb010982861ae977cc8a9d66f79/traitsui/examples/demo/Standard_Editors/TableEditor_demo.py#L17

Not only this will set the user's expectation upfront, later we can track down the issue and determine if the warning can be removed. e.g. I was able to remove one warning entirely when the sole issue is resolved: https://github.com/enthought/traitsui/commit/1b7de7821374fc23cfff9cfa7a2d4413b42defac#diff-c79f163118f7b6def1f8f0f3fbff5736f8c1679f9d387c6dc4a411a8d437c887

It that sounds good to you, please feel free to do so in a separate PR as it is slightly broader scope.

aaronayres35:
> For the ones that don't work well, since we are exposing them anyway, what do you think about adding docstrings to warn about that instead of letting the user finding it out for themselves and getting disappointed?
> e.g. https://github.com/enthought/traitsui/blob/98b6593d4e8eabb010982861ae977cc8a9d66f79/traitsui/examples/demo/Standard_Editors/TableEditor_demo.py#L5
> It is also useful to add a comment referencing the issue you opened. https://github.com/enthought/traitsui/blob/98b6593d4e8eabb010982861ae977cc8a9d66f79/traitsui/examples/demo/Standard_Editors/TableEditor_demo.py#L17
> 
> Not only this will set the user's expectation upfront, later we can track down the issue and determine if the warning can be removed. e.g. I was able to remove one warning entirely when the sole issue is resolved: [enthought/traitsui@1b7de78#diff-c79f163118f7b6def1f8f0f3fbff5736f8c1679f9d387c6dc4a411a8d437c887](https://github.com/enthought/traitsui/commit/1b7de7821374fc23cfff9cfa7a2d4413b42defac#diff-c79f163118f7b6def1f8f0f3fbff5736f8c1679f9d387c6dc4a411a8d437c887)
> 
> It that sounds good to you, please feel free to do so in a separate PR as it is slightly broader scope.

SGTM!  I might also mention on the ones that work stand-alone but not in etsdemo that changing `__main__` to `___main___` in the code editor will work, and add a reference to the issue https://github.com/enthought/traitsui/issues/1428.  Does that make sense?

kitchoi:
> add a reference to the issue enthought/traitsui#1428.

Adding reference to `___main___` (enthought/traitsui#1428) is a bit tricky, because then when enthought/traitsui#1428 is fixed and released, you have to make releases for all the demo files that have that reference, otherwise the reference would be extremely unhelpful and misleading. So probably not... we will have to fix the issue. Now that we have seen examples in different context, it is a good place to be to decide how we want to fix it.

enable/examples/demo/enable/canvas_demo.py
```diff
@@ -5,9 +5,9 @@
 
 class Demo(DemoFrame):
 
-    def create_component(self):
+    def _create_component(self):
```
aaronayres35, line 8: part of #500 
All this required to fix was this leading underscore so I figured it was fine to add with this PR.

The issue name is misleading, as the other examples in the comments of the issue are still independently broken.  they need to overwrite the `_create_component` method.  I will update issue and try to fix those in a later PR

enable/examples/demo/enable/canvas_demo.py
```diff
@@ -5,9 +5,9 @@
...
         canvas = Canvas(bgcolor="lightsteelblue", draw_axes=True)
-        from basic_move import Box
+        from enable.examples.demo.enable.basic_move import Box
```
aaronayres35, line 10: @kitchoi I was googling namespace packages in python, and was trying to understand but still am a bit confused.  It seems that doing this, without adding any `__init__.py` files works as we want it to, but I want to confirm I am not missing anything.

jwiggins: Does `from .basic_move import Box` not work?

kitchoi: The context while running the example in the `etsdemo` GUI application is a little different.The Python path there is the Python path used for running the GUI application, which is not necessarily the same as the Python path used for running the script from the command line. The absolute import removes assumption on the Python path.

This is the block of code relevant while this file is run from the GUI application, the triple-underscore-main thing can be found here:
https://github.com/enthought/traitsui/blob/e913f98d10e96624245c603d856b33ded665d565/ets-demo/etsdemo/app.py#L462-L487

kitchoi: > It seems that doing this, without adding any __init__.py files works as we want it to

This was what I expected as well.

kitchoi: @jwiggins comment is valid as in it shows the cognitive load required to maintain these examples. See https://github.com/enthought/traitsui/issues/1455

Should https://github.com/enthought/traitsui/issues/1455 be fixed, `from .basic_move import Box` should work (I think).

jwiggins: > Should enthought/traitsui#1455 be fixed, `from .basic_move import Box` should work (I think).

That seems like the better solution here

enable/examples/demo/enable/canvas_demo.py
kitchoi: This makes sense. Thank you!

enable/examples/demo/enable/canvas_demo.py
kitchoi: Sorry, I was wrong...  I should never guess but try for myself... :( 

`from .basic_move import Box` does not work, if I try to run the script from the command line (I am using master with the `create_component` -> `_create_component` fix:
```
 $ pwd
/Users/kchoi/Work/ETS/enable/examples/enable
$ python canvas_demo.py
...
  File "/Users/kchoi/Work/ETS/enable/enable/example_support.py", line 40, in _component_default
    return self._create_component()
  File "canvas_demo.py", line 10, in _create_component
    from .basic_move import Box
ModuleNotFoundError: No module named '__main__.basic_move'; '__main__' is not a package
```

kitchoi: The absolute import here does work on both the etsdemo application and from the command line.
`from basic_move import Box` does not work on etsdemo but will work if https://github.com/enthought/traitsui/issues/1455 is fixed.

@jwiggins Is it okay for us to proceed with the absolute import here to decouple this import from the pending issue?

kitchoi: From offline discussion: John says yes. :D

kitchoi (approved):
LGTM

PR #498, opened 2020-12-15
Remove USE_FONTCONFIG in font_manager

kitchoi:
The USE_FONTCONFIG is a module level constant to be flipped to change the behaviour of `kiva.fonttools.font_manager` at import time. It is hard to see how this can be used apart from running the file as a script.

This PR removes code that is never run due to the fact that `USE_FONTCONFIG` is false.


jwiggins (approved):
LGTM.

I was about to suggest that importing the module and changing the value of `USE_FONTCONFIG` would give the ability to run this code path, but even that's not possible.

I assume you still want to remove `findfont`, but didn't want to mix it up with this change?

kitchoi:
> I assume you still want to remove findfont, but didn't want to mix it up with this change?

Yes I do. It was very tempting though. 😅

PR #507, opened 2020-12-16
Use pyface undo instead of apptools.undo

### LINKED ISSUE #503: DeprecationWarning from importing apptools.undo
`enable.tools.apptools.commands` depends on `apptools.undo`, but the package has been moved to pyface in 7.2.0.

The cron job run 2 days ago showed this deprecation warning:
```
/home/travis/.edm/envs/enable-test-3.6-null-pillow/lib/python3.6/site-packages/enable/tools/apptools/commands.py:19: DeprecationWarning: apptools.undo is deprecated and will be removed in a future release. The functionality is now available via pyface.undo
  from apptools.undo.api import AbstractCommand
```

aaronayres35:
fixes #503 
 
As of Pyface 7.2.0, the undo/redo functionality that used to be specific to `apptools` is now available through pyface.  With the upcoming release of apptools, it will be deprecated. 

This PR simply replaces all uses of `apptools.undo` with `pyface.undo` and updates file names / documentation accordingly.  

Note that after this PR, the only use of `apptools` in the code base will be here:
https://github.com/enthought/enable/blob/f83869ed00eba67d36a3ced875e8de4476900058/enable/qt4/base_window.py#L499-L505
Maybe this code could use `pathlib` instead as mentioned https://github.com/enthought/apptools/issues/143?

`apptools` is still listed as a dependency in `ci/edmtool.py` but maybe that can be removed?

Also, note:  this PR will cause merge conflicts with #494, but they should be simple enough to resolve (just replace apptools with pyface in the example code that mentions it on 494)


jwiggins:
Concerning the drag and drop handling for Qt... I think some spelunking will be required to see how hard it would be to get rid of apptools (who's using it, what are they doing with the `File` objects, etc.). Although we're prepping for a major release, so now is probably the best time to "break" it.

enable/tools/pyface/api.py
```diff
@@ -6,20 +6,6 @@
 # LICENSE.txt
 #
```
jwiggins, line 7: You're renaming an `api` module here in a pretty significant way. Perhaps it would be better to leave the old module behind and import from the new module [Maybe even raise a warning for code which is importing from the old module]? @corranwebster How kind are we being to downstream code with changes like this? Do you have a preference in any direction?

examples/enable/tools/button_tool.py
```diff
@@ -10,7 +10,7 @@
 Undoable Move Tool
 ==================
 
-This example shows how to integrate a simple component move tool with apptools
+This example shows how to integrate a simple component move tool with pyface
 undo/redo infrastructure.
```
jwiggins, line 14: This is a lie. There's no undo/redo that I can see in this file.

jwiggins (commented):
A question and a comment. This should be good to go once the question is resolved.

enable/tools/pyface/api.py
corranwebster: We're being kind, but this is a major release.  I'd suggest keeping `apptools/api.py` around, but import from the pyface `api.py`.

examples/enable/tools/button_tool.py
corranwebster: Yes - it looks like a copy-paste from https://github.com/enthought/enable/blob/master/examples/enable/tools/apptools/undoable_move_tool.py

Probably the fault of 5-years-ago Corran.

enable/tools/pyface/api.py
corranwebster: Oh, and raise appropriate deprecation warnings for importing from `enable.tools.apptools.api`

jwiggins:
There was a slight misunderstanding. You should delete everything in `enable.tools.apptools` _except_ `__init__.py` and `api.py`. `api.py` should forward imports from the new `enable.tools.pyface.api` module [as well as generating a warning].

aaronayres35:
> There was a slight misunderstanding. You should delete everything in `enable.tools.apptools` _except_ `__init__.py` and `api.py`. `api.py` should forward imports from the new `enable.tools.pyface.api` module [as well as generating a warning].

so the warning should be generated from the `api` or from the `__init__.py`?  when deprecating `apptools.undo` itself I raised the warning in `__init__.py`, I'm not exactly sure what is the more correct thing to do

jwiggins:
`__init__.py` is probably the better place

enable/tools/apptools/__init__.py
```diff
@@ -0,0 +1,10 @@
+import warnings
+
+warnings.warn(
+    ("apptools.undo is deprecated and will be removed in a future release. The"
+     " functionality is now available via pyface.undo. As a result,"
+     " enable.tools.apptools has been deprecated in favor of"
+     " enable.tools.pyface."),
+    DeprecationWarning,
+    stacklevel=2
+)
```
jwiggins, line 10: This file needs a copyright header

jwiggins (approved):
One comment, otherwise LGTM

enable/tools/apptools/__init__.py
aaronayres35: I added the same one used in the `enable/tools/apptools/api.py`.  I think these will likely need to be changed once we start using https://github.com/enthought/ets-copyright-checker

jwiggins (approved):

PR #510, opened 2020-12-21
Make kiva_explorer a popup

### LINKED ISSUE #509: Make Kiva Explorer demo a popup
Rather than having Kiva explorer compressed into a tab, it probably makes sense to have it come in as a popup.  This is just a matter of changing:

https://github.com/enthought/enable/blob/2ae4f5c3f268d6f8e0de18f3022ba3cb271f5158/kiva/examples/kiva/kiva_explorer.py#L129-L132

to something like:
```
popup = ScriptedComponentView()

if __name__ == '__main__':
    popup.configure_traits()
```

This is not perfect, but better than what we have now.

aaronayres35:
fixes #509 

This is very short PR which simply changes `kiva_explorer` to come up as a popup rather than trying to compress it into the demo tab of the etsdemo application.  All this requires is a change to use the magic named `popup`

### DIFF
--- a/kiva/examples/kiva/kiva_explorer.py
+++ b/kiva/examples/kiva/kiva_explorer.py
@@ -126,6 +126,9 @@ class ScriptedComponentView(ModelView):
         title='Kiva Explorer',
     )
 
+# "popup" is a magically named variable for the etsdemo application which will
+# cause this demo to be run as a popup rather than trying to compress it into
+# a tab on the application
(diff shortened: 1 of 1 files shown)


jwiggins (approved):
You should add a comment explaining the magic naming. Otherwise LGTM

jwiggins (approved):
Thanks for the update.

PR #511, opened 2020-12-21
Clean-up a couple demos

aaronayres35:
This PR cleans up a couple demos by using absolute imports, using magically named `Demo`, and over writing `_create_component` instead of `_create_window` so that these demos can be run / run smoothly through the etsdemo application

### DIFF
--- a/enable/examples/demo/enable/tools/pyface/context_menu.py
+++ b/enable/examples/demo/enable/tools/pyface/context_menu.py
@@ -4,7 +4,7 @@ component is created and added to a container.
 """
 
 from enable.example_support import DemoFrame, demo_main
-from enable.api import Component, Container, Window
+from enable.api import Component, Container
 from enable.tools.pyface.context_menu_tool import ContextMenuTool
 
 from pyface.action.api import MenuManager, Action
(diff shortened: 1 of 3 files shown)


enable/examples/demo/enable/tools/pyface/context_menu.py
```diff
@@ -34,7 +34,9 @@ def _create_window(self):
         box.tools.append(context_menu)
         container = Container(bounds=[500,500])
         container.add(box)
-        return Window(self, -1, component=container)
```
jwiggins, line 37: Don't forget to remove orphaned imports

enable/examples/demo/enable/tools/pyface/context_menu.py
```diff
@@ -34,7 +34,9 @@ def _create_window(self):
...
+
+        return container
+
 
 if __name__ == "__main__":
-    demo_main(MyFrame)
+    demo_main(Demo)
```
aaronayres35, line 42: This example still doesn't seem to be working as it is supposed to (I'm not seeing a context menu).  These changes just allow it to run and run in `etsdemo` in its current state

jwiggins (approved):
Thanks for the updates

PR #513, opened 2020-12-24
Update CI requirements

corranwebster:
This:
- installs requirements from EDM where possible,
- removes the option to use older versions of Pillow,
- installs the most recent Pyglet from PyPI

### DIFF
--- a/ci/requirements.txt
+++ b/ci/requirements.txt
@@ -1,6 +1,2 @@
 pygarrayimage
-hypothesis<3.0.0
-fonttools==3.28.0
-reportlab<=3.1
-kiwisolver ; python_version <= "2.7"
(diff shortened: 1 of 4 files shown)


ci/requirements.txt
```diff
@@ -1,6 +1,2 @@
 pygarrayimage
-hypothesis<3.0.0
-fonttools==3.28.0
-reportlab<=3.1
-kiwisolver ; python_version <= "2.7"
-https://bitbucket.org/pyglet/pyglet/get/pyglet-1.1.4.zip ; python_version <= "2.7"
+pyglet
```
jwiggins, line 2: Do we want to be picky about a version here?

corranwebster: Don't know.  I would default to most recent, but it likely depends on what the GL backend needs.  I don't think its a fast-moving API.

jwiggins (approved):
LGTM

PR #515, opened 2020-12-24
Remove quartz CGGLContext

### LINKED ISSUE #514: Remove CGGLContext from Quartz backend
This relies on the Quartz `CGGLContext` which was deprecated in OS X 10.6 has not been available since OS X 10.10.  On recent XCode versions this prevents the Quartz backend from building.

This will also require removal of the appropriate .pxi definitions.

corranwebster:
Fixes #514.  Requires #513 since CI won't get to this point without that change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,8 @@ env:
 
 matrix:
   include:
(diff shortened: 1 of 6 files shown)


jwiggins (approved):
LGTM

PR #518, opened 2020-12-30
Add module docstrings to kiva and enable api modules

rahulporuri:
This PR adds module docstrings to `enable.api` and `kiva.api`, in an effort to improve the api documentation of kiva and enable.

### DIFF
--- a/enable/api.py
+++ b/enable/api.py
@@ -138,13 +138,6 @@ Enable Widgets
 ==============
 
 - :class:`~.AbstractWindow`
-- :class:`~.NativeScrollBar`
-- :class:`~.Compass`
-- :class:`~.Scrolled`
-- :class:`~.Slider`
-- :class:`~.TextFieldStyle`
-- :class:`~.TextField`
-- :class:`~.TextFieldGrid`
 - :class:`~.Viewport`
 - :class:`~.Window`
 
--- a/enable/api.py
+++ b/enable/api.py
@@ -1,6 +1,161 @@
 """ Enable is an interactive graphical component framework built on top of Kiva.
 
 See https://www.enthought.com/enthought/wiki/EnableProject
+
+Enable Base
+===========
+
+- :class:`~.IDroppedOnHandler`
+- :func:`~.str_to_font`
+- :func:`~.intersect_bounds`
+
+Constants
+---------
+- :attr:`~.TOP`
+- :attr:`~.VCENTER`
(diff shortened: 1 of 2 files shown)


kiva/api.py
```diff
@@ -1,3 +1,142 @@
...
+- :func:`~.trs_factor`
+- :func:`~.transform_point`
+- :func:`~.transform_points`
+
+Constants
+---------
+
+- :attr:`~.IDENTITY` - represents the :func:`~.affine_identity` matrix.
+
+Drawing Utilities
```
jwiggins, line 28: Nitpick: (also 🚲🏡) "Utilities" seems like it should include functions, but all of these are constants.

kiva/api.py
```diff
@@ -1,3 +1,142 @@
...
+
+- :attr:`~.POINT`
+- :attr:`~.LINE`
+- :attr:`~.LINES`
+- :attr:`~.RECT`
+- :attr:`~.CLOSE`
+- :attr:`~.CURVE_TO`
+- :attr:`~.QUAD_CURVE_TO`
+- :attr:`~.ARC`
+- :attr:`~.ARC_TO`
```
jwiggins, line 83: None of these are used anywhere in kiva. We shouldn't document them.

kiva/api.py
```diff
@@ -1,3 +1,142 @@
...
+
+Subpath Context Manager Constants
+---------------------------------
+
+- :attr:`~.SCALE_CTM`
+- :attr:`~.TRANSLATE_CTM`
+- :attr:`~.ROTATE_CTM`
+- :attr:`~.CONCAT_CTM`
+- :attr:`~.LOAD_CTM`
```
jwiggins, line 92: These are also of dubious usefulness to external code. I would omit them here.

enable/api.py
```diff
@@ -1,6 +1,161 @@
...
+==============
+
+- :class:`~.AbstractWindow`
+- :class:`~.NativeScrollBar`
+- :class:`~.Compass`
+- :class:`~.Scrolled`
+- :class:`~.Slider`
+- :class:`~.TextFieldStyle`
+- :class:`~.TextField`
+- :class:`~.TextFieldGrid`
```
jwiggins, line 147: Most of these are neglected and of limited usefulness. I would recommend their omission here and schedule them for deprecation.

jwiggins (approved):

PR #519, opened 2020-12-30
Import version from package instead of manually compiling the version file

rahulporuri:
Trying to build the docs on master on a windows machine is broken. Running the `docs` command raises the following error - 

```
Configuration error:
There is a programmable error in your configuration file:

  File "C:\Users\rporuri\.edm\envs\enable-test-3.6-pyqt5\lib\site-packages\sphinx\config.py", line 368, in eval_config_file
    execfile_(filename, namespace)
  File "C:\Users\rporuri\.edm\envs\enable-test-3.6-pyqt5\lib\site-packages\sphinx\util\pycompat.py", line 81, in execfile_
    exec(code, _globals)
  File "C:\Users\rporuri\work\github\ets\enable\docs\source\conf.py", line 51, in <module>
    version = release = runpy.run_path(version_file)['full_version']
  File "C:\Users\rporuri\.edm\envs\enable-test-3.6-pyqt5\lib\runpy.py", line 261, in run_path
    code, fname = _get_code_from_file(run_name, path_name)
  File "C:\Users\rporuri\.edm\envs\enable-test-3.6-pyqt5\lib\runpy.py", line 231, in _get_code_from_file
    with open(fname, "rb") as f:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\rporuri\\work\\github\\ets\\enable\\docs\\source\\..\\..\\enable\\_version.py'
```

This seems to be happening because we try creating the `enable/_version.py` and `kiva/_version.py` in the setup.py file - https://github.com/enthought/enable/blob/5834fb812c4506324897a3bd14bfe6875f5f2998/setup.py#L236-L237 - which doesn't work as expected on windows.

Instead of making more modifications to the already messy setup file, I chose to simply import the version info from the package instead of manually searching and compiling the `_version.py` file for the version info.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -46,9 +46,8 @@ copyright = '2008-2019, Enthought'
 
 # The default replacements for |version| and |release|, also used in various
 # other places throughout the built documents.
-base_path = os.path.dirname(__file__)
-version_file = os.path.join(base_path, '..', '..', 'enable', '_version.py')
-version = release = runpy.run_path(version_file)['full_version']
+from enable import __version__
+version = release = __version__
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:


jwiggins:
> This seems to be happening because we try creating the `enable/_version.py` and `kiva/_version.py` in the setup.py file
> ... which doesn't work as expected on windows.

What doesn't work? Creation of the `_version.py` files?

I think the intention here is to make building the documentation possible from a simpler environment?  We're not (currently) generating anything from the source code with autodoc.

docs/source/conf.py
```diff
@@ -46,9 +46,8 @@
 
 # The default replacements for |version| and |release|, also used in various
 # other places throughout the built documents.
-base_path = os.path.dirname(__file__)
-version_file = os.path.join(base_path, '..', '..', 'enable', '_version.py')
-version = release = runpy.run_path(version_file)['full_version']
+from enable import __version__
+version = release = __version__
```
kitchoi, line 50: I vaguely remembered there is a reason why it was done via executing the code...

The one difference I know (not sure if it is the same as the reason I think I forgot): On master, you did not have enable installed, then building documentation would still "work", but the resulting documentation does not contain any API documentation because it relies on autodoc. Here you must have enable installed to build anything at all.

kitchoi: I tested that without cleaning my local git repository. If I clean the local git repository, then `_version.py` would have been removed, but building the documentation would also fail with the original code.
But the source distribution includes `_version.py`, so it would work there without having to install enable source first. It might have been for that.

kitchoi: If `_version.py` is not available as suggested in the PR description, I am puzzled by how the new change is supposed to work.

kitchoi:
Looking deeper into this, I could reproduce the issue on OSX.

Basically we expect the following commands to work for building the documentation:
```
$ git clone <enable-repo>
$ edm install -e bootstrap click
$ edm run -e bootstrap -- python ci/edmtool.py install
$ edm run -e bootstrap -- python ci/edmtool.py docs
```

In the `install` command, `write_version_py`(in setup.py) is called via `pip install`. When it writes the `_version.py` file, the source has been copied to to temporary directory.  e.g. if I make `write_version_py` raise with the filepath it intends to write to, I get something like this:
```
    ERROR: Command errored out with exit status 1:
     command: /Users/kchoi/.edm/envs/enable-test-3.6-null/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/_6/bw6pnpwx3pzfvhd20ffcpyjm0000gn/T/pip-req-build-n2_nyvnq/setup.py'"'"'; __file__='"'"'/private/var/folders/_6/bw6pnpwx3pzfvhd20ffcpyjm0000gn/T/pip-req-build-n2_nyvnq/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/_6/bw6pnpwx3pzfvhd20ffcpyjm0000gn/T/pip-req-build-n2_nyvnq/pip-egg-info
         cwd: /private/var/folders/_6/bw6pnpwx3pzfvhd20ffcpyjm0000gn/T/pip-req-build-n2_nyvnq/
    Complete output (7 lines):
      File "/private/var/folders/_6/bw6pnpwx3pzfvhd20ffcpyjm0000gn/T/pip-req-build-n2_nyvnq/setup.py", line 247, in <module>
        write_version_py(filename='enable/_version.py')
      File "/private/var/folders/_6/bw6pnpwx3pzfvhd20ffcpyjm0000gn/T/pip-req-build-n2_nyvnq/setup.py", line 106, in write_version_py
        raise ValueError(kiva_version_path)
    ValueError: /private/var/folders/_6/bw6pnpwx3pzfvhd20ffcpyjm0000gn/T/pip-req-build-n2_nyvnq/kiva/_version.py
```

So the `kiva/_version.py` and `enable/_version.py` are written to that temporary directory. The `_version.py` files are then copied to the site-packages as part of the installation step, so the installed packages should have the `_version.py` files.

However, the `conf.py` assumes the `_version.py` file is written in the source root directory, which would be true if it was from the source distribution. But when we run the `docs` command from `edmtool.py`, it isn't the source distribution, and as explained above, the `_version.py` is not there, so `conf.py` fails.

In other ETS projects, like Traits and TraitsUI where the setup.py file has been revamped, the target version file path is computed much early on, so that the `_version.py` would be written to the source directory, instead of the temporary directory mentioned above.

kitchoi (approved):
LGTM.
I still struggle to understand or to extract the reason why the `conf.py` from many years ago try to find and execute the `_version.py` file. Now we are doing auto api documentation generation, it makes sense to assume the package has been installed.

PR #520, opened 2020-12-30
Unset 2to3 flag from setup file

rahulporuri:
Now that we don't have any Python 2 code, we can simply remove the `use_2to3` boolean flag in setup. Note that we werent using 2to3 in the recent release anyway.


jwiggins (approved):
LGTM

PR #522, opened 2020-12-30
Dont document null submodule in api docs

rahulporuri:
This PR adds `enable.null` submodule to the list of submodules that are ignored when generating the api documentation. FTR, the other modules ignored are `enable/qt4`, `enable/wx` and `*/tests`.

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -251,6 +251,7 @@ def docs(runtime, toolkit, environment):
         "enthought_sphinx_theme",
     ])
     ignore = " ".join([
+        "enable/null",
         "enable/qt4",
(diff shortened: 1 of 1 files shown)


ci/edmtool.py
```diff
@@ -251,6 +251,7 @@ def docs(runtime, toolkit, environment):
         "enthought_sphinx_theme",
     ])
     ignore = " ".join([
+        "enable/null",
```
jwiggins, line 254: I would add `enable/pyglet_backend` and `enable/vtk_backend` as well.

jwiggins (approved):

PR #524, opened 2020-12-30
Remove bad copyright headers

rahulporuri:
This PR removes bad copyright headers from the `enable` and `kiva` packages. There are a number of variants of copyright headers in `enable` and `kiva` which aren't right.

The right format for the headers is - 

```
# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
# All rights reserved.
#
# This software is provided without warranty under the terms of the BSD
# license included in LICENSE.txt and may be redistributed only under
# the conditions described in the aforementioned license. The license
# is also available online at http://www.enthought.com/licenses/BSD.txt
#
# Thanks for using Enthought open source!
```

Note that I haven't looked at the tests. Note also that there are a few modules which have copyright headers which refer to Greenbank Computing Inc or the Space Telescope Science Institute. In those cases, I didn't remove the copyright header.

This PR is in preparation for the automated addition of copyright headers to all of the modules in `enable` and `kiva`.


rahulporuri:
> @rahulporuri Could you elaborate on why these copyright headers are "bad"? Are they acknowledging the wrong parties? They contain the wrong years? Or they are bad because they don't conform to the copyright header format we now have?

Most of the above - in some cases it was wrong year ranges, in other cases it was the wrong format.

> Therefore I'd be more in favour of replacing the copyright headers with better ones in a single PR than removing these acknowledgment in one PR and then adding different ones later at an unknown time.

Makes sense. I'll work on updating the headers in this PR itself instead of opening a separate PR.

kitchoi:
Thank you!

PR #527, opened 2020-12-31
Ignore more toolkit-specific modules in apidocs

rahulporuri:
This PR updates the `docs` click command to ignore more modules when generating the api documentation - specifically the toolkit-specific modules 

- `"enable/savage/trait_defs/ui/qt4"`
- `"enable/savage/trait_defs/ui/wx"`
- `"enable/trait_defs/ui/qt4"`
- `"enable/trait_defs/ui/wx"`

Note that we are including the following modules in the api docs even though they feel like they are implementation details - https://github.com/enthought/enable/tree/master/enable/savage/svg/backends


kitchoi:
In the long run I think we want to be able to generate API documentation for toolkit specific module. At the moment we aren't before the documentation config has not set up the mocking required when the toolkit backends are absent during documentation build. 

These skips are therefore really meaning to silent warnings due to the ImportError, is that right?

kitchoi (approved):
LGTM; on the basis that enable has already been skipping documentation for toolkit specific components.

I am wondering if `kiva/trait_defs/ui/wx` needs to receive the same treatment, feel free to add this entry here post-approval.

Not entirely sure why toolkit-specific modules are skipped; but the fact that `null` is also skipped suggests to me that they are deemed implementation details to be hidden.

jwiggins:
> they are deemed implementation details to be hidden.

For the most part, yes. The only useful information about toolkits that a user of enable needs is their name [for toolkit selection]. Everything else should be abstracted away by the `Window` and `GraphicsContext` interfaces.

rahulporuri:
> These skips are therefore really meaning to silent warnings due to the ImportError, is that right?

No, it's because they are implementation detail to be hidden, like you mentioned in a later comment.

> I am wondering if kiva/trait_defs/ui/wx needs to receive the same treatment.

I left this as is because there is no interface for `KivaFontEditor` and there is not qt implementation. So, in this case, the implementation is the only way we can access/understand the relevant interface.

PR #475, opened 2020-12-07
Use unittest as test runner instead of nose

aaronayres35:
fixes #426 

This PR simply updates edmtool to run using unittest instead of nose, and changes test file names to ensure they get discovered.

PR #528, opened 2020-12-31
Rename test module so it gets discovered by unittest

rahulporuri:
This PR renames a test module so that it gets discovered by unittest. This change was missed when we switched from using nose to unittest as the test runner - #475 

On master, there are 295 tests run on a windows machine with 32 tests skipped. On this branch, there are 312 tests run with 32 tests skipped. Note that the tests were run in an environment with pyqt5 i.e.

```
python ci\edmtool.py install --toolkit pyqt5
python ci\edmtool.py test --toolkit pyqt5
```


jwiggins (approved):
Sure

PR #530, opened 2020-12-31
Fix flake8 F401 and F403 errors (import-related)

rahulporuri:
This PR fixes a number of flake8 import-related errors e.g. F401 i.e. unused imports and F403 i.e. * imports.

Note that this PR does not fix all of the F401 violations in the codebase - I chose to ignore a number of violations coming from api modules or init files. Those will be added to the list of files ignored when we have flake8 properly setup for enable.

Additionally, there were a number of imports which cause F401 errors that we might not be able to remove - to maintain backwards compatibility.

### DIFF
--- a/kiva/basecore2d.py
+++ b/kiva/basecore2d.py
@@ -43,7 +43,7 @@ from .constants import (POINT, LINE, LINES, RECT, NO_DASH, CLOSE,
                         SCALE_CTM, TRANSLATE_CTM, ROTATE_CTM, CONCAT_CTM,
                         LOAD_CTM)
 from .abstract_graphics_context import AbstractGraphicsContext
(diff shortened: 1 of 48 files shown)


kiva/basecore2d.py
```diff
@@ -43,7 +43,7 @@
                         SCALE_CTM, TRANSLATE_CTM, ROTATE_CTM, CONCAT_CTM,
                         LOAD_CTM)
 from .abstract_graphics_context import AbstractGraphicsContext
-from .line_state import LineState, line_state_equal
+from .line_state import LineState
```
jwiggins, line 46: This is causing the tests to break

jwiggins (commented):
Some of these changes may be risky because of transitive imports

PR #488, opened 2020-12-14
Remove import side effect of creating font cache in font_manager

kitchoi:
Closes #362

This PR removes the import side effect of building font cache in the `font_manager` module. The action of loading fonts and building the cache is deferred until the global font manager singleton is requested.  All code external to `font_manager` should use the `default_font_manager` callable to access this font manager singleton.

Benefits of this change:
- Make it harder to resurrect the import side-effect in other modules that depend on font_manager (it is too easy to introduce an import and not noticing that brings about additional side effects.)
- Make it easier to test `font_manager` without the import side effect. (The global singleton still makes testing harder.)

It might be easier to review this PR by commits. As an overview:
- Add the `default_font_manager` function and add tests for modules (outside of `font_manager` that uses the `fontManager` singleton
- Refactor code to make removing the import side effects easier
- Remove the import side effect
- Rewrite tests after the import side effect is gone: Its absence makes testing easier

Note:
If downstream code has only ever access the FontManager via `kiva.font.Font.findfont`, then this PR does not change the  behaviour perceived externally because #368 already defers the import side effect to where `kiva.font.Font.findfont` is called.

PR #531, opened 2021-01-04
Fix Font.findfontname AttributeError when the default font manager has not been initialized

kitchoi:
This PR fixes an AttributeError observed in #392. The error is caused by #488 where the import side effect of creating a module level global font manager cache is removed. There still exists usage of the global font manager assuming it had been created.

The first commit adds a test that would fail. The commit f665e53283aefeebc4299d5503933ed450e5c187 fixes the issue (it is by @jwiggins, cherry-picked from #392).

~Marking as draft to get CI run on the first commit in order to verify the test does fail (though I verified locally, I don't trust my setup when the subject is a module level cache), and then I will push the commit that fixes it.~ (Edited: Done)

### DIFF
--- a/kiva/fonttools/font_manager.py
+++ b/kiva/fonttools/font_manager.py
@@ -852,11 +852,11 @@ class FontProperties(object):
         Return the name of the font that best matches the font
         properties.
         """
-        filename = str(fontManager.findfont(self))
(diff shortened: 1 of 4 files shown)


jwiggins (approved):
LGTM

PR #532, opened 2021-01-06
Remove python version checks from c/c++ files

aaronayres35:
Should have been part of #415 

This PR removes all `PY_MAJOR_VERSION` and `PY_VERSION_HEX` checks from c/c++ files as we currently only support python 3.6+


aaronayres35:
I just realized that `_cython_speedups.cpp` is a generated file.  Alas, the changes in the PR may be mostly (entirely?) useless... 😞 
I can simply close the PR if these changes are unnecessary.

aaronayres35:
CI is failing and it is not worth time to investigate where I made a mistake given that these changes seem to be unnecessary.
I'm going to close this

PR #534, opened 2021-01-08
Move to a single setup.py file

### LINKED ISSUE #440: Clean up Enable/Kiva setup.py
Kiva in particular has a highly idiosyncratic usage of `setup.py` with nested `setup.py` files in sub modules, as well as custom distutils commands.

We should clean this up so that there is either a single `setup.py` (preferred), or one for Kiva and one for Enable.

In particular, we need to get rid of these subsidiary setup.py files:

https://github.com/enthought/enable/blob/master/kiva/setup.py
https://github.com/enthought/enable/blob/master/kiva/quartz/setup.py
https://github.com/enthought/enable/blob/master/kiva/agg/setup.py

jwiggins:
Closes #440

This brings all the extension building into the single top-level setup.py script. I'm kinda curious if this will pass CI, since I've only tested it on macOS.

Some changes:
* No more support for 32-bit OSX systems. That ship has sailed.
* No more `ALWAYS_32BIT_WORKAROUND` for 64-bit machines with g++ < 4.0
* The `macport` extension, used for Wx 2.6 support, is not longer built.
* The `--compiler=STR` option for `setup.py develop` has been removed.
* The custom `clean` and `build_py` commands are gone, as is the explicit use of numpy distutils.

NOTE: The `sdist` looks like it might be broken.
NOTE: This conflicts pretty massively with #392


jwiggins:
GitHub is being a jerk. Closing and reopening.

PR #536, opened 2021-01-08
Remove code coverage reporting on PRs

rahulporuri:
This PR removes coverage reporting for Pull Requests. The coverage reports have been visual noise at best in recent times and have not really contributed meaningful information in the PRs.

Apart from these code changes, i don't know if there are other changes that we need to make to completely sever ties with codecov.


kitchoi:
Thank you. Could you remove the coverage badge on README as well given that it won't be providing accurate information?

rahulporuri:
looks like there is a webhook that i need to delete after this PR is merged

codecov-webhook

jwiggins (approved):

PR #535, opened 2021-01-08
Move to a single setup.py file

jwiggins:
Closes #440

This brings all the extension building into the single top-level setup.py script.

Some changes:
* No more support for 32-bit OSX systems. That ship has sailed.
* No more `ALWAYS_32BIT_WORKAROUND` for 64-bit machines with g++ < 4.0
* The `macport` extension, used for Wx 2.6 support, is not longer built.
* The `--compiler=STR` option for `setup.py develop` has been removed.
* The custom `clean` and `build_py` commands are gone, as is the explicit use of numpy distutils.

~NOTE: The `sdist` looks like it might be broken.~
~NOTE: This conflicts pretty massively with #392~

PR #538, opened 2021-01-12
Add OSX job to CI using GitHub Actions

### LINKED ISSUE #537: Add CI job for running tests against MacOS
Currently enable CI does not test against MacOS, which leaves the package vulnerable to breakages that only occur on OSX.

(Motivated by reviewing this PR: #392, which has OSX specific changes)

kitchoi:
Closes #537

Certain components (e.g. kiva) are OSX specific and it would be convenient to developers if the tests are run by CI in an OSX environment, otherwise we rely on developers having an OSX host machine and remembering to run test suite locally.

Background: Other ETS projects are in the middle of moving CI to GitHub Actions (see Traits, TraitsUI, and Traits-Futures).

I also have separate [branch](https://github.com/enthought/enable/compare/ci-github-actions) where I included all the jobs we currently have in Travis/Appveyor. I decided that might be too much to review in one go. The OSX job is trivial enough so it would be more efficient to break this out. Once this gets merged, it can immediately benefit pending PRs such as #392 and #535 which have OSX specific changes.
I am happy to replace this PR with one that includes all the jobs, if the reviewer prefers.

### DIFF
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/.cache
-          key: ${{ runner.os }}--${{ hashFiles('ci/edmtool.py') }}
+          key: ${{ runner.os }}-${{ runner.toolkit }}-${{ hashFiles('ci/edmtool.py') }}
       - name: Setup EDM
         uses: enthought/setup-edm-action@v1
         with:
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -9,18 +9,13 @@ env:
 
 jobs:
 
-  # Test against EDM packages on Windows and OSX
+  # Test against EDM packages
   test-with-edm:
     strategy:
       matrix:
-        os: [macos-latest, windows-latest]
+        os: [macos-latest]
         toolkit: ['pyqt', 'pyqt5', 'pyside2', 'wx']
     runs-on: ${{ matrix.os }}
-    env:
-      # Set root directory for Windows, otherwise 'pip install <source-root>'
-      # complains because the Python environment is in C: drive whereas
-      # the source is checked out in D: drive.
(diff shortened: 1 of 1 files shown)


rahulporuri:
I see the following in the "Post Cache EDM packages" step of the GitHub Action in 3 of the 4 jobs.

> Unable to reserve cache with key macOS--25bb83e821c9e99b32f531926946e70de229dc2f7eb6892e69aed93c305c3b86, another job may be creating this cache.

In the fourth, I see the following

> /usr/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /Users/runner/work/enable/enable --files-from manifest.txt
> Cache saved successfully

Looks like this is happening because of the way we set the `key` when we cache. At the moment, the `key` is the same for all of the four jobs - and they should probably be different.

kitchoi:
Goo point. I added the toolkit to the cache key. Though the last build won't be exercising the cache logic as it would be the first time the workflow has seen the cache key (and hence find no cache).

rahulporuri:
I'm not sure what's going wrong but things still don't seem to be working as expected. This is what I now see for all 4 jobs in the Post Cache step -

> Cache hit occurred on the primary key macOS--25bb83e821c9e99b32f531926946e70de229dc2f7eb6892e69aed93c305c3b86, not saving cache.

kitchoi:
> I'm not sure what's going wrong but things still don't seem to be working as expected. This is what I now see for all 4 jobs in the Post Cache step -

I was puzzled too... but then I realized what went wrong (see this [commit](https://github.com/enthought/enable/pull/538/commits/a3cb3121f599052314c9f58c9701a94b27fb06d8)). It is now fixed.

rahulporuri:
>  I am happy to replace this PR with one that includes all the jobs, if the reviewer prefers.

Let's not. I think it would be easier to review if they're split up into separate PRs.

rahulporuri (approved):
🥳  LGTM

PR #542, opened 2021-01-13
Add Linux job to CI on GitHub Actions

kitchoi:
Another incremental step towards replicating all existing Travis/Appveyor jobs on GitHub Actions.

This PR adds a job to run tests against Linux.

### DIFF
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -22,14 +22,7 @@ jobs:
       - name: Install Qt dependencies for Linux
         run: |
           sudo apt-get update
-          sudo apt-get install qt5-default
           sudo apt-get install libxkbcommon-x11-0
-          sudo apt-get install libxcb-icccm4
-          sudo apt-get install libxcb-image0
-          sudo apt-get install libxcb-keysyms1
-          sudo apt-get install libxcb-randr0
-          sudo apt-get install libxcb-render-util0
-          sudo apt-get install libxcb-xinerama0
         if: matrix.toolkit != 'wx'
       - name: Install Wx dependencies for Linux
         run: |
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         toolkit: ['pyqt', 'pyqt5', 'pyside2', 'wx']
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
       - name: Install Qt dependencies for Linux
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         toolkit: ['pyqt', 'pyqt5', 'pyside2', 'wx']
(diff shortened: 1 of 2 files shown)


.github/workflows/test-with-edm.yml
```diff
@@ -9,9 +11,47 @@ env:
...
+  test-edm-linux:
+    strategy:
+      matrix:
+        toolkit: ['pyqt', 'pyqt5', 'pyside2', 'wx']
+    runs-on: ubuntu-16.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Qt dependencies for Linux
+        run: |
+          sudo apt-get update
```
rahulporuri, line 24: Any specific reason why we are running `... update`? I'm asking because this adds a few extra seconds to the overall time the job takes to finish.

.github/workflows/test-with-edm.yml
```diff
@@ -9,9 +11,47 @@ env:
...
+          sudo apt-get install libxkbcommon-x11-0
+        if: matrix.toolkit != 'wx'
+      - name: Install Wx dependencies for Linux
+        run: |
+          sudo apt-get update
```
rahulporuri, line 29: Not sure why we need to do `... update` a second time.

.github/workflows/test-with-edm.yml
```diff
@@ -9,9 +11,47 @@ env:
...
+      - name: Install click to the default EDM environment
+        run: edm --config ci/.edm.yaml install -y wheel click coverage
+      - name: Install test environment
+        run: edm run -- python ci/edmtool.py install --toolkit=${{ matrix.toolkit }}
+      - name: Run tests
+        uses: GabrielBB/xvfb-action@v1
+        with:
+          # kiva agg requires at least 15-bit color depth.
+          # The --server-args assumes xvfb-run is called, hence Linux only.
+          run: --server-args="-screen 0 1024x768x24" edm run -- python ci/edmtool.py test --toolkit=${{ matrix.toolkit }}
```
rahulporuri, line 52: Just to confirm, `1024x768x24` is `1024x768` resolution and `24` bit color?

rahulporuri (approved):
Mostly LGTM

.github/workflows/test-with-edm.yml
kitchoi: It is included because I believe this is good practice to update the package repository before installing them. If it becomes a bottleneck we may be able to omit it.

.github/workflows/test-with-edm.yml
kitchoi: This is not a second time, because this step and the step above are mutually exclusive.

.github/workflows/test-with-edm.yml
kitchoi: Yep. The number is somewhat arbitrary (apart from the depth part being greater than 15).

.github/workflows/test-with-edm.yml
rahulporuri: ahh. sorry. resolving.

rahulporuri:
still LGTM

.github/workflows/test-with-edm.yml
kitchoi: You made me think that perhaps the `sudo apt-get update` could be extracted to an earlier step - it might look more obvious it is done once only...

Issue #543, opened 2021-01-13
Tests error due to missing optional dependency - pyparsing


kitchoi:
To reproduce:
```
$ pip install .[test]
$ cd
$ python -m unittest discover enable
```
Many test errors occur, this is one of them:
```

======================================================================
ERROR: enable.savage.svg.tests.css.test_atrule (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: enable.savage.svg.tests.css.test_atrule
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/enable/savage/svg/tests/css/test_atrule.py", line 4, in <module>
    from pyparsing import ParseException
ModuleNotFoundError: No module named 'pyparsing'
```

`pyparsing` is declared as an optional dependency in the README. Tests on optional features are expected to be skipped if the optional dependencies are absent.
e.g. GL tests that depend on `pyglet` are skipped if `pyglet` is absent.

PR #544, opened 2021-01-14
Add cron job to GitHub Actions and expand extras_require

kitchoi:
This PR adds a cron job to the GitHub Actions workflow.

Different from the cron job currently setup on Travis which uses EDM, this cron job is setup to install everything from `pip`. 

This exercise also helps flush out missing dependency declaration in `setup.py`, which isn't used by `ci/edmtool.py`. With that, the `extras_require` declarations for optional dependencies is expanded here so that CI setup can call `pip install .[flag,...]`.  This will further benefit package distributors who will be looking for these declarations when they build enable.

As the names of the optional dependencies are tied to the cron job setup, I keep these two changes in the same PR.

Since tests on optional features are skipped if the optional dependency is absent (apart from the ones on SVG, see #543), I checked the CI job output to look out for unexpected skips.

Note to reviewer
--- 
The current trigger is `on: push`, for testing purposes. Once the reviewer is happy with the change, I will switch the event trigger to:
```
on:
  schedule:
    - cron:  '0 0 * * 5'
```
and request one more review again before merging.

### DIFF
--- a/.github/workflows/bleeding-edge.yml
+++ b/.github/workflows/bleeding-edge.yml
@@ -22,10 +22,17 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install Qt dependencies for Linux
+      - name: Install PySide2 (from PyPI) dependencies for Linux
         run: |
           sudo apt-get update
+          sudo apt-get install qt5-default
           sudo apt-get install libxkbcommon-x11-0
+          sudo apt-get install libxcb-icccm4
+          sudo apt-get install libxcb-image0
+          sudo apt-get install libxcb-keysyms1
+          sudo apt-get install libxcb-randr0
+          sudo apt-get install libxcb-render-util0
+          sudo apt-get install libxcb-xinerama0
         if: matrix.toolkit != 'wx'
       - name: Install Wx dependencies for Linux
         run: |
--- a/.github/workflows/bleeding-edge.yml
+++ b/.github/workflows/bleeding-edge.yml
@@ -1,0 +1,62 @@
+# This workflow installs dependencies from master
+
+name: ETS from source
+
+# TODO: Switch to schedule!
+on: push
+
+jobs:
+  test-ets:
+    strategy:
+      matrix:
+        # Note that Ubuntu 20 cannot be used because its SWIG version is 4.0
+        # (enthought/enable#360)
+        os: [ubuntu-18.04]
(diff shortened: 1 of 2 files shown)


.github/workflows/bleeding-edge.yml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  test-ets:
+    strategy:
+      matrix:
+        # Note that Ubuntu 20 cannot be used because its SWIG version is 4.0
+        # (enthought/enable#360)
+        os: [ubuntu-18.04]
+        toolkit: ['pyside2', 'wx']
+        python-version: [3.8]
```
kitchoi, line 16: I tried the combination of Python 3.9 and Ubuntu 20, but there aren't a prebuilt wheel for wxPython, which means we will have to build from source (take a long time!). But then I haven't tried bumping Python version again for Ubuntu 18.

I did not use Ubuntu 16, as an attempt to be as close to the latest as possible, and also to be different from the per-PR CI job (using Ubuntu 16) to expand coverage.

.github/workflows/bleeding-edge.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install qt5-default
+          sudo apt-get install libxkbcommon-x11-0
+          sudo apt-get install libxcb-icccm4
+          sudo apt-get install libxcb-image0
+          sudo apt-get install libxcb-keysyms1
+          sudo apt-get install libxcb-randr0
+          sudo apt-get install libxcb-render-util0
+          sudo apt-get install libxcb-xinerama0
```
rahulporuri, line 35: Are all of these needed?

rahulporuri (approved):
LGTM with one question.

.github/workflows/bleeding-edge.yml
kitchoi: Yes, because this workflow uses PySide2 from PyPI. See this build where I have only included the linux packages from the other workflow where PySide2 from EDM is used: https://github.com/enthought/enable/runs/1697094819

rahulporuri:
still LGTM

kitchoi:
Thank you @rahulporuri 
As promised, the `on` trigger is updated to 'schedule', does this still look okay to merge?

rahulporuri:
still LGTM

kitchoi:
Thank you!

PR #546, opened 2021-01-14
Mention in the README libglu1-mesa-dev is a build dependency on Linux

kitchoi:
From the exercise of setting up CI, it is evident that `libglu1-mesa-dev` is a build dependency on Linux, `pip install ` fails with lots of compilation errors if the package is absent.

This PR adds the required `libglu1-mesa-dev` (as is named on Ubuntu) on the README. The package name may vary slightly from linux distribution to distribution, though.


rahulporuri (approved):
LGTM

PR #547, opened 2021-01-14
Declare setuptools as a test dependency

### LINKED ISSUE #545: setuptools is a test dependency and should be declared
Currently there exist tests that depend on `pkg_resources`. Unless those tests can be changed, we should add `setuptools` to the test dependencies:

https://github.com/enthought/enable/blob/6b4ad53f2eaf3257b2a95ee1f0e5fd45624aee58/enable/__init__.py#L30

kitchoi:
Closes #545

`pkg_resources` is used in tests. One will get import error if `setuptools` is absent (hard to replicate in CI because `setuptool` often exists). Strictly speaking, one cannot assume a Python environment must have `setuptools`.

### DIFF
--- a/enable/__init__.py
+++ b/enable/__init__.py
@@ -49,5 +49,6 @@ __extras_require__ = {
     'test': [
         'hypothesis',
         'PyPDF2',    # for pdf drawing tests in kiva.
+        "setuptools",
     ]
 }


rahulporuri (approved):
LGTM

PR #556, opened 2021-01-29
Link against system FreeType

### LINKED ISSUE #553: Remove vendored FreeType from Kiva backend(s)
Kiva's Agg backend includes FreeType 2.3.12 as a vendored dependency. This version is ~11 years old at this point.

Given the issues we've already had with symbol clashes from the vendored Agg code, it's probably a good idea to link FreeType as a shared library. This is already being done by EDM for `matplotlib`, so we could even use the same version.

#74 might be related

jwiggins:
Closes #553

This is just the change to the build script. Another PR will remove the vendored code since that's a very big change.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -137,26 +137,47 @@ if not is_released:
                                  is_released=IS_RELEASED))
 
 
+def get_freetype_info():
+    """ Use pkg-config to locate the freetype2 library installed on the system.
+
+    If it's not installed at the system level, attempt to find it in the
+    pkgconfig directory for the currently running Python interpreter.
+    """
+    def run_cmd(cmd, env=None):
+        output = subprocess.check_output(cmd, universal_newlines=True,
+                                         env=env, stderr=subprocess.DEVNULL)
+        return output.split()
+
+    def collect_data(env=None):
+        cmd_prefix = ['pkg-config', 'freetype2']
+        commands = {'cflags': ['--cflags'], 'ldflags': ['--libs']}
+
+        data = {}
+        try:
+            for key, args in commands.items():
+                data[key] = run_cmd(cmd_prefix + args, env=env)
+        except (subprocess.CalledProcessError, OSError):
+            pass
+        return data
+
+    data = collect_data()
+    if len(data) < 2:
+        # Try again with the Python env's pkgconfig directory added
+        env = os.environ.copy()
+        env['PKG_CONFIG_PATH'] = os.path.join(
+            sys.exec_prefix, 'lib', 'pkgconfig'
+        )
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -215,6 +226,25 @@ def agg_extensions():
...
 
+    # Figure out the freetype linkage
+    if sys.platform == 'win32':
+        # XXX: Note that Windows has used FreeType for many years. This would
+        # be a change in behavior.
+        define_macros += [('KIVA_USE_WIN32', None)]
+        include_dirs += [os.path.join(agg_dir, 'font_win32_tt')]
+        agg_sources += [
+            os.path.join(agg_dir, 'font_win32_tt', 'agg_font_win32_tt.cpp'),
+        ]
```
jwiggins, line 237: It's important to check that this doesn't change things too much on Windows. If it does, we'll have to figure out how to link against an external FreeType on Windows.

setup.py
```diff
@@ -215,6 +226,25 @@ def agg_extensions():
...
 
+    # Figure out the freetype linkage
+    if sys.platform == 'win32':
+        # XXX: Note that Windows has used FreeType for many years. This would
+        # be a change in behavior.
+        define_macros += [('KIVA_USE_WIN32', None)]
+        agg_include_dirs += [os.path.join(agg_dir, 'font_win32_tt')]
+        kiva_agg_sources += [
+            os.path.join(agg_dir, 'font_win32_tt', 'agg_font_win32_tt.cpp'),
+        ]
```
jwiggins, line 237: It's important to check that this doesn't change things too much on Windows. If it does, we'll have to figure out how to link against an external FreeType on Windows.

rahulporuri:
Looks like this is no deal on Windows. Apart from CI, I just tried testing this locally and I saw the same issue.

jwiggins:
The problem was that `KIVA_USE_FREETYPE` was being forcibly defined in the code. I've fixed that now and it _should_ build

jwiggins:
OK. That took a few more tries than expected, but I've now shaken out the code rot in the `KIVA_USE_WIN32` code paths. CI should pass for real this time.

setup.py
jwiggins: Using the Win32 font rendering appears to break text rotation, according to @rahulporuri's testing

jwiggins:
Using the Win32 font renderer in Agg has bug related to affine transformations, so it's best to stick with the FreeType renderer. Unfortunately, that means a lot more complexity when compiling on Windows if we don't want to include a vendored copy of FreeType.

I'm going to close this for now and we can revisit when we're not trying to get a release out the door.

PR #558, opened 2021-02-01
Run black on kiva

jwiggins:
Same thing as #557 but with kiva

PR #559, opened 2021-02-02
Add tox.ini for flake8 control

jwiggins:
We're not ready to run flake8 on the code regularly yet, but with #557 and #558 in place we can now work towards that goal. To that end, I'm introducing a `tox.ini` config which adds ignores for machine generated and vendored files that we don't care to see linter errors for.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,14 @@
+[flake8]
+application-package-names = pyface,traits,traitsui
+application-import-names = enable,kiva
+import-order-style = appnexus
+per-file-ignores =
+    # SWIG generated
+    kiva/agg/agg.py: E,F,I,W
+    kiva/gl/gl.py: E,F,I,W
+    # Vendored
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -0,0 +1,13 @@
+[flake8]
+application-package-names = enable,kiva,pyface,traits,traitsui
```
rahulporuri, line 2: do we want to add the `application-import-names` and put `enable` and `kiva` under there instead? -> This is bound to raise a lot of new flake8 errors though.

```suggestion
application-package-names = pyface,traits,traitsui
application-import-names = enable,kiva
```

rahulporuri (approved):
LGTM

tox.ini
jwiggins: Good catch

rahulporuri (approved):
Still LGTM

jwiggins:
Thanks for the feedback

PR #560, opened 2021-02-02
Remove outdated/broken debugging constructs

rahulporuri:
This PR removes the following : 

- remove the old style `print` statements and a few new style print functions
- remove `if __name__ == "__main__"` constructs in the test modules
- remove pdb uses


jwiggins (approved):
Yes. LGTM

PR #562, opened 2021-02-02
Unconditionally list pillow as a runtime dependency

rahulporuri:
This PR lists `pillow` as a dependency unconditionally instead of conditionally adding it to the list of required packages depending on whether or not `PIL` can be imported.


jwiggins (approved):
Wow. The last release of PIL was in 2006 (according to PyPI). LGTM.

PR #563, opened 2021-02-03
Provide the right path to the version file on windows

rahulporuri:
On the main branch, the `_version.py` files don't get generated in `enable` and `kiva` - because of the path provided to `write_version_py` function. The changes made here enable the generation of the version file - which also fixes an issue where documentation build is broken on windows because of the lack of the version file.

This is the right fix for the problem I was trying to solve in #519 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -405,8 +405,13 @@ def macos_extensions():
 
 
 if __name__ == "__main__":
-    write_version_py(filename='enable/_version.py')
-    write_version_py(filename='kiva/_version.py')
+    if sys.platform.startswith("win"):
+        write_version_py(filename='enable\\_version.py')
+        write_version_py(filename='kiva\\_version.py')
+    else:
+        write_version_py(filename='enable/_version.py')
+        write_version_py(filename='kiva/_version.py')
+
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -405,8 +405,13 @@ def macos_extensions():
...
 
 if __name__ == "__main__":
-    write_version_py(filename='enable/_version.py')
-    write_version_py(filename='kiva/_version.py')
+    if sys.platform.startswith("win"):
+        write_version_py(filename='enable\\_version.py')
+        write_version_py(filename='kiva\\_version.py')
+    else:
+        write_version_py(filename='enable/_version.py')
+        write_version_py(filename='kiva/_version.py')
```
jwiggins, line 413: Can't you just write:
```python
write_version_py(filename=os.path.join('enable', '_version.py'))
write_version_py(filename=os.path.join('kiva', '_version.py'))
```
?

jwiggins (approved):
A question

setup.py
rahulporuri: duh. of course. that works. fixed.

PR #564, opened 2021-02-03
Add module docstrings to api modules in enable

rahulporuri:
This PR adds module docstrings to the api modules in enable - which improves the readability of the published API documentation.


jwiggins (approved):
LGTM

PR #565, opened 2021-02-03
Use kiva api modules where possible

rahulporuri:
This PR updates the enable and kiva testsuite to use `kiva.api` or `kiva.trait_defs.api` where possible - instead of explicitly importing from `kiva.affine`, `kiva.constants`, `kiva.fonttools` or `kiva.trait_defs.kiva_font_trait`


jwiggins (approved):
LGTM

PR #566, opened 2021-02-03
Use traits.api, traitsui.api and pyface.api where possible.

rahulporuri:
This PR updates the imports from ets package to import from their respective api modules instead of importing from their internals, where possible.


jwiggins (approved):
LGTM

PR #568, opened 2021-02-04
Remove is_string_like

### LINKED ISSUE #394: Unreachable code in kiva.font_manager.is_string_like
The lines
https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/kiva/fonttools/font_manager.py#L221-L225
are unreachable due to an indentation bug since they  most likely should be reachable.

The if statement 
https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/kiva/fonttools/font_manager.py#L217-L220

can be simplified.  The if statement can be replaced with 
```pyhton
return bool(obj.ndim == 0 and obj.dtype.kind in 'SU')
```

jwiggins:
Closes #394

This PR removes `is_string_like`, because it's not needed. It was being used in two places: `FontProperties.__init__` and `FontProperties.set_family`. The usage in `FontProperties.__init__` was removed entirely, because it was guarding a call to `self.set_fontconfig_pattern` which, as far as I can tell, was never copied into the kiva code from matplotlib (see 1fe9c35c1f28e8b8e23c8a3614dc04e606227dc5. Just look at it.).

### DIFF
--- a/kiva/fonttools/font_manager.py
+++ b/kiva/fonttools/font_manager.py
@@ -48,6 +48,7 @@ import tempfile
 import warnings
 
 from fontTools.ttLib import TTCollection, TTFont, TTLibError
+
 from traits.etsconfig.api import ETSConfig
 
 from . import afm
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM

PR #569, opened 2021-02-04
Remove the kiva.compat PIL/Pillow wrapper

### LINKED ISSUE #271: Ensure support for Pillow 3.x and 4.x
A number of methods have been deprecated/renamed and changed behaviour since 2.x.x. So It is highly probable that kiva will not work with Pillow >= 3.x

for example:
- https://pillow.readthedocs.io/en/4.1.x/releasenotes/3.0.0.html#deprecated-methods
- https://pillow.readthedocs.io/en/4.1.x/releasenotes/3.4.0.html#image-core-open-ppm-removed
- https://pillow.readthedocs.io/en/4.1.x/releasenotes/4.0.0.html#image-core-stretch-removed
- https://pillow.readthedocs.io/en/4.1.x/releasenotes/4.1.0.html#removed-deprecated-items
- https://pillow.readthedocs.io/en/4.1.x/releasenotes/4.1.0.html#closing-files-when-opening-images

### LINKED ISSUE #512: Drop support for PIL and old Pillow versions
This will be reflected three ways:

- we can remove the PIL compatibility code in `kiva/compat.py` to do with string handling, and just use the most recent API directly
- we can drop the PIL/pillow versioning in CI
- we can drop tests which use old versions of pillow if we haven't already

jwiggins:
Closes #512
Closes #271


rahulporuri (approved):
Code changes make sense to me. Not sure if you want a second, more knowledgable/experienced pair of eyes to take a look at this PR.

enable/savage/svg/document.py
```diff
@@ -203,17 +203,16 @@ def open_image(self, path):
...
         """
-        path, open = self.resolve(path)
-        fin = open(path)
-
-        from PIL import Image
-        from kiva.compat import piltostring
         import numpy
+        from PIL import Image
+
+        path, open = self.resolve(path)
```
jwiggins, line 209: `resolve` returns either `lambda x: BytesIO(...)` or `lambda x: open(...)` for the `open` here.

kiva/agg/src/graphics_context.i
```diff
@@ -848,8 +850,7 @@ namespace kiva {
                 else:
                     bmp = self.bmp_array
 
-                from kiva import compat
-                img = compat.pilfromstring(pilformat, size, bmp.tostring())
+                img = Image.frombytes(pilformat, size, bmp.tobytes())
```
jwiggins, line 853: `bmp` is a numpy array. `tobytes` is the new, more correct, name for `tostring` (which was kept for compatibility with old code like this but is now deprecated in recent versions of numpy)

kiva/agg/src/graphics_context.i
```diff
@@ -907,15 +908,15 @@ class Image(GraphicsContextArray):
...
+
         pil_img = PilImage.open(file)
 
         # Convert image to a numeric array
         if (pil_img.mode not in ["RGB","RGBA"] or
             (cvar.ALWAYS_32BIT_WORKAROUND_FLAG and pil_img.mode != "RGBA")):
             pil_img = pil_img.convert(mode="RGBA")
         depth = pil_depth_map[pil_img.mode]
-        img = frombuffer(piltostring(pil_img),uint8)
+        img = frombuffer(pil_img.tobytes(), uint8)
```
jwiggins, line 919: I _think_ we can just convert directly to an array here: `numpy.array(pil_img)`.

kiva/celiagg.py
```diff
@@ -818,7 +818,7 @@ def save(self, filename, file_format=None):
         data[..., 2] = pixels[..., 0]
         data[..., 3] = pixels[..., 3]
         size = (int(self._width), int(self._height))
-        img = pilfromstring('RGBA', size, data)
+        img = Image.frombytes('RGBA', size, data)
```
jwiggins, line 821: `Image.fromarray` might be more appropriate here.

kiva/pdf.py
```diff
@@ -600,10 +600,10 @@ def draw_image(self, img, rect=None):
             return
 
         # converted_img now holds an Agg graphics context with the image
-        pil_img = pilfromstring(
+        pil_img = Image.frombytes(
             format,
             (converted_img.width(), converted_img.height()),
-            piltostring(converted_img.bmp_array),
+            converted_img.bmp_array.tobytes(),
```
jwiggins, line 606: `Image.fromarray` might be more appropriate here.

kiva/ps.py
```diff
@@ -215,10 +215,10 @@ def device_draw_image(self, img, rect):
             return
 
         # converted_img now holds an Agg graphics context with the image
-        pil_img = pilfromstring(
+        pil_img = PilImage.frombytes(
```
jwiggins, line 218: ditto

kiva/quartz/ABCGI.pyx
```diff
@@ -1591,7 +1591,7 @@ cdef class CGBitmapContext(CGContext):
         if file_format is None:
             file_format = ''
 
-        img = pilfromstring(mode, (self.width(), self.height()), self)
+        img = PilImage.frombytes(mode, (self.width(), self.height()), self)
```
jwiggins, line 1594: This is interesting. `Image.frombytes` is necessary here because this class doesn't hold it's data in a numpy array, but _does_ support the [legacy] [buffer interface](https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#buffer-interface-legacy-no-python-equivalents-see-note-1). That said, `Image.frombuffer` should also work, and it would avoid a copy.

kiva/quartz/ABCGI.pyx
```diff
@@ -1791,10 +1790,10 @@ cdef class CGImageFile(CGImage):
         self.bmp_array = c_numpy.PyArray_SimpleNew(3, &(dims[0]), c_numpy.NPY_UBYTE)
 
         data = self.bmp_array.data
-        s = piltostring(img)
-        py_data = PyBytes_AsString(s)
+        bs = img.tobytes()
+        py_data = PyBytes_AsString(bs)
 
-        memcpy(<void*>data, <void*>py_data, len(s))
+        memcpy(<void*>data, <void*>py_data, len(bs))
```
jwiggins, line 1796: It might be better not to touch this code even though there is an array object involved.

kiva/svg.py
```diff
@@ -272,10 +272,10 @@ def device_draw_image(self, img, rect):
             return
 
         # converted_img now holds an Agg graphics context with the image
-        pil_img = pilfromstring(
+        pil_img = PilImage.frombytes(
```
jwiggins, line 275: `Image.fromarray`?

kiva/tests/agg/test_draw_dash.py
```diff
@@ -12,8 +13,7 @@ def save(img, file_name):
         size = (img.bmp_array.shape[1], img.bmp_array.shape[0])
         bgr = img.bmp_array[:, :, :3]
         rgb = bgr[:, :, ::-1].copy()
-        st = rgb.tostring()
-        pil_img = pilfromstring("RGB", size, st)
+        pil_img = Image.frombytes("RGB", size, rgb.tobytes())
```
jwiggins, line 16: `Image.fromarray`?

kiva/tests/agg/test_image.py
```diff
@@ -30,8 +29,7 @@ def save(img, file_name):
         size = (img.bmp_array.shape[1], img.bmp_array.shape[0])
         bgr = img.bmp_array[:, :, :3]
         rgb = bgr[:, :, ::-1].copy()
-        st = rgb.tostring()
-        pil_img = pilfromstring("RGB", size, st)
+        pil_img = Image.frombytes("RGB", size, rgb.tobytes())
```
jwiggins, line 32: `Image.fromarray`?

kiva/tests/agg/test_image.py
```diff
@@ -54,7 +52,7 @@ def test_name():
 def sun(interpolation_scheme="simple"):
     path = os.path.join(os.path.dirname(__file__), "doubleprom_soho_full.jpg")
     pil_img = Image.open(path)
-    img = frombuffer(piltostring(pil_img), UInt8)
+    img = frombuffer(pil_img.tobytes(), UInt8)
     img.resize((pil_img.size[1], pil_img.size[0], 3))
```
jwiggins, line 56: Direct cast to numpy array instead?

kiva/tests/test_graphics_context.py
```diff
@@ -129,7 +128,7 @@ def test_rect_scale_translate(self):
 
     def sun(self, interpolation_scheme="simple"):
         pil_img = PILImage.open("doubleprom_soho_full.jpg")
-        img = frombuffer(piltostring(pil_img), UInt8)
+        img = frombuffer(pil_img.tobytes(), UInt8)
         img.resize((pil_img.size[1], pil_img.size[0], 3))
```
jwiggins, line 132: Direct cast to array?

jwiggins (commented):
We could probably still to a bit to modernize this code.

jwiggins:
OK. I've attempted to modernize the code a bit. We'll see what CI thinks, because I didn't test locally. It will probably pass though, _because `draw_image` is not covered by the kiva drawing tests_. We should probably fix that. I encountered some code that was definitely broken during the migration to Python3 (string/bytes confusion).

rahulporuri (approved):
Still LGTM

jwiggins:
This will have better test coverage once it's combined with #572

PR #570, opened 2021-02-04
Remove the old and outdated Pyglet and VTK backends

jwiggins:
Both backends have been mostly inaccessible since the refactor in d3aa65bb8e14d2224ed78ef80c578a4b0f00b087 (just a short 10 years ago)


rahulporuri (approved):
The changes LGTM but i'm not entirely sure what you mean by "Both backends have been mostly inaccessible since the refactor in d3aa65b (just a short 10 years ago)".

I searched for uses of the pyglet and vtk backends on github and i couldn't find any.

It might be better to have this approved by a higher authority (maybe @corranwebster , @mdickinson , @rkern )?

jwiggins:
Yes. I'd prefer to hear the opinion of them before merging this.

rkern:
I think these backends can go. The projects that did use them are long-dormant. With the macOS deprecation of OpenGL, I think we'll probably need to start from scratch if we want this kind of functionality in the future.

jwiggins:
Thanks for the feedback

PR #571, opened 2021-02-04
Pyface-style toolkit selection

### LINKED ISSUE #439: Ensure "qt" works as an alias for the "qt4" toolkit name
In Pyface and TraitsUI you can now use "qt" as an alias for "qt4".  We want to ensure that this works for Enable (and possibly Kiva).

This may be significant work; in particular, it may require a redesign of Enable's toolkit and Kiva backend selection, possibly to sit on top of the same `pyface.base_toolkit` code that Pyface and TraitsUI use, as well as allowing expansion via extension points.

jwiggins:
Closes #439

I _think_ this should work, but I'm not entirely sure if it's the right approach. It makes enable directly dependent on `pyface`, which I think we're OK with now, but if that's not OK then it can be rewritten.

@corranwebster I'm really curious what your opinion is on this.

### DIFF
--- a/enable/null/toolkit.py
+++ b/enable/null/toolkit.py
@@ -1,0 +1,23 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+""" Initialize this backend.
+"""
+from pyface.base_toolkit import Toolkit
+from traits.etsconfig.api import ETSConfig
+
+
+def _wrapper(func):
+    def wrapped(name):
+        # Prefix object lookups with the name of the configured kiva backend.
+        return func(f'{ETSConfig.kiva_backend}:{name}')
+    return wrapped
+
+
+toolkit = _wrapper(Toolkit("enable", "null", "enable.null"))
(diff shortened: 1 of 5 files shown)


corranwebster (approved):
This looks fine to me, but it took a moment for me to work out the wrapper code.

I think there is an alternative approach which is to write a different version of the Pyface `find_toolkit` function which takes into account the Kiva backend as well as the main toolkit.  That would avoid the need for wrappers, but you'd then need to have a `toolkit` object in every module of the form `enable.<pyface_toolkit>.<kiva_backend>`, but it would just be a `BaseToolkit` instance (something like `BaseToolkit("enable", "qt.image", "enable.qt4.image")`).  But that is also clearly more a bit more work config (more entrypoints, more pieces of code to touch).  Probably the biggest advantage of this approach would be that a 3rd party could develop a new Enable and Kiva backend in their own library and exposing it via an entrypoint (eg. `qt.skia` or similar)

There's nothing stopping us moving to that mode later, however, and reducing code duplication is good, so feel free to merge this as-is.

jwiggins:
Giving the ability to provide kiva backends from external code sounds nice [aside from the explosion of entrypoints]. Are we painting ourselves into a corner with this change, or can it be adapted to the more general implementation later?

corranwebster:
> Are we painting ourselves into a corner with this change, or can it be adapted to the more general implementation later?

I don't think we're painting ourselves into a corner, because the API of the toolkit object will remain unchanged (you call it with a string to get the toolkit version of the object), so we could just add additional entrypoints with different names and targets (eg. 'qt.image' or 'qt_image' or something) and then deprecate the ones from this PR (there might be some weird corner cases where they might break, but until we have an actual external implementation everything should be fine).

jwiggins:
Thanks for the reviews and feedback.

PR #573, opened 2021-02-05
Clean up some of the agg tests

jwiggins:
I was getting a bit sick of images getting dumped into the current directory after each `kiva` test run, so this PR does away with that. These old tests are of dubious use, but I'm leaving them in because they at least exercise some code paths.


kiva/tests/agg/test_image.py
```diff
@@ -211,7 +201,7 @@ def bench(stmt="pass", setup="pass", repeat=5, adjust_runs=True):
         for i in range(100):
             number = 10 ** i
             time = timer.timeit(number)
-            if time > 0.2:
+            if time > 0.02:
```
jwiggins, line 204: Also these tests are now 10x faster

rahulporuri (approved):
LGTM. I'm not sure why `test_image3` existed in the first place but it does look like there is no useful information that can be extracted from there.

jwiggins:
`test_image3` is part of the old integration tests collection that was moved into `kiva` several years ago.

PR #575, opened 2021-02-05
update setup.py to allow release candidates

### LINKED ISSUE #435: Update setup.py so we can make release candidates
See similar issue https://github.com/enthought/envisage/issues/316

At the moment, the way `setup.py` is written and version computation in the setup script works, we cannot make release candidates for `enable` e.g. `5.0.0rc1`. It would be useful to bring over the changes from traits/traitsui/pyface/envisage to be able to make release candidates for enable.

aaronayres35:
closes #435 

This PR updates setup.py so we can make release candidates

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,9 +23,12 @@ from setuptools.command.install import install as base_install
 MAJOR = 4
 MINOR = 8
 MICRO = 1
-
+PRERELEASE = ""
 IS_RELEASED = False
-VERSION = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
+
+# Templates for version strings.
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -52,7 +55,7 @@ def read_version_py(path):
         code = compile(fp.read(), 'kiva._version', 'exec')
     context = {}
     exec(code, context)
-    return context['git_revision'], context['full_version']
+    return context['git_revision'], context['version']
```
aaronayres35, line 58: this and some similar changes are unnecessary because I ended up keeping `full_version` even though it is the same as `version` now.

jwiggins (approved):
LGTM

rahulporuri (approved):
Code changes LGTM and this worked as expected locally.

PR #578, opened 2021-02-08
Remove files that are not used or needed

jwiggins:
I noticed these while I was changing all the copyright headers


kiva/agg/build.bat
```diff
@@ -1,7 +0,0 @@
-rmdir /s /q build\temp.win32-2.4\Release
```
rahulporuri, line 1: I guess we don't need this file because we don't have a separate kiva.agg setup.py file?

rahulporuri (approved):
LGTM

kiva/agg/build.bat
jwiggins: Yes. All building is handled by the setup.py script.

PR #586, opened 2021-02-15
Undo removal of get_mac_context import

jwiggins:
I removed this import in error as part of #558. It's needed by the Qt Quartz backend (`enable.qt4.quartz`).

@rahulporuri, @aaronayres35: This should be cherry-picked into the 5.0.0 branch.

### DIFF
--- a/kiva/quartz/__init__.py
+++ b/kiva/quartz/__init__.py
@@ -10,6 +10,8 @@
 import sys
 
 if sys.platform == "darwin":
+    from kiva.quartz.mac_context import get_mac_context  # noqa: F401
+
     def get_macport(dc):
         """ Returns the Port or the CGContext of a wxDC (or child class)
         instance.


rahulporuri (approved):
Changes LGTM but i'm not entirely sure why we can't just update the import to use `kiva.quartz.mac_context` in `enable`. I guess the import isnt just used in enable but in other places as well e.g. in user code?

Also, looking at `enable.qt4.quartz`, the import is `from kiva.quartz import get_mac_context, ABCGI` and im not entirely sure where `ABCGI` comes from?

jwiggins:
`ABCGI` is a module in the `kiva.quartz` package, so that's how that import works.

I also briefly thought about changing the import in `enable.qt4.quartz`, but this fix restores behavior that might be depended on by external code, so I think it's best this way.

rahulporuri:
> ABCGI is a module in the kiva.quartz package, so that's how that import works.

Ahh. Makes sense.

> ... but this fix restores behavior that might be depended on by external code, so I think it's best this way.

Yeah. I agree.

Issue #574, opened 2021-02-05
Fix QPainter backend image drawing


jwiggins:
The test `kiva.tests.test_qpainter_drawing.TestQPainterDrawing.test_image` is currently disabled because there appears to be some color confusion in the QPainter kiva backend. The test draws an image containing a red square and checks the pixels drawn to see if any are red. However no pixels are red because the QPainter code draws a blue square instead. When a numpy array is passed to `draw_image`, it is assumed to be in RGB order (`QImage.Format_RGB888` or `QImage.Format_RGB32`) but ends up getting drawn as if it were in BGR order. 

https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qtgui/qimage.html#qimage-image-formats
https://www.riverbankcomputing.com/static/Docs/PyQt4/qimage.html#Format-enum

PR #589, opened 2021-02-15
Add an Image Explorer demo

jwiggins:
This is just a tiny demo which can be used to exercise the `draw_image` methods of various kiva backends. The next PR will introduce PIL `Image` support across all kiva backends and a fix for #574.

### DIFF
--- a/kiva/examples/kiva/image_explorer.py
+++ b/kiva/examples/kiva/image_explorer.py
@@ -1,0 +1,98 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
The `RGB` and `RGBA` modes work as expected on qt4, qt5 and wx but the `L` and `P` modes don't - which I think is expected (and the error which will be fixed in the future PR).

rahulporuri:
I'm guessing you want this backported to 5.0.0 @jwiggins ? If not this PR, the next one?

jwiggins:
> I'm guessing you want this backported to 5.0.0?

We can discuss it, if you like.

PR #590, opened 2021-02-16
Accept a PIL Image as input to draw_image()

### LINKED ISSUE #574: Fix QPainter backend image drawing

jwiggins:
This adds `Image` from the Python Imaging Library as an acceptable argument for `draw_image` on all relevant kiva backends. Further, it uses `Image` internally to handle conversion of color modes to something which is compatible with the color mode of the graphics context.

Fixes #574

### DIFF
--- a/kiva/cairo.py
+++ b/kiva/cairo.py
@@ -867,6 +867,18 @@ class GraphicsContext(basecore2d.GraphicsContextBase):
             )
         elif isinstance(img, Image.Image):
             converted_img = img.convert("RGBA")
+            img_width, img_height = img.width, img.height
+            img_surface = cairo.ImageSurface.create_for_data(
+                converted_img.flatten(),
+                cairo.FORMAT_RGB24,
+                img_width,
+                img_height,
+            )
+        elif hasattr(img, "bmp_array"):
+            # a kiva agg context
+            if hasattr(img, "convert_pixel_format"):
+                img = img.convert_pixel_format("rgba32", inplace=0)
+            converted_img = Image.fromarray(img.bmp_array)
             flipped_array = numpy.flipud(numpy.array(converted_img))
             img_width, img_height = img.width, img.height
             img_surface = cairo.ImageSurface.create_for_data(
--- a/kiva/cairo.py
+++ b/kiva/cairo.py
@@ -850,13 +850,10 @@ class GraphicsContext(basecore2d.GraphicsContextBase):
 
     def draw_image(self, img, rect=None):
         """
-        img is either a N*M*3 or N*M*4 numpy array, or a Kiva image
-
-        rect - what is this? assume it's a tuple (x,y, w, h)
-        Only works with numpy arrays. What is a "Kiva Image" anyway?
-        Not Yet Tested.
+        `img` is either a N*M*3 or N*M*4 numpy array, or a PIL Image
+        `rect` is a tuple (x, y, w, h)
         """
-        from kiva import agg
+        from PIL import Image
 
         if isinstance(img, numpy.ndarray):
             # Numeric array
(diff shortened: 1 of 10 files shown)


jwiggins:
Damn. Looks like `PIL.ImageQt` doesn't work the same on Pyside2...

corranwebster:
One thing to be careful about - I think Enable uses the agg backend for rendering backbuffers no matter what the Kiva library being used is and then (I suspect) uses `draw_image` to do the blitting.  Removing the ability of the draw_image to accept agg images may break this.

This is the relevant code, I think:
https://github.com/enthought/enable/blob/8bc13b1709a7c32b7961ee11b3e2e903af9f70ee/enable/component.py#L766-L805

jwiggins:
> Removing the ability of the `draw_image` to accept agg images may break this.

The `hasattr(img, "bmp_array"):` checks are basically equivalent to checking `isinstance(img, agg.GraphicsContextArray)`. Additionally, most backends can accept an instance of themselves as the image argument.

kiva/cairo.py
```diff
@@ -862,18 +859,15 @@ def draw_image(self, img, rect=None):
...
-            converted_img = img.convert_pixel_format("rgba32", inplace=0)
-            flipped_array = numpy.flipud(converted_img.bmp_array)
-            img_width, img_height = (
-                converted_img.width(),
-                converted_img.height(),
-            )
+        elif isinstance(img, Image.Image):
+            converted_img = img.convert("RGBA")
+            flipped_array = numpy.flipud(numpy.array(converted_img))
+            img_width, img_height = img.width, img.height
```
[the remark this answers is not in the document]
jwiggins: Actually, these names are used later in the method

kiva/qpainter.py
```diff
@@ -553,48 +553,27 @@ def radial_gradient(self, cx, cy, r, fx, fy, stops, spread_method,
...
-            """
-            y, x, d = array.shape
-            pad = (lambda v: (4 - (v % 4)) % 4)
-            nx = max(x + pad(x), 12)
-            ny = max(y, 12)
-            if x == nx and y == ny:
-                return array
-            ret = np.zeros((ny, nx, d), dtype=np.uint8)
-            ret[:y, :x] = array[:]
-            return ret
```
rahulporuri, line 574: Just to confirm, we don't need to copy and pad the array data - because the use of `PIL.Image.Image` means that this is implicitly handled. Correct?

kiva/tests/test_qpainter_drawing.py
```diff
@@ -42,11 +42,6 @@ def create_graphics_context(self, width, height):
 
         return GraphicsContext((width, height))
 
-    @unittest.expectedFailure
-    def test_image(self):
-        """ QPainter interprets images as BGRA. """
-        super().test_image()
```
rahulporuri, line 48: i'm not entirely sure why this test was removed - instead of unmarking it as an expected failure.

rahulporuri (approved):
The code changes make sense to me and IIUC, the code is implicitly tested. The code definitely looks cleaner - and mostly the same between the various kiva backends.

Not sure if adding unittests make the code easier to understand.

kiva/tests/test_qpainter_drawing.py
jwiggins: It only existed here so that it could be marked with `expectedFailure`. It's no longer needed here.

kiva/qpainter.py
jwiggins: We do, but PIL handles it for us now.

jwiggins:
@corranwebster Do you have any more feedback or questions?

corranwebster:
> @corranwebster Do you have any more feedback or questions?

I think I'm good with this.

jwiggins:
Thanks for the feedback

Issue #517, opened 2020-12-28
HiDPI draws incorrectly w/ OpenGL backend


jwiggins:
When using the OpenGL backend on a computer with a HiDPI display, the drawing only covers the bottom quadrant of the Enable `Window`:
![Screen Shot 2020-12-28 at 12 09 24 PM](https://user-images.githubusercontent.com/587491/103210330-a3668080-4905-11eb-9d04-db4085caaeff.png)

Additionally, the `kiva.tests.test_gl_drawing` tests fail when comparing the output images to what was expected because the images from the GC are scaled for the higher DPI display. This appears to be something that Pyglet's `Window` class is doing (as seen by calling `get_framebuffer_size`).

This might be tackled at the same time as #412

Issue #412, opened 2020-09-25
HiDPI/Retina support for Enable


corranwebster:
Recent releases of Wx and Qt permit support of high-resolution HiDPI/Retina displays.  Enable should detect these situations and adapt the resolution of image-based `Window` implementations to render appropriately.

All that is likely needed is to increase the size of the array being rendered into and apply a global scale factor to the gc immediately before calling the top-level component's draw method, so implementation is hopefully easy.

Issue #390, opened 2020-02-15
Agg backends don't scale line widths


corranwebster:
When rendering with the QPainter backend, or other backends like the PDF backend, scaling is applied to rendered lines, so they get wider (seed the PDF below, which the rectangle has been scaled by (2, 1) vs. the same image rendered with the agg backend.

[test.pdf](https://github.com/enthought/enable/files/4208600/test.pdf)

![test](https://user-images.githubusercontent.com/600761/74589988-3eae7a00-5002-11ea-9ff8-e9574bf6a1b6.png)

This inconsistency will possibly create problems as appearance may be different on different backends, or may be problematic when trying to save a report including a Chaco plot or similar.

jwiggins:
I think I might know how to fix this in `celiagg`, but not necessarily in `kiva.agg`. [This article](https://www.haiku-os.org/documents/dev/painter_and_how_agg_works/) is a very lucid explanation of how the rasterization pipeline of AGG works.

If this can be fixed, then the solution will involve wrapping the `VertexSource` object as late as possible by `conv_transform`. Or at the very least, making sure `conv_transform` is wrapped around `conv_stroke` so that transformation happens after the stroke geometry is created.

jwiggins:
OK. I've fixed this in `celiagg`. It wasn't too hard. `kiva.agg` is a different story entirely. For context: https://github.com/enthought/enable/blob/8bc13b1709a7c32b7961ee11b3e2e903af9f70ee/kiva/agg/src/kiva_compiled_path.h#L29-L42

Basically what's going on in `kiva.agg` is that the `path` of `kiva::graphics_context_base` holds the affine transformation for the whole graphics context. (that's right) So as points are added to the path, they are transformed by the current, er, transform. Then when rasterizing a path, no transform needs to be used because the path's points are already where they need to be. This is different than Quartz, `QPainter`, `celiagg`, and probably other vector graphics libraries. I'm not sure why it was done this way in `kiva.agg` (performance?). But the takeaway is that fixing this bug is a much larger refactor and much riskier than the one required for `celiagg`.

PR #591, opened 2021-02-16
Add HiDPI support for the Qt toolkit

jwiggins:
This is the Qt half of #412 and #517. I haven't looked into how Wx does HiDPI display yet, but figuring out how to do it for Qt5 first will likely make adding Wx "easy".

Known broken:
* ~`Component.use_backbuffer` creates a low resolution buffer.~
* OpenGL clipping is a bit off
* ~`GraphicsContext.clip_to_rect` uses absolute coordinates, so supporting HiDPI by changing the scale of a GC will break lots of code [but not all] which uses `clip_to_rect`~ `GraphicsContext.clip_to_rect` is not uniformly implemented with respect to the affine transformation across all kiva backends! (#592)
* ~`GraphicsContext.draw_marker_at_points` in the `kiva.agg` backend will draw nothing if there's anything more than a translation applied to the transformation matrix.~ Covered by #594

### DIFF
--- a/enable/abstract_window.py
+++ b/enable/abstract_window.py
@@ -14,7 +14,7 @@ from numpy import dot
 
 # Enthought library imports
 from traits.api import (
-    Any, Bool, Event, HasTraits, Instance, List, Property, Trait, Tuple,
+    Any, Bool, Event, Float, HasTraits, Instance, List, Property, Trait, Tuple,
 )
 
 
@@ -55,6 +55,14 @@ class AbstractWindow(HasTraits):
     # When a component captures the mouse, it can optionally store a
     # dispatch order for events (until it releases the mouse).
     mouse_owner_dispatch_history = Trait(None, None, List)
+
+    # A scaling constant applied to any GraphicsContext used for drawing the
+    # window's component.
+    base_pixel_scale = Float(1.0)
+
+    # When True, allow `base_pixel_scale` to be greater than 1 if the
+    # underlying toolkit supports it.
+    high_resolution = Bool(True)
 
     # The background window of the window.  The entire window first gets
     # painted with this color before the component gets to draw.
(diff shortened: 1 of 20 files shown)


corranwebster:
This looks generally fine in terms of implementation.  I haven't taken it for a spin yet, so there may be some glitches out there still.

There are some edge-cases that I can see around what happens if the resolution changes while the window is live (either by the user changing the resolution, or moving between screens), but as-is should be no worse than what we have now.  I think it can be supported in Qt if we wanted to go there, but worst case is that you get low-resolution images after moving from low-res to high-res; or you render at high-res when working with low-res.

This PR should include some tests: testing the Kiva GCs should be straightforward, but testing that the resolution is set correctly in CI is probably going to need a mock/patch of the appropriate Qt methods - I have no idea what headless boxes are going to report as screen resolution.

corranwebster:
Just had a thought where there may be additional breakage: in save methods we probably need to adjust the dpi appropriately.  Pillow will take the raw data just fine, but won't pass the scale information through to saved files.

jwiggins:
I've added some modifications to the kiva drawing tests which exercise the pixel scaling code.

jwiggins:
*Shakes fist at OpenGL backend* 😠

jwiggins:
#593 added to track saved image DPI concerns.

corranwebster:
Playing around with this a little, it seems to be working for the QPainter backend, but I think #390 might be impacting the drawing in the Agg backend - a little hard to tell as the number of pixels a line requires depends on the fractional position of the line - or alternatively it may be that we are snapping lines to coordinates somewhere (I have a vague recollection of such code).

jwiggins:
#390 is definitely a factor for the AGG backends. They're definitely implemented incorrectly, but I can only imagine how to fix `celiagg` and not how to fix `kiva.agg`.

corranwebster (approved):

jwiggins:
Thanks for the testing and feedback.

PR #592, opened 2021-02-17
Celiagg's clip_to_rect() should respect the transform matrix

### LINKED ISSUE #304: Celiagg backend doesn't save images correctly
From enable, when I call `save` to put the contents of a celiagg gc into a .png file, I get this:

component_ex

where if I use the old agg backend (or indeed anything else), I get this:

component_ex

which is what I would expect based on the colors, etc. I was using.

Based on a little experimentation, the [channel shuffling](https://github.com/enthought/enable/blob/master/kiva/celiagg.py#L748) doesn't seem to be needed

jwiggins:
As part of the exploration for #591, I discovered that `clip_to_rect` should respect the current transformation and the celiagg backend was not doing this.

Fixes #304 (as a side effect of adding unit tests for celiagg)

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -90,6 +90,7 @@ supported_combinations = {
 
 dependencies = {
     "apptools",
+    "celiagg",
     "coverage",
     "Cython",
     "fonttools",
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM but can you find time to add a regression test?

jwiggins:
I'll see what I can do

jwiggins:
Thanks for the review

PR #594, opened 2021-02-18
Allow scaling in agg's draw_marker_at_points

jwiggins:
This will be need to keep Chaco's colormapped scatterplot renderer working if #591 is merged.

### DIFF
--- a/kiva/agg/src/kiva_graphics_context.h
+++ b/kiva/agg/src/kiva_graphics_context.h
@@ -1170,7 +1170,7 @@ namespace kiva
         int success = 0;
         agg24::trans_affine ctm = this->get_ctm();
 
-        if ( only_translation(ctm) &&
+        if ( only_scale_and_translation(ctm) &&
             (this->state.line_width == 1.0 ||
              this->state.line_width == 0.0))
              //&& this->state.line_join == JOIN_MITER )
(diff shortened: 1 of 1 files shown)


kiva/agg/src/kiva_graphics_context.h
```diff
@@ -1184,12 +1184,16 @@ namespace kiva
...
+            get_scale(ctm, &sx, &sy);
 
             for(int i = 0; i < Npts*2; i+=2)
             {
-                m.marker((int)(pts[i]+tx), int(pts[i+1]+ty), size, type);
+                mx = pts[i];
+                my = pts[i+1];
+                ctm.transform(&mx, &my);
+                // XXX: Assuming scale is uniform in both directions
+                m.marker(int(mx), int(my), size * sx, type);
```
rahulporuri, line 1196: I am not entirely sure how this change works.

Previously, we were storing the translation in x and y as `tx` and `ty` - and then adding markers at `pts[i]+tx` and `pts[i+1]+ty`.

Now, we're storing the scale information in `sx` and `sy`, transforming the context manager by `pts[i]` and `pts[i+1]` and then also add markers at `pts[i]` and `pts[i+1]`?

jwiggins: `ctm` is the "current transformation matrix". We're applying it to `(mx, my)`, which the position of the current marker. Then we're calling the `marker` method on the marker renderer to render a marker at the position `(mx, my)` and scaling its size by `sx`.

rahulporuri (approved):

jwiggins:
Thanks for the review

PR #595, opened 2021-02-19
Add HiDPI support for Wx

### LINKED ISSUE #412: HiDPI/Retina support for Enable

### LINKED ISSUE #517: HiDPI draws incorrectly w/ OpenGL backend

jwiggins:
This is the Wx part of #591

Fixes #517
Fixes #412

### DIFF
--- a/enable/wx/base_window.py
+++ b/enable/wx/base_window.py
@@ -177,6 +177,10 @@ class BaseWindow(AbstractWindow):
         # Create the delegate:
         self.control = control = self._create_control(parent, wid, pos, size)
 
+        # Figure out the pixel scale factor
+        if self.high_resolution:
+            self.base_pixel_scale = self.control.GetContentScaleFactor()
+
         # Set up the 'erase background' event handler:
         control.Bind(wx.EVT_ERASE_BACKGROUND, self._on_erase_background)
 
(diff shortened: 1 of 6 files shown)


rahulporuri (approved):
LGTM but i definitely see a few unrelated fixes in this PR.

jwiggins:
> i definitely see a few unrelated fixes in this PR.

Don't look a gift horse in the mouth

rahulporuri:
> Don't look a gift horse in the mouth

🤣

PR #605, opened 2021-02-22
Add face_index support to celiagg and kiva.agg backends.

### LINKED ISSUE #391: Matching TTF fonts isn't working well
Several problems with the algorithm:

- [ ] ttfFontProperties looks into the TFFont to find the 'sfnt4' string with styling information.  It then does very dumb matching for key phrases, such as "italic", but doesn't normalize case, doesn't account for phrases which are sub-phrases of other phrases (eg. "bold" in "demibold"), and could get mismatches if a phrase is part of the name of the font (eg. "book" or "roman").
- [x] when matching .ttc files we are only returning the file, not the index of the font in the .ttc file.  FreeType2 needs this index to be able to find the correct font.  Currently the API has no way of even passing this information through to the Agg/Celiagg.

jwiggins:
Fixes #391

~Leaving this as a draft for now, because it depends on the latest release of celiagg which is not yet available via EDM.~

Aside from that this is a pretty scary change due to how twisty the font manager code is [and uncertainty about 3rd-party callers of said code]. It's be great to get some extra eyes on this before we move forward.


kiva/fonttools/tests/test_font.py
```diff
@@ -30,17 +30,19 @@ def test_find_font_empty_name(self):
         # the path from which the font manager loads font files, then this test
         # can be less fragile.
         font = Font(face_name="")
-        font_file_path = font.findfont()
+        font_file_path, face_index = font.findfont()
```
kitchoi, line 33: This is a breaking change to a public API. I believe there exists downstream code that calls findfont like this.
Is this intended? Alternative to breaking the API, one can add a method for returning the index and keeping `findfont` unchanged?

kitchoi (commented):
The little bit of exposure I had with font manager does not seem enough to qualify me as that second pair of 👀 .  I would want to sit down to give this a harder look if I am to review this, but realistically speaking, I am not sure I have the bandwidth to do that. I am sorry.

One comment after a quick glance.

kiva/fonttools/tests/test_font.py
kitchoi: > I believe there exists downstream code that calls findfont like this.

Or maybe not. Not sure.

jwiggins: There _shouldn't_ be downstream code which calls this, but that doesn't mean there _isn't_.

kitchoi: I see. In that case, if this signature is to be changed, perhaps it could be changed to return an object that wraps the file path and the index?
If we really really need to restore the old behaviour (e.g. it breaks things downstream), it might be enough to implement `__fspath__ ` to make the new object pathlike.

jwiggins:
We can wrap this up once #633 is merged.

kiva/fonttools/tests/test_font.py
jwiggins: Done

kiva/agg/agg-24/font_freetype/agg_font_freetype.cpp
```diff
@@ -567,12 +569,17 @@ namespace agg24
...
         unsigned i;
+
+        // Build the lookup string by combining the face_index and face_name
+        snprintf(m_face_lookup_scratch_space, _face_lookup_scratch_length,
+                 "%04u%s", face_index, face_name);
+
         for(i = 0; i < m_num_faces; ++i)
         {
-            if(strcmp(face_name, m_face_names[i]) == 0) return i;
+            if(strcmp(m_face_lookup_scratch_space, m_face_names[i]) == 0) return i;
```
jwiggins, line 582: See celiagg/celiagg#84

jwiggins:
OK. I've tested this on Windows and macOS and am happy where it's at. I'm ready for a full review!

kiva/agg/agg-24/font_freetype/agg_font_freetype.cpp
```diff
@@ -490,6 +490,7 @@ namespace agg24
 
 
 
+    static const unsigned _face_lookup_scratch_length = 1024;
```
corranwebster, line 493: This is limiting the total length of the face name + face index in the lookup string?

Is there a possible failure mode where a user passes a _really_ long font name, even if it is bogus?

jwiggins: Seemed like a reasonable value, but maybe [32k](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation) is better?

For performance reasons, I'd really like to avoid allocating this every time `find_face` is called (which is every time `load_font` is called, which can be a lot).

aaronayres35:
I have not reviewed the code changes at all, but here is an update of what I see when I run the benchmark suite on this branch:

show_text:
Screen Shot 2021-03-05 at 9 52 18 AM

For some strange reason kiva.agg does not show up in this table, even though on the main table it shows it passed.

show_text_2x:
Screen Shot 2021-03-05 at 9 53 11 AM


show_text_radial_gradient:
Screen Shot 2021-03-05 at 9 53 30 AM

kiva/agg/agg-24/font_freetype/agg_font_freetype.cpp
corranwebster: Looks like TrueType font names can't be longer than a UInt16 (digging in here: https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6name.html).  So it depends - if this holds a path, then I guess max path length + a bit makes sense, otherwise 64k + a bit should be safe.

jwiggins: This holds a path. I suppose the same "allocate a new buffer when the input grows" approach could be used here as what is used for the `m_signature` member variable in this same class.

aaronayres35:
After a git clean and a rebuild, I no longer see any italicized text 👍

kiva/agg/agg-24/font_freetype/agg_font_freetype.cpp
```diff
@@ -567,12 +569,24 @@ namespace agg24
 
 
     //------------------------------------------------------------------------
-    int font_engine_freetype_base::find_face(const char* face_name) const
+    int font_engine_freetype_base::find_face(const char* face_name, unsigned face_name_len, unsigned face_index)
     {
         unsigned i;
+
+        if ((face_name_len + 4 + 1) > m_face_lookup_scratch_len)
```
rahulporuri, line 576: i'm not entirely sure what the magical 4 and 1 here are.

rahulporuri (approved):
LGTM with a couple of comments. Given the current refactor of font code that is happening in parallel, I guess we need to redo some of these changes?

kiva/fonttools/tests/test_font.py
```diff
@@ -30,17 +30,19 @@ def test_find_font_empty_name(self):
...
         font = Font(face_name="ProbablyNotFound")
 
         # There will be warnings as there will be no match for the requested
         # face name.
         with self.assertWarns(UserWarning):
-            font_file_path = font.findfont()
-        self.assertTrue(os.path.exists(font_file_path))
+            spec = font.findfont()
+        self.assertTrue(os.path.exists(spec.filename))
+        self.assertEqual(spec.face_index, 0)
```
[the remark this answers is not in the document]
jwiggins: A test for `face_index`? What do you mean? And yes, it'll probably be easier after integrating these changes with the new font manager code.

kiva/agg/agg-24/font_freetype/agg_font_freetype.cpp
jwiggins: 4 for the face index (this is the `%04u` in the format string) and 1 for the NULL.

jwiggins:
Thanks for the feedback and testing. I'll integrate this into my larger font manager refactor and then we'll go from there.

PR #607, opened 2021-02-22
Add Cython and SWIG to cron job build deps

jwiggins:
Cron job is failing: https://github.com/enthought/enable/actions/runs/579901110

Adding Cython and SWIG to the environment _should_ fix the problem


rahulporuri (approved):
LGTM

jwiggins:
🤞 Hopefully that gets us running again

PR #609, opened 2021-02-23
Pin celiagg to a working version

jwiggins:
Celiagg 1.1.1 breaks text rendering on Windows. Until that's fixed, we should pin the installed celiagg version to the previous release (1.0.3).


rahulporuri (approved):
LGTM

jwiggins:
Thanks for the quick review

PR #611, opened 2021-02-23
Update Documentation for 5.0.0

aaronayres35:
targeting [gh-pages]

These are the steps I followed for building the documentation:
```
git clean -ffxd
git checkout 5.0.0
python ci/edmtool.py install
edm shell -e enable-test-3.6-null
python setup.py install
exit
python ci/edmtool.py docs
git checkout gh-pages
git pull
git checkout -b gh-pages-5.0
mkdir 4.6 # it appeared that 4.6 was last time gh-pages was updated 
mkdir 5.0
rsync -r --exclude="\.buildinfo" docs/build/html/* 5.0/
rm -r docs
rm -r enable
rm -r enable.egg-info/
mv `ls -d [![:digit:]]*` 4.6/
git add .
git commit -m "move old docs into 4.6 and new docs into 5.0 folders"
cp -r 5.0/* .
git add .
git commit -m "add current docs to root directory"
```

To test:
`python -m http.server`


jwiggins:
> `git add .`

I think this is what blew up your branch. You brought in an entire built version of a enable. It's huge.

Right. The `gh-pages` branch is weird because it's almost like a completely different repository. Perhaps it would be best to move the docs out-of-tree, then check out `gh-pages` and immediately do a `git clean -fxd`. Once you're at a clean base, then you should sync the build docs from wherever safe place you put them.

aaronayres35:
> > `git add .`
> 
> I think this is what blew up your branch. You brought in an entire built version of a enable. It's huge.
> 
> Right. The `gh-pages` branch is weird because it's almost like a completely different repository. Perhaps it would be best to move the built docs out-of-tree, then check out `gh-pages` and immediately do a `git clean -fxd`. Once you're at a clean base, then you should sync the built docs from wherever safe place you put them.

Ah yeah I see, my mistake.  I will do as you suggested in a new PR

PR #612, opened 2021-02-23
Update Documentation for 5.0.0

aaronayres35:
targeting [gh-pages]

These are the steps I followed for building the documentation:
```
git clean -ffxd
git checkout 5.0.0
python ci/edmtool.py install
edm shell -e enable-test-3.6-null
python setup.py install
exit
python ci/edmtool.py docs
# move docs/build/html folder to my desktop
git checkout gh-pages
git clean -ffxd
git pull
git checkout -b gh-pages-5.0
mkdir 4.6 # it appeared that 4.6 was last time gh-pages was updated 
mkdir 5.0
mv `ls -d [![:digit:]]*` 4.6/
git add .
git commit -m "move old docs to 4.6 folder"
# move html folder containing built docs into cwd
rsync -r --exclude="\.buildinfo" html/* 5.0/
rm -r html
cp -r 5.0/* .
git add .
git commit -m "add current docs to 5.0 folder and root directory"
```

To test:
`python -m http.server`

I think this should be better now.  The diff is still rather large (1,185 files changed), but I think the process made sense.


aaronayres35:
Seems like for some reason github isn't picking up on certain files being renamed and shows them as new files from when I moved the old contents into the 4.6 folder

jwiggins:
What's the purpose behind the `4.6` and `5.0` directories? `5.0` is a full copy of the root directory, so I think it's not currently necessary.

It looks like you're doing what is already done for traits (and possibly other ETS packages)? In that case, the `4.6` directory is all that is needed. Are we linking these different versions from anywhere, or are they an easter egg?

rahulporuri:
> It looks like you're doing what is already done for traits (and possibly other ETS packages)? In that case, the 4.6 directory is all that is needed. Are we linking these different versions from anywhere, or are they an easter egg?

There was discussion earlier to enable users to select the specific version to browse the documentation for (for all active ETS projects). One of the things that came out of that discussion was to allow creation of permalinks to a specific version - which aren't possible unless we use the `5.0` full copy. (Well, this is one solution and it's the easiest one - which we've chosen to go with for now).

rahulporuri (approved):
LGTM. I didn't verify all of the changes made in this PR - but I did checkout this branch and test that the served docs worked as expected.

Issue #360, opened 2019-07-11
SWIG 4.0 appears to break the kiva.agg backend


jwiggins:
See #359 for the full saga. For now, the solution is to just build with 3.0.x.

yves-surrel:
I confirm the problem and the short-term fix

swt2c:
Any progress on the long-term fix?  :)  Debian has upgraded to swig 4.0 and we're seeing problems there as well.  It seems that `_swig_setattr` has been removed as well.

```
...
test_minimum_vertical (enable.tests.qt4.scrollbar_test.ScrollBarTest) ... Traceback (most recent call last):
  File "/build/1st/python-enable-4.8.1/.pybuild/cpython3_3.8_enable/build/enable/qt4/base_window.py", line 213, in paintEvent
    self.handler.paintEvent(event)
  File "/build/1st/python-enable-4.8.1/.pybuild/cpython3_3.8_enable/build/enable/qt4/base_window.py", line 65, in paintEvent
    self._enable_window._paint(event)
  File "/build/1st/python-enable-4.8.1/.pybuild/cpython3_3.8_enable/build/enable/abstract_window.py", line 462, in _paint
    self._gc = self._create_gc(size)
  File "/build/1st/python-enable-4.8.1/.pybuild/cpython3_3.8_enable/build/enable/qt4/image.py", line 20, in _create_gc
    gc = GraphicsContext((size[0]+1, size[1]+1), pix_format=pix_format,
  File "/build/1st/python-enable-4.8.1/.pybuild/cpython3_3.8_enable/build/kiva/agg/__init__.py", line 38, in __init__
    GraphicsContextArray.__init__(self, pixel_map.bmp_array,
  File "/build/1st/python-enable-4.8.1/.pybuild/cpython3_3.8_enable/build/kiva/agg/agg.py", line 859, in __init__
    _swig_setattr(self, GraphicsContextArray, 'this', obj)
NameError: name '_swig_setattr' is not defined
Aborted
```

lateralguy:
I was facing a similar issue after rolling back the packages to the following version:
enable = 4.8.0
swig = 3.0.12
The issue was resolved.

PR #637, opened 2021-02-26
Install SWIG with apt-get

jwiggins:
The cron job is still broken because it was trying to get SWIG from PyPI. The correct source for SWIG is `apt-get`.

And after one test run, `celiagg` also needs to be installed but we can get that from PyPI.

### DIFF
--- a/.github/workflows/bleeding-edge.yml
+++ b/.github/workflows/bleeding-edge.yml
@@ -44,13 +44,15 @@ jobs:
         run: sudo apt-get install libglu1-mesa-dev
       - name: Install build dependencies
         run: |
+          sudo apt-get install swig
           python -m pip install -U pip setuptools wheel
           python -m pip install numpy
           python -m pip install Cython
-          python -m pip install swig==3.0.11
       - name: Install prebuilt wxPython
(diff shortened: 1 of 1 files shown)


.github/workflows/bleeding-edge.yml
```diff
@@ -44,13 +44,15 @@ jobs:
         run: sudo apt-get install libglu1-mesa-dev
       - name: Install build dependencies
         run: |
+          sudo apt-get install swig
```
rahulporuri, line 47: I'm not entirely sure how this will fix the issue because I thought Swig 4.x breaks the agg backend - https://github.com/enthought/enable/issues/360. And IIUC, CI on this PR won't break because this is a cron job.

Can you edit the job schedule to make it run in this PR so that we can verify that this PR fixes the cron job?

jwiggins: It works: https://github.com/enthought/enable/runs/1986218229

This is because we restrict the jobs to Ubuntu 18.04 precisely _because of_ #360.

rahulporuri (approved):
LGTM

jwiggins:
Thanks for the review

PR #638, opened 2021-02-26
Fix KeySpec.from_string

### LINKED ISSUE #636: KeySpec.from_string class method does not work
It fails because the set instantiation syntax is invalid: https://github.com/enthought/enable/blob/c78ee6f73e66c1f3235ddff8659c2fbe6efa45ae/enable/base_tool.py#L76

jwiggins:
Closes #636

### DIFF
--- a/enable/base_tool.py
+++ b/enable/base_tool.py
@@ -73,7 +73,7 @@ class KeySpec(object):
         codes = s.split("+")
         key = codes[-1]
         modifiers = set(code.lower() for code in codes[:-1])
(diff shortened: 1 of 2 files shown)


rahulporuri (approved):
LGTM

rahulporuri:
Surprisingly, it looks like this bug was introduced during the Python 2 to 3 conversion in https://github.com/enthought/enable/pull/209

jwiggins:
Thanks for the review.

PR #635, opened 2021-02-25
Don't mess up the component bounds in HiDPI mode

jwiggins:
Fixes #634

PR #640, opened 2021-02-26
Cherry pick fixes into the maint/5.0 branch

jwiggins:
We'll be releasing a 5.0.1 version in the not-too-distant future. Included in that should be the fixes from #635 and #638.


rahulporuri (approved):
LGTM. The changelog entries will happen in a separate PR, right?

jwiggins:
> The changelog entries will happen in a separate PR, right?

We can handle that as part of the release process.

PR #641, opened 2021-02-26
Fix font selection in the QPainter backend

jwiggins:
This is a bit of a drive-by fix. While testing #605 and comparing to the `kiva.qpainter` backend, I realized that it was ignoring bold and italic preferences for fonts. This corrects that, such that the `style` attribute of a `Font` is now honored.


kiva/qpainter.py
```diff
@@ -612,15 +618,23 @@ def draw_image(self, img, rect=None):
     # Drawing Text
     # ----------------------------------------------------------------
 
-    def select_font(self, name, size, textEncoding):
+    def select_font(self, name, size, style="regular", encoding=None):
```
jwiggins, line 621: Note that this is the signature from `AbstractGraphicsContext`

rahulporuri (approved):
LGTM

jwiggins:
Thanks for the review

PR #645, opened 2021-02-28
Fix save() in the quartz backend

jwiggins:
`save()` doesn't work at all currently. Implement the buffer interface and use `PIL.Image.frombuffer()`

### DIFF
--- a/kiva/quartz/ABCGI.pyx
+++ b/kiva/quartz/ABCGI.pyx
@@ -1523,30 +1523,29 @@ cdef class CGBitmapContext(CGContext):
     def width(self):
         return CGBitmapContextGetWidth(self.context)
 
-    def __getsegcount__(self, void* tmp):
-        cdef int *lenp
-        lenp = <int *>tmp
-        if lenp != NULL:
-            lenp[0] = self.height()*self.bytes_per_row
-        return 1
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
+        cdef Py_ssize_t shape[2]
+        cdef Py_ssize_t strides[2]
 
-    def __getreadbuffer__(self, int segment, void** ptr):
-        # ignore invalid segment; the caller can't mean anything but the only
-        # segment available; we're all adults
-        ptr[0] = self.data
-        return self.height()*self.bytes_per_row
+        shape[0] = self.bytes_per_row
+        shape[1] = self.height()
+        strides[0] = self.bytes_per_row
+        strides[1] = 1
 
-    def __getwritebuffer__(self, int segment, void** ptr):
-        # ignore invalid segment; the caller can't mean anything but the only
-        # segment available; we're all adults
-        ptr[0] = self.data
-        return self.height()*self.bytes_per_row
+        buffer.buf = <char *>(self.data)
+        buffer.format = 'b'
+        buffer.internal = NULL
+        buffer.itemsize = 1
+        buffer.len = self.height() * self.bytes_per_row
+        buffer.ndim = 2
+        buffer.obj = self
+        buffer.readonly = 1
+        buffer.shape = shape
+        buffer.strides = strides
+        buffer.suboffsets = NULL
 
-    def __getcharbuffer__(self, int segment, char** ptr):
-        # ignore invalid segment; the caller can't mean anything but the only
-        # segment available; we're all adults
-        ptr[0] = <char*>(self.data)
-        return self.height()*self.bytes_per_row
+    def __releasebuffer__(self, Py_buffer *buffer):
+        pass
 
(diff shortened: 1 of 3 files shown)


jwiggins:
We didn't have drawing tests for the quartz backend, so I added them. The drawing tests rely on `save()` to work at all.

rahulporuri:
Closing and opening the PR to trigger CI jobs. Not sure why but the PR has been stuck on Yellow for a while now i.e. "Waiting for status to be reported"

kiva/quartz/ABCGI.pyx
```diff
@@ -1523,30 +1523,29 @@ cdef class CGBitmapContext(CGContext):
...
+        buffer.len = self.height() * self.bytes_per_row
+        buffer.ndim = 2
+        buffer.obj = self
+        buffer.readonly = 1
+        buffer.shape = shape
+        buffer.strides = strides
+        buffer.suboffsets = NULL
+
+    def __releasebuffer__(self, Py_buffer *buffer):
+        pass
```
aaronayres35, line 1548: I lacked any context on these changes so I just tried to follow along with this https://cython.readthedocs.io/en/latest/src/userguide/buffer.html 

From what I can see changes LGTM.  I'm guessing we don't have a need to do the reference count stuff in `__releasebuffer__` here?

kiva/tests/drawing_tester.py
```diff
@@ -186,17 +186,23 @@ def assertImageSavedWithContent(self, filename):
         """ Load the image and check that there is some content in it.
         """
         image = numpy.array(Image.open(filename))
-        # default is expected to be a totally white image
+        # Default is expected to be a totally white image.
+        # Therefore we check if the whole image is not white.
+
+        # Previously this method checked for red pixels. However this is
+        # not [currently] possible with the quartz backend because it writes
+        # out image with premultiplied alpha and none of its pixels are the
```
aaronayres35, line 194: This is from the `if 'A' in mode:` block in the above file correct?

aaronayres35 (approved):
LGTM

I can confirm the changes in this branch fixed the issues I was having with the quartz backend when running the benchmarking script.

kiva/tests/drawing_tester.py
jwiggins: It is not. The premultiplied alpha setting happens in the `__init__` method of the graphics context. I couldn't find a way to have normal alpha, so I changed the test and added this note.

kiva/quartz/ABCGI.pyx
jwiggins: Thanks for pointing this out. I re-read the Cython docs again and you're right that we don't need to worry about reference counting. We don't mutate the data in a way that is dangerous to buffers created here, so we don't need any special considerations in `__releasebuffer__`.

That said, I've made a small change to use dynamically allocated memory for `shape` and `strides` since the first version (with stack-allocated arrays) was likely quite fragile.

kiva/quartz/ABCGI.pyx
```diff
@@ -1523,30 +1523,39 @@ cdef class CGBitmapContext(CGContext):
...
+    def __releasebuffer__(self, Py_buffer *buffer):
+        """ When PyBuffer_Release is called on buffers from __getbuffer__.
+        """
+        # Just deallocate the shape and strides allocated by __getbuffer__.
+        # Since buffer.obj is a referenced counted reference to this object
+        # (thus keeping this object alive as long a connected buffer exists)
+        # and we don't mutate `self.data` outside of __init__ and __dealloc__,
+        # we have nothing further to do here.
+        PyMem_Free(buffer.shape)
+        PyMem_Free(buffer.strides)
```
jwiggins, line 1558: @aaronayres35 A bit more detail for future readers 😉

kiva/tests/drawing_tester.py
jwiggins: So there are a few constraints here:
* `alpha_info` can only be `kCGImageAlphaPremultipliedLast` or `kCGImageAlphaPremultipliedFirst` if `save()` is called on a graphics context.
* From the "Supported Pixel Formats" table on [this page](https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/drawingwithquartz2d/dq_context/dq_context.html#//apple_ref/doc/uid/TP30001066-CH203), you can see that only 8, 16, and 32bit pixel formats are supported, so we can't even make a packed 24bit pixel buffer which would make alpha-free saving possible.

We can definitely implement saving RGB/BGR images, but it's a question of time allocation.

aaronayres35: I see, thank you!

jwiggins:
Thanks for the feedback

PR #646, opened 2021-03-01
Document the available kiva backends

### LINKED ISSUE #621: Kiva: document available backends
We should at least explicitly list all available backends, and provide notes about where they might differ from the "normal" behaviour (eg. unsupported features, extra capabilities such as pagination for PDF backend, etc.)

jwiggins:
Closes #621

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -13,6 +13,7 @@ Enable Documentation
   enable/toolkit_selection.rst
 
   kiva/overview.rst
+  kiva/backends.rst
   kiva/quickref.rst
   kiva/fonts.rst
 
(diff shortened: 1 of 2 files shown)


corranwebster:
It might make sense to also mention the Python libraries that need to be installed (eg. celiagg, pycairo, reportlab, pyglet, etc.) for each backend.  I think most of these are listed in the `__extras_requires__` of `enable/__init__.py` but some are missing like `cairo` (possibly because there are different possible cairo wrappers?).

Similarly, there might be a point to mention if there are OS limitations (beyond the ones you have mentioned, such as Quartz).

rahulporuri (approved):
LGTM

jwiggins:
Thanks for the review

PR #647, opened 2021-03-01
Add a benchmark program for kiva graphics contexts

jwiggins:
This is a small program which collects timing information for key drawing methods of `GraphicsContext` for each [available] backend and then optionally generates a report in HTML format.

I'd like to continue to improve this program, but I wanted to keep this initial PR under 500 LOC. Ideas for future improvements:
* Add hyperlinks for the image outputs written by each benchmark to the timings table
* "Benchmark" the non-UI backends to get a clearer picture of feature coverage.
* Add a GitHub action which generates this report and adds it to the documentation site so that we can get an idea about relative performance over time.
* Don't ignore the GL backend.

### DIFF
--- a/enable/gcbench/bench.py
+++ b/enable/gcbench/bench.py
@@ -1,0 +1,129 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+import importlib
+import inspect
+import os
+import time
+
+import numpy as np
+
+_MAX_DURATION = 1.0
+_SIZE = (512, 512)
+_BACKENDS = {
+    "ui": {
+        "kiva.agg": "enable.null.image",
+        "cairo": "enable.null.cairo",
+        "celiagg": "enable.null.celiagg",
+        "qpainter": "enable.null.qpainter",
+        "quartz": "enable.null.quartz",
+    },
+    "file": {
+        "pdf": "enable.null.pdf",
+        "ps": "enable.null.ps",
+        "svg": "enable.null.svg",
+    },
+}
+
+
+def benchmark(outdir=None):
+    """ Benchmark all backends
+    """
+    suite = gen_suite()
+
+    results = {}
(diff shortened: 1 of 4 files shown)


enable/gcbench/bench.py
```diff
@@ -0,0 +1,129 @@
...
+# Thanks for using Enthought open source!
+import importlib
+import inspect
+import os
+import time
+
+import numpy as np
+
+_MAX_DURATION = 1.0
+_SIZE = (512, 512)
```
rahulporuri, line 18: Should we be working with larger screen sizes e.g. (1024, 1024) given that we are regularly working with larger screens now? Not sure how much of an effect that will have on the benchmarking.

rahulporuri (approved):
LGTM with one bikeshedding-style comment. I tested this locally (with and without the output directory) and it works as expected.

>  Ideas for future improvements:

Can we create a separate issue to keep track of these improvements?

rahulporuri:
The text rendering stands out to me 

|celiagg|agg|qpainter|
|---|---|---|
celiagg show_text 2x | kiva agg show_text 2x | qpainter show_text 2x|

enable/gcbench/bench.py
jwiggins: Yeah... Does this might lead to better results since you're drawing more pixels and spending more time in lower-level non-Kiva code. Only downside I see is that it results in larger images which we plan to upload along with the HTML output (when adding benchmarking results to documentation). We maybe don't need to worry about it too much. Images can be made smaller by making them easier to compress ("less noise").

jwiggins:
> The text rendering stands out to me

Whoa. That `QPainter` result looks wrong (why no space between lines?).

jwiggins:
I'm curious how Celiagg is ending up with a sans-serif font...

jwiggins:
I'll open issues for the other stuff. Merging now. Thanks for the review

rahulporuri:
> There's confusion about whether stroke color or fill color applies to text. I'm tempted to just make whatever AGG does the "correct" way and make sure the other backends match.

I found the non-black color surprising because I thought Black fill color would be the default. I'm happy with QPainter overall - apart from the issue with space between lines.

> I'm curious how Celiagg is ending up with a sans-serif font...

Isn't this a known issue with font usage on windows + celiagg?

jwiggins:
> Isn't this a known issue with font usage on windows + celiagg?

Yes, but I would still expect celiagg to use the same font as `kiva.agg`

Issue #416, opened 2020-09-25
Use Traits observers instead of listeners


corranwebster:
This perhaps should wait for Enable 5.1 rather than 5.0, but we should switch to using the new Traits observer system rather than the current listener system (ie. `on_trait_change`).

aaronayres35:
NOTE: 
Some of the changes in enable will require a corresponding change in `traitsUI`.  For example trying to observe a method defined on a base class in traitsUI in enable code.  That handler method will need its signature changed for things to work.
ref: https://github.com/enthought/enable/pull/663#discussion_r590504516

PR #654, opened 2021-03-01
Convert on_trait_change decorators to observe

aaronayres35:
part of #416 

This PR replaces all uses of the `@on_trait_change` decorator with the observe equivalent.

### DIFF
--- a/enable/savage/compliance/crosshair.py
+++ b/enable/savage/compliance/crosshair.py
@@ -11,7 +11,7 @@
 """
 
 from enable.api import BaseTool, ColorTrait, LineStyle
-from traits.api import Bool, Float, HasTraits, List, Tuple, on_trait_change
+from traits.api import Bool, Float, HasTraits, List, Tuple, observe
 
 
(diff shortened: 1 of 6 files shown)


enable/savage/compliance/crosshair.py
```diff
@@ -84,8 +84,8 @@ def normal_mouse_move(self, event):
         self.svg_coords = event.x, y
         event.handled = True
 
-    @on_trait_change("svg_coords,mouse_in")
-    def ensure_redraw(self):
+    @observe("svg_coords,mouse_in")
+    def ensure_redraw(self, event=None):
```
aaronayres35, line 88: This method and `cell_properties_changed` in a later file are never called standalone, but they are not privately named.  So I made `event` a kwarg to protect against breaks from other callers, but this is likely overkill/unneeded. I am happy to change this to just `event` if a reviewer prefers.

rahulporuri (approved):
LGTM

PR #655, opened 2021-03-01
fix edmtool test-all command

aaronayres35:
This solution is lifted directly from traitsUI.
Previously running `python ci/edmtool.py test-all` would only run the test suite on one (randomly chosen - first item in a set) toolkit.  Now all toolkits are run.


rahulporuri (approved):
LGTM

PR #656, opened 2021-03-01
Replace dynamic uses of on_trait_change with observe

aaronayres35:
part of #416 

This PR starts replacing some of the dynamic uses of `on_trait_change` with observe.  I had flagged it as a WIP but I think I will remove that now just so the PR is smaller to review and make additional changes in a separate PR.

### DIFF
--- a/enable/constraints_container.py
+++ b/enable/constraints_container.py
@@ -252,7 +252,7 @@ class ConstraintsContainer(Container):
         """
         # Remove stale components from the map
         for item in event.removed:
-            item.on_trait_change(
+            item.observe(
                 self._component_size_hint_changed,
                 "layout_size_hint",
                 remove=True,
@@ -267,7 +267,7 @@ class ConstraintsContainer(Container):
         """
         # Clear the component maps
         for key, item in self._component_map.items():
-            item.on_trait_change(
+            item.observe(
                 self._component_size_hint_changed,
                 "layout_size_hint",
                 remove=True,
@@ -277,7 +277,7 @@ class ConstraintsContainer(Container):
         # Check the new components
         self._check_and_add_components(new)
 
-    def _component_size_hint_changed(self):
+    def _component_size_hint_changed(self, event):
         """ Refresh the size hint contraints for a child component
         """
         self.relayout()
@@ -369,7 +369,7 @@ class ConstraintsContainer(Container):
                 raise ValueError(msg)
 
             self._component_map[key] = item
-            item.on_trait_change(
+            item.observe(
                 self._component_size_hint_changed, "layout_size_hint"
             )
 
--- a/enable/constraints_container.py
+++ b/enable/constraints_container.py
@@ -253,7 +253,7 @@ class ConstraintsContainer(Container):
         # Remove stale components from the map
         for item in event.removed:
             item.observe(
-                self._component_size_hint_changed,
+                self._handle_changed_component_size_hint,
                 "layout_size_hint",
                 remove=True,
             )
@@ -268,7 +268,7 @@ class ConstraintsContainer(Container):
         # Clear the component maps
         for key, item in self._component_map.items():
             item.observe(
-                self._component_size_hint_changed,
+                self._handle_changed_component_size_hint,
                 "layout_size_hint",
                 remove=True,
             )
@@ -277,7 +277,7 @@ class ConstraintsContainer(Container):
         # Check the new components
         self._check_and_add_components(new)
 
-    def _component_size_hint_changed(self, event):
+    def _handle_changed_component_size_hint(self, event):
         """ Refresh the size hint contraints for a child component
         """
         self.relayout()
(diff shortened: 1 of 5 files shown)


enable/scrolled.py
```diff
@@ -535,8 +539,8 @@ def _handle_vertical_scroll(self, position):
                     view_position=[viewport.view_position[0], position]
                 )
 
-    def _mouse_thumb_changed(self, object, attrname, event):
-        if event == "down" and not self.continuous_drag_update:
+    def _mouse_thumb_changed(self, event):
```
aaronayres35, line 542: This is confusing code, this looks like a magic named method for when a trait `mouse_thumb` changes, however this class does not have a `mouse_thumb` trait!  It is only used as a dynamically hooked up handler (see above).

This is confusing because if it were a magically named method the signature would be "_name_changed(name, old, new)" whereas when it is just an otc handler it is "handler(object, name, new)".  
This specifically isn't relevant for observe now, but this practice is one we probably want to avoid as it is confusing.

enable has lots of uses of these magically named methods, but it appears sometimes they get hooked up with on_trait_change in addition to already listening because of the traits magic for their name

enable/constraints_container.py
```diff
@@ -277,7 +277,7 @@ def __components_changed(self, new):
         # Check the new components
         self._check_and_add_components(new)
 
-    def _component_size_hint_changed(self):
+    def _component_size_hint_changed(self, event):
```
jwiggins, line 280: You should rename this method so that it doesn't look like a static notification handler.

enable/drag.py
```diff
@@ -39,7 +39,7 @@ def init(self, components, bounds_rect, drag_root, drag_bounds_rect,
         self.window = components[0].window
         self.drag_over = []
 
-        self.on_trait_change(self.drag_done, drag_event)
+        self.observe(self.drag_done, drag_event)
```
jwiggins, line 42: How do we know `drag_event` is compatible with `observe`?

enable/drag.py
```diff
@@ -156,18 +156,18 @@ def drag_done(self, event):
...
 
         # 'Unhook' the drag done notification handler:
-        self.on_trait_change(self.drag_done, self.drag_event, remove=True)
+        self.observe(self.drag_done, self.drag_event, remove=True)
 
         # Compute the new drag bounds:
-        x = event.x
-        y = event.y
+        x = event.new.x
+        y = event.new.y
```
jwiggins, line 163: This looks a bit sketchy. Like `event` is actually a `DragEvent` instance. So adding observe here makes things very confusing.

enable/scrolled.py
jwiggins: I think these should be renamed where we know it's safe to do so. Like if a trait with the name "mouse_thumb" doesn't exist anywhere in Enable or Chaco, then we could be pretty sure that a rename was safe here.

enable/drag.py
aaronayres35: `event.new` is the `DragEvent` instance, i was going to do `drag_event = event.new` and propagate the change down, but `self.drag_event` already exists.  Maybe `new_drag_event`?  

I agree this is very confusing, I will go back over this code

enable/drag.py
aaronayres35: Looking into this more I am quite confused... 

So the this `DragHandler` class is defining an `init` not `__init__` method.  However, it subclasses `Interactor`, which subclasses `HasTraits`.  Neither of these subclasses define/call an `init` method.  This is something that is defined for traitsUI editors to do additional initialization.

I put a print statement in the method and ran the test suite only to see it was never called.  This obviously doesn't guarantee anything but it raised some red flags about this class.  

searching for `DragHandler` in the enable and chaco codebases yielded no results.  I then searched on github and found no results (although my view of the full Enthought github is limited).

I don't think the code in this file would have worked correctly previously as I don't see when `init` could ever have been called, and then any other methods accessing `self._____` would fail as well... I also cant see the class used anywhere.  

@jwiggins should we consider removing this code?  It is very possible I am missing something and we want to keep it, but currently I am pretty confused

aaronayres35: likewise for `DragResizeHandler` in `enable/drag_resize.py`

aaronayres35: did a little more github digging, and outside of copyright headers, and style fixes the `enable/drag.py` file effectively hasn't been touched since it was written... which was in 2003

jwiggins: #685 explicitly names this module.

In the context of this PR what do you want to do? Leave it untouched, or edit it so that it doesn't show up when looking for old notification patterns?

aaronayres35: Ah that makes sense, I hadn't seen that on 685 thanks!

I am happy to just leave it untouched here

jwiggins (approved):
LGTM

PR #659, opened 2021-03-02
Add OpenGL support to the benchmark program

### LINKED ISSUE #652: Kiva Benchmark: Add OpenGL
The initial version of the benchmark program ignores OpenGL. We should look into including it now.

jwiggins:
Fixes #652

This wasn't as hard as I thought it would be. I just stole code from the kiva drawing tests.

### DIFF
--- a/enable/gcbench/bench.py
+++ b/enable/gcbench/bench.py
@@ -21,6 +21,7 @@ _BACKENDS = {
         "kiva.agg": "enable.null.image",
         "cairo": "enable.null.cairo",
         "celiagg": "enable.null.celiagg",
+        "opengl": "enable.gcbench.opengl",
         "qpainter": "enable.null.qpainter",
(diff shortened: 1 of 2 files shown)


rahulporuri:
|-|x|2x|
|---|---|---|
|draw marker at points|opengl draw_marker_at_points|opengl draw_marker_at_points 2x|
|draw path at points|opengl draw_path_at_points|opengl draw_path_at_points 2x|
|show text|opengl show_text|opengl show_text 2x|
|draw path|opengl draw_path|opengl draw_path 2x|

rahulporuri (approved):
LGTM. The code worked as expected locally but there were a bunch of issues that I got to know about looking at the output of the benchmark.

jwiggins:
Yeah. The OpenGL backend is pretty rotted.

Thanks for the review.

Issue #651, opened 2021-03-01
Kiva Benchmark: Add a GitHub Action which runs the benchmark and publishes the results


jwiggins:
I'm not sure if this is possible, but it would we nice if the action could open a PR against `gh-pages` which could then be reviewed by a human and merged as appropriate.

Step one is the determine the feasibility of this task.

rahulporuri:
This is a tangential comment but instead of linking to the images in the benchmark data, I feel it'll be better if we can create and display the images in a new/separate table - that way, comparing the output across the kiva backends becomes straightforward.

jwiggins:
I had _exactly_ the same thought. We should definitely do that.

jwiggins:
GitHub doesn't directly provide an action which can open new PRs, so we would need to roll it ourselves or audit one of the already available actions (and then pin to the audited commish).

I think this is currently more trouble than we have time for, but hopefully that will change.

Issue #648, opened 2021-03-01
Make sure all kiva backends implement the AbstractGraphicsContext faithfully


jwiggins:
Since `AbstractGraphicsContext` is the published and documented interface for kiva backends, we should make sure that all current backends actually use that interface.

* `select_font` is not consistently implemented
* Other methods?

corranwebster:
> The PDF, PS, and SVG backends have no `CompiledPath`

All the backends that inherit from `basecore2d` can use the Agg `CompiledPath` implementation, but they don't expose it in the module API:
https://github.com/enthought/enable/blob/dae67fea0b6729cf43841c1e45bcda62ea463739/kiva/basecore2d.py#L704-L743

jwiggins:
I'm pretty averse to creating new dependencies on the Agg backend. We should consider a different way.

corranwebster:
This isn't new, just noting that there is a default implementation.

I agree thought: I'd really like a pure python implementation of CompiledPath, together with a proper definition of the expected interface.

PR #661, opened 2021-03-02
Add a page for comparing backend outputs

jwiggins:
Mentioned on #651, but there's no issue.

### DIFF
--- a/enable/gcbench/bench.py
+++ b/enable/gcbench/bench.py
@@ -122,9 +122,9 @@ def gen_timings(gc, func):
 
     times = np.array(times)
     return {
-        "max": times.max() * 1000,
-        "min": times.min() * 1000,
         "mean": times.mean() * 1000,
+        "min": times.min() * 1000,
+        "max": times.max() * 1000,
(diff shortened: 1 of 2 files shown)


enable/gcbench/bench.py
```diff
@@ -122,9 +122,9 @@ def gen_timings(gc, func):
 
     times = np.array(times)
     return {
-        "max": times.max() * 1000,
-        "min": times.min() * 1000,
         "mean": times.mean() * 1000,
+        "min": times.min() * 1000,
+        "max": times.max() * 1000,
```
jwiggins, line 127: I'm using the order of this dictionary when turning it into an HTML table

rahulporuri: should be fine now that we only support python >= 3.6.

rahulporuri (approved):
I didn't thoroughly examine the code but I did check the output - which looks awesome. Thanks for putting this together! LGTM.

jwiggins:
Yeah. I really like seeing all the outputs next to each other. This is really helpful for #648

PR #663, opened 2021-03-02
Replace dynamic uses of on_trait_change with observe part 2

aaronayres35:
part of #416 

This PR finishes the job started in #656 by replacing the remaining dynamic uses of `on_trait_change` with `observe`.

### DIFF
--- a/enable/abstract_window.py
+++ b/enable/abstract_window.py
@@ -218,7 +218,7 @@ class AbstractWindow(HasTraits):
         # If possible, size the new component according to the size of the
         # toolkit control
         size = self._get_control_size()
-        if self.component is not None:
+        if new is not None:
             new.observe(self.component_bounds_updated, "bounds")
             if size is not None:
                 pix_scale = self.base_pixel_scale
--- a/enable/abstract_window.py
+++ b/enable/abstract_window.py
@@ -205,7 +205,7 @@ class AbstractWindow(HasTraits):
     def _component_changed(self, old, new):
         if old is not None:
             old.observe(
-                self.component_bounds_changed, "bounds", remove=True
+                self.component_bounds_updated, "bounds", remove=True
             )
             old.window = None
 
@@ -219,7 +219,7 @@ class AbstractWindow(HasTraits):
         # toolkit control
         size = self._get_control_size()
         if hasattr(self.component, "bounds"):
-            new.observe(self.component_bounds_changed, "bounds")
+            new.observe(self.component_bounds_updated, "bounds")
             if size is not None:
                 pix_scale = self.base_pixel_scale
                 if getattr(self.component, "fit_window", False):
@@ -237,7 +237,7 @@ class AbstractWindow(HasTraits):
         self._update_region = None
         self.redraw()
 
-    def component_bounds_changed(self, event):
+    def component_bounds_updated(self, event):
         """
         Dynamic trait listener that handles our component changing its size.
         """
--- a/enable/abstract_window.py
+++ b/enable/abstract_window.py
@@ -204,7 +204,7 @@ class AbstractWindow(HasTraits):
 
     def _component_changed(self, old, new):
         if old is not None:
-            old.on_trait_change(
+            old.observe(
                 self.component_bounds_changed, "bounds", remove=True
             )
             old.window = None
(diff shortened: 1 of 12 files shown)


enable/savage/trait_defs/ui/qt4/svg_button_editor.py
```diff
@@ -93,7 +93,7 @@ def prepare(self, parent):
         """
         name = self.extended_name
```
aaronayres35, line 94: BEWARE: extended_name

enable/abstract_window.py
```diff
@@ -218,25 +218,26 @@ def _component_changed(self, old, new):
...
-                    size[0] / pix_scale, size[1] / pix_scale
-                ]
-            elif hasattr(self.component, "resizable"):
-                if "h" in self.component.resizable:
-                    self.component.outer_x = 0
-                    self.component.outer_width = size[0] / pix_scale
-                if "v" in self.component.resizable:
-                    self.component.outer_y = 0
-                    self.component.outer_height = size[1] / pix_scale
+        if hasattr(self.component, "bounds"):
```
aaronayres35, line 221: I believe this was previously a bug, as there is a case in which size is None when a component is `new`, but later the method is called with that same component as `old`.  We were trying then to remove a listener that had never gotten hooked up.  `observe` complained about this and the test suite failed, with `on_trait_change` it flew under the radar

enable/abstract_window.py
```diff
@@ -218,25 +218,26 @@ def _component_changed(self, old, new):
...
+                        self.component.outer_y = 0
+                        self.component.outer_height = size[1] / pix_scale
         self._update_region = None
         self.redraw()
 
-    def component_bounds_changed(self, bounds):
+    def component_bounds_changed(self, event):
         """
         Dynamic trait listener that handles our component changing its size;
         bounds is a length-2 list of [width, height].
```
jwiggins, line 243: `bounds` is no longer a variable in this method.

enable/abstract_window.py
jwiggins: I just added a unit test which exercises this part of the code in #635. Can you add a test which would break if this change weren't here?

enable/drawing/drawing_canvas.py
```diff
@@ -159,18 +159,14 @@ def add_button(self, *buttons):
...
+                self._canvas_bounds_changed, "bounds.items", remove=True
             )
 
         if new:
-            new.on_trait_change(self._canvas_bounds_changed, "bounds")
-            new.on_trait_change(self._canvas_bounds_changed, "bounds_items")
+            new.observe(self._canvas_bounds_changed, "bounds.items")
 
-    def _canvas_bounds_changed(self):
+    def _canvas_bounds_changed(self, event):
```
jwiggins, line 169: Rename if you can determine that it would be safe

enable/text_field.py
```diff
@@ -460,12 +460,15 @@ def _refresh_viewed_line(self, line):
     def _acquire_focus(self, window):
         self._draw_cursor = True
         window.focus_owner = self
-        window.on_trait_change(self._focus_owner_changed, "focus_owner")
+        window.observe(self._focus_owner_changed, "focus_owner")
         self.request_redraw()
 
-    def _focus_owner_changed(self, obj, name, old, new):
+    def _focus_owner_changed(self, event):
```
jwiggins, line 466: This seems a bit sketchy. Can this method be renamed?

enable/savage/trait_defs/ui/qt4/svg_button_editor.py
jwiggins: Meaning what, exactly? We can be sure that `name` won't break when passed to `observe`?

aaronayres35: Yeah, a quick search of the code base for `extended_name` only gives this result here and the corresponding same thing in the wx file.  Any downstream users that set this trait _may_ run into problems if the extended name they are passing does not conform to the new mini language, and that is effectively out of our hands ... 

A similar scenario was encountered in pyface and we mentioned it in the release announcement.  I _imagine_ the vast majority of users would not be affected (I dont even know how many users of this code we have in total).  Nonetheless, it leaves a door open for potential breaks.


Another thing to worry about here though that I'm not sure what we would want to do for:  The traitsUI code for `Editor` contains the code that would theoretically remove this listener (in either the `dispose` method or the `_update_editor` method itself.  I also just realize I made a dumb mistake! The method being observed is `_update_editor` not `update_editor` which is defined in traitsUI.  We are still in the process of updating traitsUI to observe, so that method will need its signature changed... I believe we will need that around first, before we can make this change as this code won't work currently.  It must be uncovered in the test suite hence we don't see a failure here

enable/abstract_window.py
aaronayres35: I've added a test now.  It previously failed, and is now passing (so in that sense it is good). The test itself feels a bit ugly to me though in general :/ I'm very open to suggested modifications

jwiggins (approved):
LGTM. The test for `AbstractWindow` looks fine.

enable/savage/trait_defs/ui/qt4/svg_button_editor.py
```diff
@@ -93,7 +93,7 @@ def prepare(self, parent):
...
         if name != "None":
-            self.context_object.on_trait_change(
+            self.context_object.observe(
```
aaronayres35, line 96: 🚨 This depends on changes to be made in traitsUI ! 🚨 

similarly the changes made in `enable/savage/trait_defs/ui/wx/svg_button_editor.py`

I'm thinking we probably want to undo these changes here and hold off until a traitsUI release?

An alternative is to overwrite the `_update_editor` method here on the subclass and change the signature.  However, even with that, traitsUI is still responsible for removing this listener in the base Editor classes `dispose` method.  We could also overwrite that here, but at that point things are getting a little ugly and its probably better to just hold offf?

enable/abstract_window.py
```diff
@@ -218,28 +218,28 @@ def _component_changed(self, old, new):
...
-                    size[0] / pix_scale, size[1] / pix_scale
-                ]
-            elif hasattr(self.component, "resizable"):
-                if "h" in self.component.resizable:
-                    self.component.outer_x = 0
-                    self.component.outer_width = size[0] / pix_scale
-                if "v" in self.component.resizable:
-                    self.component.outer_y = 0
-                    self.component.outer_height = size[1] / pix_scale
+        if self.component is not None:
```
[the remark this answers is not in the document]
aaronayres35: I thought about this, I believe they are always the same object.  The code that follows does things with both, ie. `new.observe`, `self.component.outer_position = ...`
I wasn't sure what made more sense so I left it, but I'm happy to change given you had the same question

enable/savage/trait_defs/ui/qt4/svg_button_editor.py
aaronayres35: I've left a comment on the issue: https://github.com/enthought/enable/issues/416#issuecomment-794107671

jwiggins: OK. I think we're good here

PR #665, opened 2021-03-02
Replace depends_on with observe in properties

aaronayres35:
last part of #416 

This PR replaces all uses of `depends_on` in properties with `observe`.

### DIFF
--- a/enable/tools/drag_tool.py
+++ b/enable/tools/drag_tool.py
@@ -62,7 +62,7 @@ class DragTool(BaseTool):
 
     # private property to hold the current list of KeySpec instances of the
     # cancel keys
-    _cancel_keys = Property(List(KeySpec), depends_on="cancel_keys")
+    _cancel_keys = Property(List(KeySpec), observe="cancel_keys")
 
     # ------------------------------------------------------------------------
     # Interface for subclasses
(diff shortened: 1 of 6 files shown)


enable/tools/drag_tool.py
```diff
@@ -62,7 +62,7 @@ class DragTool(BaseTool):
 
     # private property to hold the current list of KeySpec instances of the
     # cancel keys
-    _cancel_keys = Property(List(KeySpec), depends_on="cancel_keys")
+    _cancel_keys = Property(List(KeySpec), observe="cancel_keys")
```
jwiggins, line 65: Do you need to do anything special if you're observing a `List` trait?

jwiggins (approved):
Just one question.

enable/tools/drag_tool.py
aaronayres35: Only if you are observing contents of the list, e.g. `container.value` used to listen for changes to the value trait of any item in the container, or listening to `container[]`, or `container_items`.  If you are just observing the list itself, you can leave as is

Example from the pyface changes 
Screen Shot 2021-03-02 at 11 50 13 AM

jwiggins: Cool. Thanks for clarifying

PR #666, opened 2021-03-02
Run the kiva benchmark on all backends

### LINKED ISSUE #650: Kiva Benchmark: Include file-based backends
The PDF, PS, and SVG backends should be exercised for capability tracking purposes, but not for performance.

jwiggins:
Resolves #650

Screen Shot 2021-03-02 at 5 12 45 PM

As you can see, there's something wrong with the PDF backend. That can be fixed elsewhere. Additionally, the code for writing the HTML output is getting really awful. It could benefit from some clean up if we think we might add any more enhancements.

### DIFF
--- a/enable/gcbench/bench.py
+++ b/enable/gcbench/bench.py
@@ -38,16 +38,26 @@ def benchmark(outdir=None):
     """
     suite = gen_suite()
 
-    results = {}
-    # NOTE: Only checking UI backends for now
-    for name, mod_name in _BACKENDS["ui"].items():
-        print(f"Benchmarking backend: {name}", end="")
-        try:
-            module = importlib.import_module(mod_name)
-        except ImportError:
-            print(" ... Not available")
-            continue
-        results[name] = benchmark_backend(suite, name, module, outdir=outdir)
+    results = {t: {} for t in _BACKENDS}
+    for btype, backends in _BACKENDS.items():
+        for name, mod_name in backends.items():
+            print(f"Benchmarking backend: {name}", end="")
+            try:
+                module = importlib.import_module(mod_name)
+            except ImportError:
+                print(" ... Not available")
+                continue
+
+            # UI backends are checked for performance, File backends are not.
+            if btype == "ui":
+                results[btype][name] = benchmark_backend(
+                    suite, name, module, outdir=outdir
+                )
+            else:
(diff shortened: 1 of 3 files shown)


jwiggins:
I've got the PDF backend partially working now. We can clean up the remaining red X's elsewhere.

enable/gcbench/bench.py
```diff
@@ -38,16 +38,26 @@ def benchmark(outdir=None):
...
-    # NOTE: Only checking UI backends for now
-    for name, mod_name in _BACKENDS["ui"].items():
-        print(f"Benchmarking backend: {name}", end="")
-        try:
-            module = importlib.import_module(mod_name)
-        except ImportError:
-            print(" ... Not available")
-            continue
-        results[name] = benchmark_backend(suite, name, module, outdir=outdir)
+    results = {t: {} for t in _BACKENDS}
```
aaronayres35, line 41: could 't' be 'backend_type` here or something like that?

enable/gcbench/publish.py
```diff
@@ -108,10 +114,8 @@ def _build_comparison_table(backends, comparisons):
         link = f'<a href="{name}.html">'
         row = [f"<td>{link}{name}</a></td>"]
         for bend in backends:
-            # Each backend stat includes a "valid" flag
-            stat, valid = stats[bend]
-            # Which gets used to add a CSS class for styling
-            klass = "valid" if valid else "invalid"
+            # Each backend stat includes a CSS class for table styling
+            stat, klass = stats[bend]
```
aaronayres35, line 118: maybe this stats could be called something different in the for loop above?  It is the stats after they have been run though a formatting function eg `_format_benchmark`.  

It is pretty inconsequential, but the code names are a little confusing as was thinking `stats` was a dict of a different form at first

aaronayres35 (approved):
LGTM, tested locally and everything looks correct.  The links to pdf and ps worked as well

Just a couple comments about some variable names.  The code is a little thick to read through and names might help make it slightly more understandable.  Also some numpy docstrings giving more details about parameters for the major functions could be very useful, but feel free to ignore at least for this PR.  That can be done as some cleanup later if we find the time

jwiggins:
> The code is a little thick to read through and names might help make it slightly more understandable.

That's being charitable. I'll take a quick shot at some cleanup this morning.

Thanks for the review/testing.

PR #667, opened 2021-03-02
Add gradients to benchmark suite

aaronayres35:
closes #664 

This PR adds 2 new benchmarks to the benchmark suite.  One draws a path and fills it using `linear_gradient`.  The other draws some text and fills it using `radial_gradient` after setting the text drawing mode to `TEXT_FILL_STROKE`.  I went with `TEXT_FILL_STROKE` instead of `FILL_STROKE` as the results seemed more sensible.  

Here are the results I currently see:
<details>

Screen Shot 2021-03-02 at 2 18 56 PM

Sorry for the very small screenshot.  The backends are in the order kiva.agg, celiagg, opengl, qpainter, quartz
Screen Shot 2021-03-02 at 2 19 54 PM
The discrepancy in the qpainter fill I suspect is related to #660 ?

Screen Shot 2021-03-02 at 2 20 13 PM

</details>

PR #668, opened 2021-03-03
Clean up the benchmark suite

jwiggins:
#666 was pretty exhausting to implement because of all the output differences introduced by non-GUI backends.

This is a general cleanup of the benchmark code, with emphasis on the `publish` function and introducing some real objects for the result data to hopefully make things a little more readable.

After this, I think the only thing left is to increase the size of the suite (ala #667) and _more importantly_, use the results as guidance for fixing problems and harmonizing backend outputs.

### DIFF
--- a/enable/gcbench/bench.py
+++ b/enable/gcbench/bench.py
@@ -14,10 +14,12 @@ import time
 
 import numpy as np
 
+from enable.gcbench.data import BenchResult, BenchTiming
+
 _MAX_DURATION = 1.0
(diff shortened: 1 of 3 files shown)


aaronayres35 (approved):
LGTM, thank you!

I ran the benchmark again and everything looked correct.  I like the table orientation being vertical now as it is easier to scroll

jwiggins:
Thanks for the review

Issue #414, opened 2020-09-25
Retire SWIG-based Agg backend in favour of Cython-based Celiagg


corranwebster:
The SWIG-based Agg wrapper is showing its age: it is using an old version of Agg; it doesn't work with modern GCCs, the build process is idiosyncratic, and it is overall hard to maintain and modify.  Celiagg is a Cython-based alternative wrapper of the Agg backend with better overall architecture and, importantly, is a stand-along package which is independent of Kiva.  There is already a Celiagg backend for Kiva.

We should switch the default `image` backend to use Celiagg, and deprecate and retire the SWIG-based Agg wrapper.

Most of the pieces are in place for this, but it will likely be a multi-phase process:

1. Audit Celiagg for feature compatibility (does it do everything right) and performance (we can take some performance degradation in exchange for reducing maintenance overhead, but 10x slower is too much).
2. If there are discrepancies, fix them.
3. Expose the current Agg backend as a separate-but-equivalent Kiva backend to the image backend ('oldagg' or 'swigagg' or something, not just 'agg').
4. Switch the image backend to use Celiagg (while keeping the current celiagg backend)
5. Deprecate the agg backend
6. Remove the agg backend

jwiggins:
#251 is connected here.

I believe the root issue is that `GraphicsContext.draw_image` is expecting a `GraphicsContext` instance as a possible argument. Unfortunately this is done in such a way which causes a hard dependency on the current agg backend in kiva.

Here's all the serious occurrences of context objects being blitted directly, as far as I can tell:
- https://github.com/enthought/enable/blob/4.8.1/kiva/agg/src/graphics_context.i#L719-L737
- https://github.com/enthought/enable/blob/4.8.1/kiva/quartz/ABCGI.pyx#L736-L737
- https://github.com/enthought/enable/blob/4.8.1/enable/component.py#L738-L775 (`self._backbuffer` is a context instance)
- `Scrollbar` has lots of `draw_image` calls, but appears to be otherwise broken since the `image_for` method which can create those images is nowhere to be found.

Some refactoring (and probably deprecations) will need to be done to make changes here possible.

rkern:
Can you clarify why you think there is a "hard dependency" on the current agg backend?

- The first link just goes away if we blow away the current `agg` backend, so that's no problem.

[...]

PR #669, opened 2021-03-03
Add an 'oldagg' backend pointing to kiva.agg

jwiggins:
This is step 3 from #414. Before switching the `image` kiva backend to celiagg, we need to make `kiva.agg` accessible via an unambiguous name.

### DIFF
--- a/enable/null/oldagg.py
+++ b/enable/null/oldagg.py
@@ -1,0 +1,31 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+from kiva.agg import CompiledPath, GraphicsContextArray as GraphicsContext
+
+__all__ = [
+    "CompiledPath", "GraphicsContext", "NativeScrollBar", "Window",
+    "font_metrics_provider"
+]
+
+
+class NativeScrollBar(object):
+    pass
+
+
+class Window(object):
+    pass
+
+
+def font_metrics_provider():
+    from kiva.api import Font
+
+    gc = GraphicsContext((1, 1))
+    gc.set_font(Font())
+    return gc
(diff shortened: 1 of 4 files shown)


corranwebster:
Do we want a `kiva/oldagg.py` as well?  For now it could just import everything from `kiva.image`.

jwiggins:
Come to think of it, we might...

jwiggins:
I'm not sure we're ready to "throw the switch" yet, so it's not being done in this PR

enable/null/oldagg.py
```diff
@@ -0,0 +1,18 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+from .image import (
```
aaronayres35, line 10: I believe this is similar to Rahul's question, but why do we pull from `.image` not `.agg` in these?

Because if we eventually have image point to celiagg, then oldagg effectively becomes an alias for celiagg not agg?

aaronayres35: > `.agg` doesn't exist.

I am still a little confused.  how does the code in `kiva/image.py` work then as it seems to just import from `.agg`, see: https://github.com/enthought/enable/blob/fe462b81d6b815b1ef5b673e80d9c956d87f0b39/kiva/image.py#L19

My understanding was that image was basically just an alias pointing at agg, and it would move to point to celiagg in the future (ie those imports change to `.celiagg`). 

If the image contents are going to eventually be moved into these oldagg modules when `.image` points to celiagg that makes sense.  But the image contents seem to be pulling from `.agg` so it would make sense to me if these files effectively just copied that code over now rather than later?

I believe I am still missing something

aaronayres35: ahhh I see, my mistake, that makes sense now
thank you!

rahulporuri (approved):
LGTM. IIUC, the next step (which we will do at somepoint in the future) is to move the contents of `*.image` into the `*.oldagg` modules and import `*.celiagg` stuff into the `*.image` modules. Is that correct John?

jwiggins:
That is correct

jwiggins:
Thanks for the reviews

PR #672, opened 2021-03-03
Document GraphicsContext state

### LINKED ISSUE #615: Kiva: Basic documentation of GC state
User-facing documentation about GC state:

- what the state holds (colors, line styles, transform, etc.)
- the state stack and how it works
- using `get_state`, `set_state`, `with gc:`
- explain why drawing code should wrap drawing in `with gc:` if it changes any state.

jwiggins:
Closes #615


docs/source/kiva/compiled_path_ex.py
```diff
@@ -17,4 +17,4 @@
     gc.add_path(path)
     gc.fill_path()
 
-gc.save("example.png")
+gc.save("compiled_path_ex.png")
```
aaronayres35, line 20: the changes to this file seem unrelated to the main purpose of this PR, but even so the changes make sense as something we want to do

jwiggins: You're right, it's not related. I had made a similar change in the new state example (which was copied from the compiled path example)

docs/source/kiva/state.rst
```diff
@@ -0,0 +1,83 @@
...
+
+After leaving the ``with`` block, the state is now restored to its default. A
+line is drawn from some point 20 units away from the origin -- which was
+displaced by the default scale and translate -- to a point which is next to the
+rectangle that was drawn a few lines above.
+
+.. literalinclude:: state_ex.py
+  :lines: 27-29
+  :linenos:
+  :lineno-match:
```
aaronayres35, line 83: This example showcases 2 approaches to drawing something with some transformation applied.  You can perform the transformations yourself to what you want to draw like what is done here for the lines, or you can transform the entire coordinate system and act as though you are drawing at the origin as is done in the with block.  
e.g in an analogy Corran made in the LnL if you want to write your name upside down either figure out how to write characters upside down, or turn the paper upside down write it, and flip the paper back.  

The example shows this well, but it might be good to re-enforce the notion with a short concluding paragraph.  This would highlight the benefit of GC state / using `with gc:`.  I think the analogy illustrates the idea of why it can be so powerful.  Additionally we could mention in some cases drawing needs to be done on a gc you don't own.  You draw your part, but don't want to mess with state once you are done, as you don't want your changes to mess with drawing done elsewhere.

aaronayres35 (approved):
LGTM, just a couple nit picky comments and a request for a quick "why this is useful/powerful" concluding paragraph

docs/source/kiva/state.rst
```diff
@@ -0,0 +1,83 @@
...
+* Stroke color (:py:meth:`set_stroke_color`, :py:meth:`get_stroke_color`)
+* Line width (:py:meth:`set_line_width`)
+* Line join style (:py:meth:`set_line_join`)
+* Line cap style (:py:meth:`set_line_cap`)
+* Line dashing (:py:meth:`set_line_dash`)
+* Global transparency (:py:meth:`set_alpha`, :py:meth:`get_alpha`)
+* Anti-aliasing (:py:meth:`set_antialias`, :py:meth:`get_antialias`)
+* Miter limit (:py:meth:`set_miter_limit`)
+* Flatness (:py:meth:`set_flatness`)
+* Image interpolation (:py:meth:`set_image_interpolation`, :py:meth:`get_image_interpolation`)
```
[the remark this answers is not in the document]
jwiggins: Looks like I missed clipping. As with all Kiva features, there are inconsistencies across backends. However, the _interface_ specifies these things as belonging to the state.

docs/source/kiva/state.rst
```diff
@@ -0,0 +1,83 @@
...
+current path and stroked.
+
+.. literalinclude:: state_ex.py
+  :lines: 18-25
+  :linenos:
+  :lineno-match:
+
+After leaving the ``with`` block, the state is now restored to its default. A
+line is drawn from some point 20 units away from the origin -- which was
+displaced by the default scale and translate -- to a point which is next to the
```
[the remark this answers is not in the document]
jwiggins: I disagree. I think it should be "translation"

docs/source/kiva/state.rst
jwiggins: This is a good point. The correct way to draw the lines would be with a transformation applied so that the polar coordinate conversion was unnecessary.

I'll add a bit more detail at the end. Your example of being passed a context that you don't own is excellent and super relevant here.

docs/source/kiva/state.rst
```diff
@@ -0,0 +1,83 @@
+Kiva State
+==========
+Kiva is a "stateful" drawing API. What this means is that the graphics context
+has a collection of state which affect the results of its drawing actions.
```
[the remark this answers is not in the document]
jwiggins: Fixed

aaronayres35 (approved):
LGTM

jwiggins:
Thanks for the feedback

PR #673, opened 2021-03-03
Clean up the benchmark suite code

jwiggins:
I alluded to this in my feedback on #667. This PR removes the deferred imports from `enable.gcbench.suite` and leaves a helpful note for any future coders who want to add an import there.

Also included is a drive-by change of the `show_text_radial_gradient` benchmark which shows that celiagg is doing it correctly (or at the very least how I expect):
celiagg show_text_radial_gradient_2x

### DIFF
--- a/enable/gcbench/suite.py
+++ b/enable/gcbench/suite.py
@@ -8,10 +8,23 @@
 #
 # Thanks for using Enthought open source!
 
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM

PR #674, opened 2021-03-03
Fix set_font() for PDF and SVG backends

jwiggins:
Font lookup was a bit screwy with all of the file-based Kiva backends. This fixes `set_font` for PDF and SVG, at least with respect to the benchmark program.

### DIFF
--- a/kiva/pdf.py
+++ b/kiva/pdf.py
@@ -19,12 +19,14 @@
 
 # standard library imports
 import copy
+import os
 import warnings
 
 from numpy import ndarray, pi
 
 # ReportLab PDF imports
-import reportlab.pdfbase.pdfmetrics
+from reportlab.pdfbase import pdfmetrics
+from reportlab.pdfbase.ttfonts import TTFont
 import reportlab.pdfbase._fontdata
 from reportlab.pdfgen import canvas
 
@@ -35,6 +37,7 @@ from .line_state import is_dashed
 from .constants import FILL, STROKE, EOF_FILL
 import kiva.constants as constants
 import kiva.affine as affine
+from kiva.fonttools.font import Font
 
 
 cap_style = {}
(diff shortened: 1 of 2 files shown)


kiva/pdf.py
```diff
@@ -621,7 +624,18 @@ def set_font(self, font):
...
+        try:
+            self.gc.setFont(face_name, font.size)
+        except KeyError:
+            # Face name is not recognized.
+            # Register the TTF font that kiva's font_manager located
+            filename = font.findfont()
+            if not os.path.splitext(filename)[-1] in (".ttf", ".ttc"):
+                raise
+
+            pdfmetrics.registerFont(TTFont(face_name, filename))
```
jwiggins, line 637: As far as I can tell, this embeds the font in the PDF file. I'm not sure if we want that or not.

PDF has a set of "safe" fonts which come with Acrobat Reader. I think maybe what we want to do in this method is try to map all requested fonts to the nearest match in this list.

aaronayres35: > As far as I can tell, this embeds the font in the PDF file.

Based off the link below, I believe this is true as well.

> PDF has a set of "safe" fonts which come with Acrobat Reader. I think maybe what we want to do in this method is try to map all requested fonts to the nearest match in this list.

I am unsure about this / why it would be necessary, but my understanding of all this is very much not complete / still growing.  When would embedding the font in the PDF file not work or be problematic?

ref: https://www.reportlab.com/docs/reportlab-userguide.pdf
pg. 28 and 49-52

aaronayres35 (approved):
LGTM

I re-ran the script and pdf/svg no longer fail for the `show_text` benchmark.  These are the results I see:

Screen Shot 2021-03-04 at 10 27 00 AM

Screen Shot 2021-03-04 at 10 26 51 AM

Admittedly I am not 100% confident in my understanding of all the details of the changes made.  Nonetheless, this PR certainly seems to be an incremental improvement, and from what I do understand the changes make sense.

kiva/pdf.py
jwiggins: > When would embedding the font in the PDF file not work or be problematic?

You're right. It's probably OK.

PR #675, opened 2021-03-03
Place Kiva docs before Enable docs

### LINKED ISSUE #603: Place Kiva docs before Enable docs?
Since `Kiva` is the lower level library, it might make sense to have the documentation set up so that the kiva docs appear first in the table of contents and then the `Enable` docs follow (enable builds on kiva so good to understand kiva first?).  I could however imagine arguments for leaving it as is. 

This is a very minor issue, but trivial to fix if we decide to do so.

jwiggins:
Closes #603


docs/source/index.rst
```diff
@@ -4,6 +4,12 @@ Enable Documentation
 .. toctree::
   :maxdepth: 2
 
+  kiva/overview.rst
+  kiva/backends.rst
+  kiva/quickref.rst
+  kiva/fonts.rst
+  kiva/compiled_path.rst
```
aaronayres35, line 11: while were shuffling things, maybe put `kiva/quickref.rst` last?

aaronayres35 (approved):
LGTM, one minor comment

docs/source/index.rst
jwiggins: OK

PR #676, opened 2021-03-03
kiva tutorial

### LINKED ISSUE #622: Kiva Tutorial
A tutorial on how to use Kiva to do drawing based around a specific goal (eg. rendering some type of diagram, eg. a simple circuit diagram).

aaronayres35:
closes #622 

Adds a tutorial for drawing this:
tutorial


jwiggins (commented):
This is off to a solid start.

What do you think about adding some structure to this so that components line resistors and meters are drawn by functions? The whole diagram drawing could then fit in a small "paragraph" of code which gives a high level overview of what is drawn and the functions contain the nitty-gritty details. That would also be a bit closer to how kiva code exists in the wild, at least in the context of Chaco tools and plot renderers.

docs/source/kiva_tutorial/tutorial.py
```diff
@@ -0,0 +1,142 @@
...
+
+# step 2) draw dots for wire connections
+points = np.array([
+    [200., 150.],
+    [200., 200.],
+    [400., 150.],
+    [400., 200.],
+    [550., 130.] 
+])
+gc.draw_marker_at_points(points, 4.0, CIRCLE_MARKER)
```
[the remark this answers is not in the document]
aaronayres35: I've tried to do this now in the `tutorial_advanced.py`
I left it there rn, but I think I will pull it in here too

docs/source/kiva_tutorial/tutorial.rst
```diff
@@ -0,0 +1,146 @@
...
+
+.. literalinclude:: tutorial.py
+    :lines: 13-17
+
+.. image:: images/step_1.png
+  :width: 600
+  :height: 400
+
+That was easy. Now, let's draw the dots indicating wire connections. To do
+this, we can use the optimized :py:meth:`draw_marker_at_points` method.
```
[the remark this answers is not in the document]
aaronayres35: I added a note mentioning this for now.  I agree we shouldn't be recommending it, so I think I will remove this entirely soon. 

In order to go with the flow of increasing complexity through the tutorial, maybe I will move this step a little later, and do it with `draw_path_at_points`?  Or maybe just do it in a for loop but do it right before the resistors where I use draw_path_at_points and use it as a kind of segue saying something like "some backends actually have optimized methods for doing this sort of thing as python for loops can be slow.  One of which is draw_path_at_points which we will now demonstrate to draw the resistors..."

docs/source/kiva_tutorial/tutorial.rst
```diff
@@ -0,0 +1,172 @@
...
+.. literalinclude:: tutorial.py
+    :lines: 70-77
+
+If you run just this code, you will notice that things don't look quite right,
+as the original lines for the wires still show underneath our resistors. We can
+get rid of these by drawing white lines over the relevant portions of the wire
+before we call the above code. Note that this is not the most performant
+approach we could take especially in the context of an interactive application.
+In general we want to strive to touch each pixel as few times as necessary.
+However, in this case to keep the tutorial simple, we can run
```
jwiggins, line 134: We can run ... ? Did you mean to put a colon at the end of this line?

docs/source/kiva_tutorial/tutorial.rst
```diff
@@ -0,0 +1,172 @@
...
+  :height: 400
+
+
+Now, for the switch. Just like above, we are going to want to "remove" the
+previously drawn wire. Then, for the actual switch, we want to effectively just
+rotate that segment of wire outwards. To do we tranform our coordinates so that
+we can simply draw a line along the x axis of the same length as the gap we
+just created. Thus, we translate to the lower edge of the switch, and rotate
+our axis to our desired angle. The beauty of the context manager then reverts
+our coordinate system back to as it was before.
```
jwiggins, line 150: "The beauty of the context manager then reverts" reads a bit strangely.

docs/source/kiva_tutorial/tutorial.rst
```diff
@@ -0,0 +1,172 @@
...
+
+.. literalinclude:: tutorial.py
+    :lines: 80-98
+
+.. image:: images/step_6.png
+  :width: 600
+  :height: 400
+
+We will leave the drawing of the battery as an exercise for the reader, but the
+full code for the example is available :download:`here <tutorial.py>`.
```
jwiggins, line 160: Here too, I suggest wrapping the code in a bare .rst file with a single `literalinclude` and then using `:ref:` instead of `:download:` for the linking.

docs/source/kiva_tutorial/tutorial.rst
```diff
@@ -0,0 +1,172 @@
...
+
+
+This tutorial was intended as a ramp up for drawing with kiva. Many of the
+approaches taken in the code we chosen with the idea of teaching, not
+optimizing code, in mind. In pactice, you probably would not want to implement
+drawing in this because it is not as performant as it could be. For example,
+
+Now that you better understand the basics of what
+kiva drawing is all about, please refer to the 
+:download:`advanced version of the tutorial  <tutorial_advanced.py>` to
```
jwiggins, line 170: Ditto. `:ref:`

jwiggins (approved):
🥇I'm really happy with this overall. Still have a bit of feedback but it's quite close now.

docs/source/kiva_tutorial/tutorial.py
```diff
@@ -0,0 +1,137 @@
...
+resistor_path.move_to(0,0)
+resistor_path_points = [(i*10+5, 10*(-1)**i) for i in range(8)]
+for x, y in resistor_path_points:
+    resistor_path.line_to(x,y)
+resistor_path.line_to(80, 0)
+gc.draw_path_at_points(resistor_locations, resistor_path, STROKE)
+gc.save("images/step_45.png")
+
+# step 6) switch
+# white out the wire
```
corranwebster, line 81: "Whiting out" like this will work on pixel-based backends, but may not work on postscript/pdf backend.  Check those to see what you get.

As a style guide, it's generally better to draw only what you want (using clipping, for example, to avoid drawing in certain areas).

docs/source/kiva_tutorial/tutorial.rst
```diff
@@ -0,0 +1,172 @@
+Introduction
+============
+
+This tutorial is intended to help you get up and running working with kiva.
+Kiva is a backend agnostic 2D vector drawing interface.  In other words, it is
```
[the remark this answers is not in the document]
corranwebster: I would suggest "Kiva".

docs/source/kiva_tutorial/tutorial.rst
```diff
@@ -0,0 +1,172 @@
...
+  :height: 400
+
+We will leave the drawing of the battery as an exercise for the reader, but the
+full code for the example is available :download:`here <tutorial.py>`.
+
+
+This tutorial was intended as a ramp up for drawing with kiva. Many of the
+approaches taken in the code we chosen with the idea of teaching, not
+optimizing code, in mind. In pactice, you probably would not want to implement
+drawing in this because it is not as performant as it could be. For example,
```
corranwebster, line 166: Missing sentence after "For example,"

corranwebster (approved):
Very nice!

A couple of minor nits from me, and you might go over it with an eye to where you might be able to tighten up the prose, but particularly the advanced example was pretty much exactly what I had in mind.

docs/source/index.rst
```diff
@@ -10,6 +10,7 @@ Enable Documentation
   kiva/compiled_path.rst
   kiva/state.rst
   kiva/quickref.rst
+  kiva_tutorial/index
```
corranwebster, line 13: For consistency with other links:
```suggestion
  kiva_tutorial/index.rst
```

docs/source/kiva_tutorial/tutorial.py
aaronayres35: That makes sense, I had only considered the performance issue john mentioned, not that it might not even work.  I don't think the example will work at all with postscript/pdf because we use `draw_marker_at_points`.  Nonetheless, it might be worth adding a sentence saying not only that we generally wouldn't want to do this for performance reasons, but also that it may not even produce correct results if using a different backend. 

I believe the `advanced_tutorial.py` should work, although I still need to try running it with different backends to make sure

jwiggins:
:shipit:

Issue #593, opened 2021-02-18
save() methods of GraphicsContext implementations should record DPI


jwiggins:
When saving images from high resolution `GraphicsContext` instances, it would be nice to use the `dpi` parameter of various [PIL file format writers](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html) to expose the expected display resolution of the image.

PR #677, opened 2021-03-04
Handle 2x benchmark rendering more elegantly

jwiggins:
Working on #593 led me to this more uniform output for the benchmark suite. All output images appear to be the same size now, but the 2x ones are just crisper. Additionally, the 1x images are completely covered by whatever gets drawn.

### DIFF
--- a/enable/gcbench/publish.py
+++ b/enable/gcbench/publish.py
@@ -187,7 +187,7 @@ def _format_output(result):
     link.
     """
     if result.output_format in (".png", ".svg"):
-        return _img(result.output)
+        return _img(result.output, *result.output_size)
     else:
         return _link(result.output, "download")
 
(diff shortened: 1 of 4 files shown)


enable/gcbench/publish.py
```diff
@@ -201,8 +201,8 @@ def _format_timing(result):
 
 
 # HTML utils
-def _img(src):
-    return f'<img src="{src}" />'
+def _img(src, width, height):
+    return f'<img src="{src}" width="{width}" height="{height}"/>'
```
jwiggins, line 205: HTML is doing all the work to make the 2x images appear normally.

aaronayres35: Maybe this / something similar should be a comment in the code?  It may be overkill though, so feel free to ignore

aaronayres35 (approved):
LGTM

I re-ran the script and everything looks nice 😄

jwiggins:
Thanks for the quick review

enable/gcbench/publish.py
jwiggins: Eh... I'm not that concerned.

PR #678, opened 2021-03-04
Supply DPI when saving images

### LINKED ISSUE #593: save() methods of GraphicsContext implementations should record DPI

jwiggins:
Fixes #593

### DIFF
--- a/kiva/agg/src/graphics_context.i
+++ b/kiva/agg/src/graphics_context.i
@@ -370,6 +370,7 @@ namespace kiva {
                 _swig_setattr(self, GraphicsContextArray, 'thisown2', 1)
 
                 self.bmp_array = ary
+                self.base_scale = base_pixel_scale
 
             def __del__(self, destroy=_agg.destroy_graphics_context):
                 try:
@@ -839,15 +840,26 @@ namespace kiva {
                 """
                 from PIL import Image
 
+                FmtsWithDpi = ('jpg', 'png', 'tiff', 'jpeg')
                 FmtsWithoutAlpha = ('jpg', 'bmp', 'eps', "jpeg")
                 size = (self.width(), self.height())
                 fmt = self.format()
+
+                if pil_options is None:
+                    pil_options = {}
+
+                file_ext = filename[-3:].lower() if isinstance(filename, str) else ''
+                if (file_ext in FmtsWithDpi or
+                        (file_format is not None and
+                         file_format.lower() in FmtsWithDpi)):
+                    dpi = int(72 * self.base_scale)
+                    pil_options["dpi"] = (dpi, dpi)
 
                 # determine the output pixel format and PIL format
                 if fmt.endswith("32"):
                     pilformat = "RGBA"
                     pixelformat = "rgba32"
-                    if (isinstance(filename, str) and filename[-3:].lower() in FmtsWithoutAlpha) or \
+                    if file_ext in FmtsWithoutAlpha or \
                        (file_format is not None and file_format.lower() in FmtsWithoutAlpha):
                         pilformat = "RGB"
                         pixelformat = "rgb24"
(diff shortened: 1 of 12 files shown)


kiva/agg/src/graphics_context.i
```diff
@@ -839,15 +840,26 @@ namespace kiva {
...
 
+                FmtsWithDpi = ('jpg', 'png', 'tiff', 'jpeg')
                 FmtsWithoutAlpha = ('jpg', 'bmp', 'eps', "jpeg")
                 size = (self.width(), self.height())
                 fmt = self.format()
 
+                if pil_options is None:
+                    pil_options = {}
+
+                file_ext = filename[-3:].lower() if isinstance(filename, str) else ''
```
aaronayres35, line 851: It seems we are following what was done for `FmtsWithoutAlpha` here, but we assume the extension is only 3 characters.  We could use `filename.rpartition(".")[-1].lower()` instead?  I guess we only need it for extensions in `FmtsWithDpi` / `FmtsWithoutAlpha` so not really needed...
Feel free to ignore

kiva/agg/src/graphics_context.i
```diff
@@ -839,15 +840,26 @@ namespace kiva {
...
+                if (file_ext in FmtsWithDpi or
+                        (file_format is not None and
+                         file_format.lower() in FmtsWithDpi)):
+                    dpi = int(72 * self.base_scale)
```
aaronayres35, line 855: where exactly does the 72 come from?  could it be wrapped in a variable just so it has a name?

kiva/celiagg.py
```diff
@@ -837,14 +839,23 @@ def save(self, filename, file_format=None):
...
-                (file_format.lower() in no_alpha_formats)):
+        if ext in no_alpha_formats or file_format.lower() in no_alpha_formats:
             img = img.convert('RGB')
 
-        img.save(filename, format=file_format)
+        # Check the output format to see if it can handle DPI
+        dpi_formats = ('jpg', 'png', 'tiff', 'jpeg')
+        if ext in dpi_formats or file_format.lower() in dpi_formats:
+            # Assume 72dpi is 1x
+            dpi = int(self.base_scale * 72)
```
aaronayres35, line 855: Ah I understand 72 now.  This same comment could be added in the previous file rather than wapping it in a variable.  Or both could be wrapped up with a variable

kiva/quartz/ABCGI.pyx
```diff
@@ -1609,18 +1613,30 @@ cdef class CGBitmapContext(CGContext):
...
+
+        file_ext = (
+            os.path.splitext(filename)[1][1:] if isinstance(filename, str)
+            else ''
+        )
+
+        # Check te output format to see if DPI can be passed
+        dpi_formats = ('jpg', 'png', 'tiff', 'jpeg')
+        if file_ext in dpi_formats or file_format.lower() in dpi_formats:
+            dpi = int(72 * self.base_scale)
```
aaronayres35, line 1627: same comment here

kiva/tests/test_agg_drawing.py
```diff
@@ -19,6 +19,9 @@ class TestAggDrawing(DrawingImageTester, unittest.TestCase):
     def create_graphics_context(self, width, height, pixel_scale):
         return GraphicsContext((width, height), base_pixel_scale=pixel_scale)
 
+    def test_save_dpi(self):
+        self.assertEqual(self.save_and_return_dpi(), 144)
```
aaronayres35, line 23: maybe a comment here / in the other tests about the 144

aaronayres35 (approved):
LGTM

Effectively just one comment about magic numbers

kiva/agg/src/graphics_context.i
jwiggins: Yeah, I copied the code as it was. That said, you version using `rpartition` should do just fine.

jwiggins:
Thanks for the feedback

PR #689, opened 2021-03-07
Add Brush classes

corranwebster:
These are classes intended to make it easier to paint and style filled regions.  This PR simply adds the classes and an example of usage.

Subsequent PRs will integrate this into existing code (eg. to add a background brush class to the base component).

### DIFF
--- a/enable/brush.py
+++ b/enable/brush.py
@@ -97,11 +97,47 @@ class ColorStop(HasStrictTraits):
         self.updated = True
 
 
+class Gradient(HasStrictTraits):
+    """ A color gradient. """
+
+    #: The sequence of color stops for the gradient.
+    stops = List(Instance(ColorStop), update=True)
+
+    #: A trait which fires when the gradient is updated.
+    updated = Event()
+
+    def to_array(self):
+        """ Return a sorted list of stop arrays.
+
+        This is the raw form of the stops required by Kiva.
+
+        Returns
+        -------
+        stop_array_list : arrays
+            A list of array of (offset, r, b, b, a) values corresponding to
+            the color stops.
+        """
+        return array([
+            stop.to_array()
+            for stop in sorted(self.stops, key=attrgetter("offset"))
+        ])
+
+    @observe("stops.items.updated")
+    def observe_stops(self, event):
+        self.updated = True
+
+    def _stops_default(self):
+        return [
+            ColorStop(offset=0.0, color="white"),
+            ColorStop(offset=0.0, color="black"),
+        ]
+
+
 class GradientBrush(Brush):
     """ A brush that paints a color gradient. """
 
     #: The sequence of color stops for the gradient.
-    stops = List(Instance(ColorStop), update=True)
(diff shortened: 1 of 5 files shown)


enable/brush.py
```diff
@@ -0,0 +1,224 @@
...
+        ])
+
+    @observe("stops.items.updated")
+    def observe_stops(self, event):
+        self.updated = True
+
+    def _stops_default(self):
+        return [
+            ColorStop(offset=0.0, color="white"),
+            ColorStop(offset=0.0, color="black"),
```
jwiggins, line 132: Should be offsets 0 and 1, right?

corranwebster:
There's bits of the gradient example which could be pulled out and either added to the gradient object itself, or turned into more general things.  But for now an example is good enough.

enable/examples/demo/enable/gradient_demo.py
```diff
@@ -0,0 +1,384 @@
...
+    """
+
+    #: The selected color stop.
+    model = Instance(ColorStop)
+
+    #: The attribute to link to the drag.
+    x_attr = "offset",
+
+    #: The bounds on the valus of the attribute.
+    x_bounds = (0.0, 1.0),
```
jwiggins, line 324: Tuples that should not be tuples.

enable/brush.py
```diff
@@ -0,0 +1,224 @@
...
+
+        Parameters
+        ----------
+        gc : graphics context
+            The graphics context to use the brush with.
+        """
+        gc.linear_gradient(
+            *self.start,
+            *self.end,
+            self.gradient.to_array(),
```
jwiggins, line 189: What do you think about caching this array, so that it doesn't have to be built every time the gradient is used?

enable/brush.py
```diff
@@ -0,0 +1,224 @@
...
+    @observe("+update")
+    def observe_update_traits(self, event):
+        self.updated = True
+
+
+class ColorBrush(Brush):
+    """ A simple brush that paints a solid color. """
+
+    #: The color to brush the region with.
+    color = transparent_color_trait(update=True)
```
jwiggins, line 56: I feel like `ColorTrait("transparent", update=True)` would read better here. `transparent_color_trait` seems like something that can't have its color reassigned.

jwiggins (approved):
Couple comments, but this is otherwise 👍

enable/examples/demo/enable/gradient_demo.py
```diff
@@ -0,0 +1,384 @@
...
+    """ A tool that handles dragging color stops on a gradient.
+
+    The first and last color stops can't be dragged.
+    """
+
+    #: The selected color stop.
+    model = Instance(ColorStop)
+
+    #: The attribute to link to the drag.
+    x_attr = "offset",
```
jwiggins, line 321: The trailing comma causes this demo to crash

corranwebster:
Assuming that tests pass, this is good to go.

jwiggins (approved):

PR #691, opened 2021-03-08
Normalize some signatures of Kiva methods

jwiggins:
This is part of an ongoing effort to address #648

* Changes to method signatures were made in `AbstractGraphicsContext` to reflect how a majority of backends have something implemented (gradients, `show_text`, `draw_path_at_points`, `set_line_dash`).
* `show_text` with the `point=` kwarg was refactored for many of the lesser-used backends.
* `select_font` now has the correct signature on celiagg, PDF, and Quartz backends.
* QPainter uses the fill color for text now, matching the behavior of AGG

### DIFF
--- a/kiva/pdf.py
+++ b/kiva/pdf.py
@@ -656,20 +656,6 @@ class GraphicsContext(GraphicsContextBase):
         font = self.gc._fontname
         self.gc.setFont(font, size)
 
-    def set_character_spacing(self):
-        """
-        """
-        pass
-
-    def get_character_spacing(self):
-        """ Get the current font """
-        raise NotImplementedError
-
-    def set_text_drawing_mode(self):
-        """
-        """
-        pass
-
     def set_text_position(self, x, y):
         """
         """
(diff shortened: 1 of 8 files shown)


kiva/qpainter.py
```diff
@@ -690,10 +696,14 @@ def show_text(self, text, point=None):
...
 
-        self.gc.save()
-        self.gc.setTransform(unflip_trans, True)
-        self.gc.drawText(QtCore.QPointF(pos[0], self._flip_y(pos[1])), text)
-        self.gc.restore()
+        # Make some temporary modifications to the state
+        with self:
+            # Kiva uses the fill color for text
+            brush = self.gc.brush()
+            self.gc.setPen(brush.color())
```
aaronayres35, line 703: verified this with the benchmark 👍

kiva/pdf.py
```diff
@@ -647,20 +656,6 @@ def set_font_size(self, size):
...
-        pass
-
-    def get_character_spacing(self):
-        """ Get the current font """
-        raise NotImplementedError
-
-    def set_text_drawing_mode(self):
-        """
-        """
-        pass
```
aaronayres35, line 662: Just to clarify, why did these get removed?

aaronayres35 (approved):
LGTM

are there any documentation updates that need to happen?  It looks like we use `automethod` now so I don't think there should.
Looking at that though I noticed we still have some old doc files in `docs/kiva/agg`.  We can probably delete these?  eg https://github.com/enthought/enable/blob/master/docs/kiva/agg/interface.txt

jwiggins:
Documentation is covered by `:automethod:`. Basically, these changes are bringing the backend implementations in line with what's specified by that documentation.

PR #692, opened 2021-03-08
Remove old/broken code

### LINKED ISSUE #685: Remove enable.control and related modules
The `enable.control` module has never worked in the available history of the codebase (due to bad imports).  It should be removed.  Similarly for `enable.scrollbar`.

Removing `enable.scrollbar` means that `enable.scroll_handler` can probably be removed as well.

There are old, working examples of this code (for Python ~2.5-ish values of "working") deep in Enthought's historical codebase if they ever need to be revived, but it may be less effort just to re-write from scratch.

Edit: I think `enable.drag` and `enable.drag_resize` which provide `DragHandler` and `DragResizeHandler` also fall into this category.

Edit again: also `enable.render_controller`?

Edit once more: also `enable.radio_group`

aaronayres35:
closes #685

This PR removes modules which are not used anywhere / have been broken for some time as described in the issue.  It also removes a few sections of commented out code.

None of these removals should cause any other code to break that wasn't already broken.


rahulporuri (approved):
LGTM. Let's give folks until the end of the day to complain if they don't want this PR to get merged.

rahulporuri:
I've searched the enthought org in github for all of the removed modules - and i don't see any *working uses (*working because there are older imports from the enthought namespace). I get a lot of false positives when I try to search all code in github.

I think it's safe to merge this PR and release the changes in the upcoming release - I dont expect any internal projects to have issues - and I don't expect external users to have issues either.

Issue #555, opened 2021-01-28
API for adding application-specific fonts


rkern:
There is a need for an application to provide one or more fonts that are to be used in the application, particularly for Chaco labels. For example, supporting CJK languages across platforms can benefit by providing a single suitably-licensed font and specifying that the appropriate `preferred_fonts` category uses it.

One can currently hack through the implementation details of `kiva.fonttools.font_manager` to accomplish this, but there is a need for a blessed, documented API to accomplish this.

corranwebster:
This will probably require more than just modifying the `kiva.fonttools` code, as a most of the Kiva backends just take the name of the font and then use their own internal font discovery mechanisms to resolve it (eg. `QPainter` backend will use Qt font discovery).

jwiggins:
It might be worth checking if [`QFontDatabase.addApplicationFont`](https://doc.qt.io/qt-5/qfontdatabase.html#addApplicationFont) can help here.

jwiggins:
@rahulporuri We should consider adding this to 5.1.0

jwiggins:
`QFontDatabase.addApplicationFont` works as expected with the QPainter kiva backend. As was already pointed out, there will need to be matching changes in the `kiva.fonttools` code in order to work with other backends.

jwiggins:
[`wx.Font.AddPrivateFont`](https://docs.wxpython.org/wx.Font.html#wx.Font.AddPrivateFont) exists as well. It has some weird caveats, but since Wx support is supplied on a best effort basis...

PR #693, opened 2021-03-09
Move the font scanning code into private modules

jwiggins:
NOTE: This does not _yet_ modify the behavior of `FontManager`.

This PR is intended to isolate the font file scanning code currently in `kiva.fonttools.font_manager` and cover it with tests. In a future PR, we'll actually start using this code.

Tangentially related to #391 and #555

### DIFF
--- a/kiva/fonttools/_scanner.py
+++ b/kiva/fonttools/_scanner.py
@@ -1,0 +1,538 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+"""
+####### NOTE #######
+This is based heavily on matplotlib's font_manager.py rev 8713,
+but has been modified to not use other matplotlib modules
+####################
+
+Authors   : John Hunter <jdhunter@ace.bsd.uchicago.edu>
+            Paul Barrett <Barrett@STScI.Edu>
+            Michael Droettboom <mdroe@STScI.edu>
+Copyright : John Hunter (2004,2005), Paul Barrett (2004,2005)
+License   : matplotlib license (PSF compatible)
+            The font directory code is from ttfquery,
+            see license/LICENSE_TTFQUERY.
+"""
+import glob
+import logging
+import os
+import subprocess
+import sys
+
+from fontTools.ttLib import TTCollection, TTFont, TTLibError
+
+from kiva.fonttools import afm
+from kiva.fonttools._constants import weight_dict
+from kiva.fonttools._util import get_ttf_prop_dict, weight_as_number
+
+logger = logging.getLogger(__name__)
+# Error message when fonts fail to load
+_FONT_ENTRY_ERR_MSG = "Could not convert font to FontEntry for file %s"
+
+# OS Font paths
+MSFolders = r"Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders"
+MSFontDirectories = [
+    r"SOFTWARE\Microsoft\Windows NT\CurrentVersion\Fonts",
+    r"SOFTWARE\Microsoft\Windows\CurrentVersion\Fonts",
+]
+OSXFontDirectories = [
+    "/Library/Fonts/",
+    "/Network/Library/Fonts/",
+    "/System/Library/Fonts/",
+]
+X11FontDirectories = [
+    # an old standard installation point
+    "/usr/X11R6/lib/X11/fonts/TTF/",
+    # here is the new standard location for fonts
+    "/usr/share/fonts/",
+    # documented as a good place to install new fonts
+    "/usr/local/share/fonts/",
+    # common application, not really useful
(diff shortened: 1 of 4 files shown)


kiva/fonttools/_scanner.py
```diff
@@ -0,0 +1,538 @@
...
+    OSXFontDirectories.append(path)
+    path = os.path.join(home, ".fonts")
+    X11FontDirectories.append(path)
+
+
+class FontEntry(object):
+    """ A class for storing Font properties. It is used when populating
+    the font lookup dictionary.
+    """
+    def __init__(self, fname="", name="", style="normal", variant="normal",
```
rahulporuri, line 75: ~not sure what the difference between `fname` and `name` is.~ looking at the rest of the class, it's clear that `fname` is the filepath and `name` is the font name. Would it be possible to rename `fname` to `filepath` or will that break backwards compatibility?

kiva/fonttools/_scanner.py
```diff
@@ -0,0 +1,538 @@
...
+
+
+def create_font_list(fontfiles, fontext="ttf"):
+    """ Creates a list of :class`FontEntry` instances from a list of provided
+    filepaths.
+
+    The default is to create a list of TrueType fonts. An AFM font list can
+    optionally be created.
+    """
+    # FIXME: This function is particularly difficult to debug
```
rahulporuri, line 103: Is this comment still valid? If so, why? It mighe be useful to add a bit more detail here as to why debugging it is difficult.

kiva/fonttools/_scanner.py
```diff
@@ -0,0 +1,538 @@
...
+
+    def __repr__(self):
+        fname = os.path.basename(self.fname)
+        return (
+            f"<FontEntry '{self.name}' ({fname}) {self.style} {self.variant} "
+            f"{self.weight} {self.stretch}>"
+        )
+
+
+def create_font_list(fontfiles, fontext="ttf"):
```
rahulporuri, line 96: Is it necessary to specify that this function only handles `fontext` - `ttf` and `afm`? ~Or does it also implicitly handle `ttc` font files?~ looking at more code in this module, looks like `ttf` and `ttc` are handled by the same function

kiva/fonttools/_scanner.py
```diff
@@ -0,0 +1,538 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
```
rahulporuri, line 1: correct me if im wrong but it looks like this module has two distinct functions - `FontEntry` and the code to create `FontEntry` instances from font files and the code to scan system fonts.

Is it overkill to split these two functions into separate modules?

kiva/fonttools/_scanner.py
```diff
@@ -0,0 +1,538 @@
...
+    #  Styles are: italic, oblique, and normal (default)
+    if font.get_angle() != 0 or name.lower().find("italic") >= 0:
+        style = "italic"
+    elif name.lower().find("oblique") >= 0:
+        style = "oblique"
+    else:
+        style = "normal"
+
+    #  Variants are: small-caps and normal (default)
+    # !!!!  Untested
```
rahulporuri, line 432: Do you think it makes sense to replace `!!!!` with the usual `FIXME` or `TODO`?

kiva/fonttools/tests/test_scanner.py
```diff
@@ -0,0 +1,304 @@
...
+
+from .._scanner import (
+    _afm_font_property, _build_afm_entries, _ttf_font_property,
+    create_font_list, FontEntry, scan_system_fonts
+)
+
+data_dir = resource_filename("kiva.fonttools.tests", "data")
+is_macos = (sys.platform == "darwin")
+is_windows = (sys.platform in ("win32", "cygwin"))
+is_generic = not (is_macos or is_windows)
```
rahulporuri, line 26: im not entirely sure what's a generic platform - does generic here mean linux?

rahulporuri (approved):
Mostly LGTM with a few comments/questions.

kiva/fonttools/_scanner.py
jwiggins: 🤷‍♂️ This one is probably OK. I don't think any external code is dealing directly with `FontEntry` instances.

kiva/fonttools/_scanner.py
jwiggins: I'll look into this, but I won't do it in this PR.

kiva/fonttools/tests/test_scanner.py
jwiggins: Generic unix? Anything not macOS or Windows with X11, basically. That could be a lot of platforms, but in practice it's linux and freebsd (and solaris?).

kiva/fonttools/_scanner.py
```diff
@@ -0,0 +1,538 @@
...
+but has been modified to not use other matplotlib modules
+####################
+
+Authors   : John Hunter <jdhunter@ace.bsd.uchicago.edu>
+            Paul Barrett <Barrett@STScI.Edu>
+            Michael Droettboom <mdroe@STScI.edu>
+Copyright : John Hunter (2004,2005), Paul Barrett (2004,2005)
+License   : matplotlib license (PSF compatible)
+            The font directory code is from ttfquery,
+            see license/LICENSE_TTFQUERY.
```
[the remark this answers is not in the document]
jwiggins: I can't find that license in the matplotlib repo...

PR #699, opened 2021-03-10
Remove kiva.fonttools.sstruct

### LINKED ISSUE #694: Remove kiva.fonttools.sstruct
I can't find any code which actually uses the `kiva.fonttools.sstruct` module. It should probably be removed.

jwiggins:
Closes #694. There is plenty of explanation for why this is OK in the issue (basically it was only there for the long gone vendored copy of fonttools).


rahulporuri (approved):
:shipit:

jwiggins:
Oops. Forgot that this module had associated tests.

PR #701, opened 2021-03-10
Split FontProperties out into its own module

jwiggins:
More font manager reorganization. This is the last big chunk to be copied out of `kiva.fonttools.font_manager`. The next step is to delete all the now duplicated code.


aaronayres35 (approved):
LGTM

Hooray for new tests 🎉 😎

PR #702, opened 2021-03-10
Move some constants not yet moved

jwiggins:
These constants weren't moved in #701, so doing it here will reduce the noise of the upcoming code purge PR.

### DIFF
--- a/kiva/fonttools/_constants.py
+++ b/kiva/fonttools/_constants.py
@@ -8,6 +8,17 @@
 #
 # Thanks for using Enthought open source!
 
+font_family_aliases = {
+    "serif",
+    "sans-serif",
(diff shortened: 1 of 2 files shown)


aaronayres35 (approved):
LGTM

It is effectively inconsequential, but the other constants were left in `font_manager.py` when they were moved over.  Was there a specific reason for removing these in this PR and not the others in the previous PR?  Not a suggested change or anything, just curious to make sure I'm not missing something

jwiggins:
The reason is that I have a bunch of parallel changes in flight (all sourced from a branch with a huge diff). The other constants were being moved in #701. I wanted to reduce the size of the big code deletion PR that's coming up.

PR #703, opened 2021-03-10
Ignore kiva backends in the API docs

### LINKED ISSUE #696: Ignore kiva backends in kiva API docs
`kiva.abstract_graphics_context.AbstractGraphicsContext` is what the users should be relying on instead of relying on the specific `GraphicsContext` defined in the various kiva backends e.g. `cairo`, `celiagg`, `image`, etc.

This is related to #648 which aims to ensure that all backends faithfully implement the `AbstractGraphicsContext` interface.

Note that we should point it out in the documentation - somewhere in the kiva documentation - that the `AbstractGraphicsContext` interface is what the users should rely on.

jwiggins:
Fixes #696

I also added `:noindex:` directives to all the methods in `kiva/quickref.rst` because the duplication was causing a bunch of warnings.

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -246,16 +246,31 @@ def docs(runtime, toolkit, environment):
         "sphinx",
         "enthought_sphinx_theme",
     ])
-    ignore = " ".join([
+    enable_ignore = " ".join([
+        "enable/gcbench",
         "enable/null",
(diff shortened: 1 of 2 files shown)


aaronayres35 (approved):
LGTM

I see a huge drop in the number of warnings with this PR :)

PR #700, opened 2021-03-10
Use fonttools.afmLib instead of ad-hoc AFM parser

jwiggins:
`kiva.fonttools.afm` should not exist. Once all this font chaos is done, it won't.

This PR starts the ball rolling by changing the AFM parsing in the new code to use `fonttools.afmLib:AFM`. While I was at it, I copied `fonttools` test AFM file. Now we're actually exercising the code with some real data!

PR #704, opened 2021-03-11
Break out the font scoring methods and test them

jwiggins:
Ok, I'm pretty sure that along with #700, I'm really done breaking things out of `kiva.fonttools.font_manager`. The next PR is the big one.

This PR moves the font scoring methods to their own module and covers them _liberally_ with tests. A bug in `score_size` was fixed (trying to call a `dict` instance).

The font scoring methods are used by `FontManager.findfont` to compare `FontEntry` instances with `FontProperties` instances to determine their similarity.

### DIFF
--- a/kiva/fonttools/_score.py
+++ b/kiva/fonttools/_score.py
@@ -1,0 +1,135 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+"""
+####### NOTE #######
+This is based heavily on matplotlib's font_manager.py SVN rev 8713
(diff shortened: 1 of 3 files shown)


jwiggins:
I went ahead and actually deleted the old methods from `FontManager`.

kiva/fonttools/_score.py
```diff
@@ -0,0 +1,135 @@
...
+but has been modified quite a bit in the decade since it was copied.
+####################
+"""
+from kiva.fonttools._constants import (
+    font_family_aliases, font_scalings, preferred_fonts, stretch_dict,
+    weight_dict
+)
+
+# Each of the scoring functions below should return a value between
+# 0.0 (perfect match) and 1.0 (terrible match)
```
rahulporuri, line 23: bikeshed : that seems backwards to me.

rahulporuri (approved):
LGTM.

I did a quick search through the enthought codebases on GitHub to see if there are any uses of the methods - and I don't see any. None whatsoever. Searching in general on GitHub raises a lot of false negatives - uses of the method on the matplotlib font manager.

I'm confident that removing the methods won't cause much (if any) pain.

jwiggins:
Thanks for the review!

PR #705, opened 2021-03-11
Add tests for str_to_font

jwiggins:
Checking the coverage before and after my refactor, I saw that `str_to_font` was not covered (at least when looking at `kiva.fonttools` in isolation). This PR corrects that.


rahulporuri (approved):
LGTM

PR #711, opened 2021-03-11
Add a way for applications to supply custom fonts

### LINKED ISSUE #555: API for adding application-specific fonts

jwiggins:
Closes #555

This works well with the QPainter and AGG backends. It does _not_ work with the Quartz backend. Also, I was not able to test this properly with Wx because the feature isn't supported in the wheel that I installed (perhaps it's a macOS thing?).

### DIFF
--- a/kiva/api.py
+++ b/kiva/api.py
@@ -97,6 +97,7 @@ Fonts
 =====
 
 - :class:`~.Font`
+- :func:`~.add_application_fonts`
 
 Font Constants
 --------------
@@ -146,4 +147,4 @@ from .constants import (
     INVERTED_TRIANGLE_MARKER, PLUS_MARKER, DOT_MARKER, PIXEL_MARKER
 )
 from ._cython_speedups import points_in_polygon
-from .fonttools import add_application_font, Font
+from .fonttools import add_application_fonts, Font
--- a/kiva/api.py
+++ b/kiva/api.py
@@ -146,4 +146,4 @@ from .constants import (
     INVERTED_TRIANGLE_MARKER, PLUS_MARKER, DOT_MARKER, PIXEL_MARKER
 )
 from ._cython_speedups import points_in_polygon
-from .fonttools import Font
+from .fonttools import add_application_font, Font
--- a/kiva/fonttools/_scan_sys.py
+++ b/kiva/fonttools/_scan_sys.py
@@ -109,6 +109,38 @@ def scan_system_fonts(fontpaths=None, fontext="ttf"):
     return [fname for fname in fontfiles if os.path.exists(fname)]
 
 
+def scan_user_fonts(fontpaths=None, fontext="ttf"):
+    """ Search for fonts in the specified font paths.
+    """
+    if fontpaths is None:
+        return []
+
+    if isinstance(fontpaths, str):
+        fontpaths = [fontpaths]
+
+    fontfiles = set()
+    fontexts = _get_fontext_synonyms(fontext)
+    for path in fontpaths:
+        path = os.path.abspath(path)
+        if os.path.isdir(path):
+            # For directories, find all the fonts within
+            files = []
+            for ext in fontexts:
+                files.extend(glob.glob(os.path.join(path, "*." + ext)))
+                files.extend(glob.glob(os.path.join(path, "*." + ext.upper())))
+
+            for fname in files:
+                if os.path.exists(fname) and not os.path.isdir(fname):
+                    fontfiles.add(fname)
+        elif os.path.exists(path):
+            # For files, make sure they have the correct extension
+            ext = os.path.splitext(path)[-1][1:].lower()
+            if ext in fontexts:
+                fontfiles.add(path)
+
+    return sorted(fontfiles)
+
+
 # ----------------------------------------------------------------------------
 # utility funcs
 
--- a/kiva/fonttools/_scan_sys.py
+++ b/kiva/fonttools/_scan_sys.py
@@ -111,6 +111,11 @@ def scan_system_fonts(fontpaths=None, fontext="ttf"):
 
 def scan_user_fonts(fontpaths=None, fontext="ttf"):
     """ Search for fonts in the specified font paths.
+
+    Returns
+    -------
+    filepaths : list of str
+        A list of unique font file paths.
     """
     if fontpaths is None:
         return []
--- a/kiva/fonttools/font_manager.py
+++ b/kiva/fonttools/font_manager.py
@@ -137,17 +137,23 @@ class FontManager:
         """
         self.__default_weight = weight
 
-    def update_fonts(self, filenames):
+    def update_fonts(self, paths):
         """ Update the font lists with new font files.
+
+        The specified ``paths`` will be searched for valid font files and those
+        files will have their fonts added to internal collections searched by
+        :meth:`findfont`.
+
+        Parameters
+        ----------
+        filenames : list of str
+            A list of font file paths or directory paths.
         """
-        afm_paths = scan_user_fonts(filenames, fontext="afm")
-        ttf_paths = scan_user_fonts(filenames, fontext="ttf")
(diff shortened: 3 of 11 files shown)


jwiggins:
Perhaps this: `QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-runner'` has something to do with the failure?

jwiggins:
Works just fine on Windows 🎉

jwiggins:
From the Qt docs:
> Adding application fonts on Unix/X11 platforms without fontconfig is currently not supported.

Also, `QFontDatabase::addApplicationFont` is a static method.

jwiggins:
macOS + pyqt appears to have crashed when running the OpenGL drawing tests... Sort of an improvement. The tests are passing for the new feature.

rahulporuri:
I've been trying to experiment with the new `add_application_font` api ~but i'm having trouble finding an actual font file i can download and make use of~ - i'm not sure what the font name in the test font files is which are used in the tests.

Update : I came across the Noto fonts (https://web.archive.org/web/20150411172453if_/http://www.google.com/get/noto/) while browsing the Microsoft Font Typography Vendor page (https://docs.microsoft.com/en-gb/typography/font-list/). I was able to successfully add it in the `kiva_explorer` using 

```
fontfilepath = "C:\\Users\\rporuri\\Downloads\\NotoSans-hinted\\NotoSans-Regular.ttf"
add_application_font(fontfilepath)
fontfilepath = "C:\\Users\\rporuri\\Downloads\\NotoSans-hinted\\NotoSerif-Regular.ttf"
add_application_font(fontfilepath)
```

but then when I tried using the font via `gc.set_font(Font('Noto', size=42))`, I get nothing. I don't see errors specific to the Noto font and I just see the default font used in the kiva explorer application.

kiva/api.py
```diff
@@ -146,4 +146,4 @@
     INVERTED_TRIANGLE_MARKER, PLUS_MARKER, DOT_MARKER, PIXEL_MARKER
 )
 from ._cython_speedups import points_in_polygon
-from .fonttools import Font
+from .fonttools import add_application_font, Font
```
rahulporuri, line 149: can you add this to the module docstrings as well?

kiva/fonttools/_scan_sys.py
```diff
@@ -109,6 +109,38 @@ def scan_system_fonts(fontpaths=None, fontext="ttf"):
...
+            for fname in files:
+                if os.path.exists(fname) and not os.path.isdir(fname):
+                    fontfiles.add(fname)
+        elif os.path.exists(path):
+            # For files, make sure they have the correct extension
+            ext = os.path.splitext(path)[-1][1:].lower()
+            if ext in fontexts:
+                fontfiles.add(path)
+
+    return sorted(fontfiles)
```
rahulporuri, line 141: I didn't know that sorting a set will return a list - in case anyone was wondering if it'll be a problem that we're returning an empty list above but a sorted set here.

kiva/fonttools/_scan_sys.py
```diff
@@ -109,6 +109,38 @@ def scan_system_fonts(fontpaths=None, fontext="ttf"):
     return [fname for fname in fontfiles if os.path.exists(fname)]
 
 
+def scan_user_fonts(fontpaths=None, fontext="ttf"):
+    """ Search for fonts in the specified font paths.
```
rahulporuri, line 113: It might be useful to specify in the docstring that this returns a list of unique font files for the given path.

kiva/fonttools/font_manager.py
```diff
@@ -139,10 +139,15 @@ def set_default_weight(self, weight):
...
-        #  !!!!  Needs implementing
-        raise NotImplementedError
+        afm_paths = scan_user_fonts(filenames, fontext="afm")
+        ttf_paths = scan_user_fonts(filenames, fontext="ttf")
+
+        self.afmlist.extend(create_font_list(afm_paths))
+        self.ttflist.extend(create_font_list(ttf_paths))
+
+        self.ttf_lookup_cache = {}
+        self.afm_lookup_cache = {}
```
rahulporuri, line 150: Just to confirm, we're resetting these because these get used in the `findfont` method - and if we don't reset them, we end up not being able to discover the newly added font files. Right?

kiva/fonttools/font_manager.py
```diff
@@ -139,10 +139,15 @@ def set_default_weight(self, weight):
 
     def update_fonts(self, filenames):
```
rahulporuri, line 140: Technically, this accepts filenames and file directories because `scan_user_fonts` supports files and dirs. Do we want to make that explicit in the method docstring?

rahulporuri (approved):
LGTM with a few questions/comments. I tested this in the kiva explorer with a font I downloaded and things worked as expected.

There's one minor user improvement I request. When I downloaded the Font, I downloaded a zip archive which contained 4 font files - one for each style i.e. regular, bold, bold italic and italic. I needed to add each font file separately to be able to use `style` kwarg to `Font` correctly. Do you think it makes sense to support passing directories to `add_application_font`?

We should also update the user documentation - but we can also do that in a separate PR if you prefer that.

kiva/fonttools/font_manager.py
jwiggins: That's right.

rahulporuri:
I tried using the kiva explorer on wx (because i wanted to try this PR on wx) and I ran into a few issues across ETS - https://github.com/enthought/pyface/issues/901, https://github.com/enthought/traitsui/issues/1535 and https://github.com/enthought/enable/issues/713.

rahulporuri:
After fixing issues with the ETS libraries (traitsui and pyface) underneath, I was actually able to get this working with the wx toolkit - and it works perfectly 🎉

jwiggins:
The macOS test is calling `abort()` from some PyObjC code...

```
Fatal Python error: Aborted

Current thread 0x0000000109bbddc0 (most recent call first):
  File "/Users/runner/work/enable/enable/.edm/envs/enable-test-3.6-pyqt/lib/python3.6/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 775 in __call__
  File "/Users/runner/work/enable/enable/.edm/envs/enable-test-3.6-pyqt/lib/python3.6/site-packages/pyglet/libs/darwin/cocoapy/runtime.py", line 805 in __call__
  File "/Users/runner/work/enable/enable/.edm/envs/enable-test-3.6-pyqt/lib/python3.6/site-packages/pyglet/window/cocoa/__init__.py", line 314 in dispatch_events
  File "/Users/runner/work/enable/enable/.edm/envs/enable-test-3.6-pyqt/lib/python3.6/site-packages/kiva/tests/test_gl_drawing.py", line 76 in draw_and_check
...
```
* https://github.com/enthought/enable/blob/master/kiva/tests/test_gl_drawing.py#L76
* https://github.com/pyglet/pyglet/blob/v1.5.15/pyglet/window/cocoa/__init__.py#L314
* https://github.com/pyglet/pyglet/blob/v1.5.15/pyglet/libs/darwin/cocoapy/runtime.py#L769-L805

kiva/fonttools/font_manager.py
jwiggins: Actually this isn't needed.

kiva/fonttools/_scan_sys.py
jwiggins: Done

jwiggins:
I've disabled the GL drawing test on everything but Linux now. CI is finally happy (and I still hate the GL backend).

PR #712, opened 2021-03-11
Add new on_drag_leave Enum trait to DragTool

### LINKED ISSUE #551: DragTool.end_drag_on_leave does cancel, not drag_end
The `end_on_drag_leave` flag for the `DragTool` is a bit misleading because it in fact calls `_cancel_drag` (and subsequently `drag_cancel`), rather than exiting cleanly via `drag_end`. IMO, it'd be preferable if users of DragTool could choose whether a mouse leave during a drag cancels, ends, or does nothing.

aaronayres35:
fixes #551 

It seems like this fix is going to require a change in behavior and may potentially "break" existing code.

Currently the `end_drag_on_leave` flag does not actually lead to a drag being "ended" on a mouse leave.  Instead it is canceled.  My thoughts were to either add a `cancel_drag_on_leave` trait to do what `end_drag_on_leave` currently does (ie cancel the drag) and change `end_drag_on_leave` to actually end (see first commit). However, this requires one of these having preference over the other if both are True, and in actuality you would only ever want one behavior, so I removed `end_drag_on_leave` and made a new `on_drag_leave` enum trait.

This however is a change in the public api so we need to tread very carefully.  I am unsure what exactly we want to go with here, and thus am leaving this as a WIP for now awaiting further discussion.

I have tested manually using this example https://github.com/enthought/chaco/blob/master/examples/demo/edit_line.py.
I will write unit test(s) once the approach has been decided. CI will fail as I have not yet updated the existing tests.  Opened the PR early so other could see and discuss

### DIFF
--- a/enable/tests/test_drag_tool.py
+++ b/enable/tests/test_drag_tool.py
@@ -8,6 +8,7 @@
 #
 # Thanks for using Enthought open source!
 import unittest
+import warnings
 
 from traits.api import Int
 
@@ -74,7 +75,9 @@ class DragToolTestCase(EnableTestAssistant, unittest.TestCase):
         tool = self.tool
         tool.end_drag_on_leave = True
         tool._drag_state = "dragging"  # force dragging state
-        event = self.mouse_leave(interactor=tool, x=0, y=0)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            event = self.mouse_leave(interactor=tool, x=0, y=0)
         self.assertEqual(tool.canceled, 1)
         self.assertEqual(tool._drag_state, "nondrag")
         self.assertTrue(event.handled)
@@ -116,7 +119,9 @@ class DragToolTestCase(EnableTestAssistant, unittest.TestCase):
         tool.end_drag_on_leave = True
         tool.on_drag_leave = 'end'
         tool._drag_state = "dragging"  # force dragging state
-        event = self.mouse_leave(interactor=tool, x=0, y=0)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            event = self.mouse_leave(interactor=tool, x=0, y=0)
 
         # end_drag_on_leave should be handled like normal
         self.assertEqual(tool.canceled, 1)
--- a/enable/tests/test_drag_tool.py
+++ b/enable/tests/test_drag_tool.py
@@ -75,8 +75,7 @@ class DragToolTestCase(EnableTestAssistant, unittest.TestCase):
         tool = self.tool
         tool.end_drag_on_leave = True
         tool._drag_state = "dragging"  # force dragging state
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", DeprecationWarning)
+        with self.assertWarns(DeprecationWarning):
             event = self.mouse_leave(interactor=tool, x=0, y=0)
         self.assertEqual(tool.canceled, 1)
         self.assertEqual(tool._drag_state, "nondrag")
@@ -119,8 +118,7 @@ class DragToolTestCase(EnableTestAssistant, unittest.TestCase):
         tool.end_drag_on_leave = True
         tool.on_drag_leave = 'end'
         tool._drag_state = "dragging"  # force dragging state
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", DeprecationWarning)
+        with self.assertWarns(DeprecationWarning):
             event = self.mouse_leave(interactor=tool, x=0, y=0)
 
         # end_drag_on_leave should be handled like normal
--- a/enable/tests/test_drag_tool.py
+++ b/enable/tests/test_drag_tool.py
@@ -19,8 +19,14 @@ class DummyTool(DragTool):
 
     canceled = Int
 
+    ended = Int
+
     def drag_cancel(self, event):
         self.canceled += 1
+        return True
+
+    def drag_end(self, event):
+        self.ended += 1
         return True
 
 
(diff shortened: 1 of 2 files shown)


jwiggins (commented):
I'm glad you pushed this as a WIP, because I think we perhaps should change nothing and instead add better documentation above the trait and in the docstring for `drag_end` (and maybe `drag_cancel` too).

I'll start by saying `end_drag_on_leave` is a bad name. #naming-things-is-hard, right? So a new trait with clear meaning might be a good long term fix. But would better documentation work too?

At least from my reading the `DragTool` code, `drag_cancel` is the callback for drags which end prematurely and `drag_end` is for drags which end normally. That's not at all clear just from reading the docstrings, so we should fix that. But I think if we start using `drag_end` for drags which ended prematurely that it's just as bad. We would have to introduce yet another callback method `drag_finished` (or something like that; naming things is hard).

corranwebster:
I agree with John here - poorly named trait, documentation could be better (but it is explicitly called out in the `drag_cancel` docstring).  If a code fix is needed, it would be to create a new `cancel_drag_on_leave` trait that is an alias for `end_drag_on_leave` (or vice-versa) and aim to remove `end_drag_on_leave` in Enable 6 or some similar future point.

My intuition is that the "ending" a drag normally when the mouse is still down will lead to an odd UX, but I could be convinced by a concrete example where the behaviour is desirable.

aaronayres35:
> My intuition is that the "ending" a drag normally when the mouse is still down will lead to an odd UX, but I could be convinced by a concrete example where the behaviour is desirable.

A concrete example use case, coming straight from the horse's mouth, would be "trying to enclose something with a bounding box near the border of an image".
You don't want the box to extend beyond the border, (hence you don't want `end_drag_on_leave = False`), but you also don't want to cancel if you drag outside the border.

enable/tools/drag_tool.py
```diff
@@ -24,10 +24,11 @@ class DragTool(BaseTool):
     # The mouse button used for this drag operation.
     drag_button = Enum("left", "right")
 
-    # End the drag operation if the mouse leaves the associated component?
+    # Do nothing, cancel or end the drag operation if the mouse leaves the
+    # associated component?
     # NOTE: This behavior depends on "mouse_leave" events, which in general
     # are not fired when `capture_mouse` is True (default).
```
aaronayres35, line 30: I dont think this note is correct?  If I leave `capture_mouse` as true, but set `end_drag_on_leave` to true and mess around with examples, I am seeing mouse_leave events fired and the drags cancelled on leave?
Maybe I am missing something

jwiggins: It might be different between Qt and Wx.

aaronayres35:
Okay, I've taken a new stab at this given the feedback.

Basically, the current implementation is:
if a user sets `end_drag_on_leave = True` the existing behavior is completely preserved aside from a deprecation warning being raised.  The new `on_drag_leave` trait is effectively ignored (to avoid any potential conflicts).  This way existing code in the wild can go on as is without risk of breaking.  I've added the deprecation warning and added to comments / docstrings to make it more clear that `end_drag_on_leave` actually means `drag_cancel` will be called.  Nonetheless the trait is kept exactly as is, we just warn about it being potentially confusing / misleading.

If a user does not explicitly set `end_drag_on_leave = True`, they are free to use the new feature.  Namely, they can chose to set the new `on_drag_leave` enum trait to either "end" or "cancel" to do what you would expect given the choice.  This provides the appropriate level of flexibility and lets users chose what they want.  Whether you want to cancel or end on a leave or do nothing is context specific, and users should be free to decide on the appropriate behavior for their particular usecase.

finally, if neither trait is set, behavior is preserved as the `on_drag_leave` defaults to `None`.

I've added unit tests to check the `on_drag_leave` options work as expected, and that `on_drag_leave` does nothing if `end_drag_on_leave` is set.

Note that this means if users want the new feature to end on leave not cancel after having perviously had `end_drag_on_leave = True`, they have to set `on_drag_leave = 'end'` **and** remove the `end_drag_on_leave = True`.

enable/tools/drag_tool.py
```diff
@@ -24,11 +27,19 @@ class DragTool(BaseTool):
...
+    # providing this functionality
+    # Cancel the drag operation if the mouse leaves the associated component?
     # NOTE: This behavior depends on "mouse_leave" events, which in general
     # are not fired when `capture_mouse` is True (default).
     end_drag_on_leave = Bool(False)
 
+    # Do nothing, cancel or end the drag operation if the mouse leaves the
+    # associated component?
+    # NOTE: This behavior depends on "mouse_leave" events, which in general
+    # are not fired when `capture_mouse` is True (default).
```
aaronayres35, line 40: maybe I should make it more clear here that `on_drag_leave = 'cancel'` is equivalent to `end_drag_on_leave = True` not `on_drag_leave = 'end'`?  That confusion is sort of the root problem at hand

enable/tools/drag_tool.py
aaronayres35: Completely orthogonal, but this just made me remember that chaco no longer lists wx as being supported in its `ci/edmtool.py` / wx is not exercised on its CI ...

enable/tools/drag_tool.py
```diff
@@ -197,8 +216,24 @@ def _drag_button_up(self, event):
         return False
 
     def _drag_mouse_leave(self, event):
-        if self.end_drag_on_leave and self._drag_state == "dragging":
+        if self.end_drag_on_leave:
+            # raise deprecation warning
+            msg = ("end_drag_on_leave is now deprecated as its name was "
+                  "misleading. It triggers a drag_cancel not drag_end on "
+                  "leave. Use new on_drag_end Enum trait instead.")
+            frame = inspect.currentframe().f_back
```
jwiggins, line 224: You don't appear to be using `frame` anywhere. (and you won't need the `inspect` import if you remove it)

enable/tools/drag_tool.py
```diff
@@ -197,8 +216,24 @@ def _drag_button_up(self, event):
         return False
 
     def _drag_mouse_leave(self, event):
-        if self.end_drag_on_leave and self._drag_state == "dragging":
+        if self.end_drag_on_leave:
+            # raise deprecation warning
+            msg = ("end_drag_on_leave is now deprecated as its name was "
+                  "misleading. It triggers a drag_cancel not drag_end on "
+                  "leave. Use new on_drag_end Enum trait instead.")
```
jwiggins, line 223: Nit: flake8 is complaining about indentation here

enable/tests/test_drag_tool.py
```diff
@@ -82,3 +88,40 @@ def test_mouse_leave_drag_state(self):
...
+        self.assertTrue(event.handled)
+
+    def test_on_drag_leave_no_op(self):
+        """ If end_drag_on_leave is set, on_drag_leave trait is ignored. """
+
+        tool = self.tool
+        tool.end_drag_on_leave = True
+        tool.on_drag_leave = 'end'
+        tool._drag_state = "dragging"  # force dragging state
+        event = self.mouse_leave(interactor=tool, x=0, y=0)
```
jwiggins, line 119: It would be nice to suppress the `DeprecationWarning` here so that the test output is a bit cleaner.

jwiggins (approved):
Could you please add some more details to the `drag_end` callback like you did for `drag_cancel`?

enable/tests/test_drag_tool.py
aaronayres35: That makes sense to me as long as we make sure we don't forget this deprecation exists entirely.  I will suppress the warning and open an issue

jwiggins: Maybe use `assertWarns`?

PR #714, opened 2021-03-12
Use a database object to manage fonts

jwiggins:
This will be useful to #711 (for avoiding duplicate additions).

Instead of using lists inside of `FontManager` to hold collections of `FontEntry` objects, this PR introduces a `FontDatabase` class which ensures all known entries are unique and provides a way to query entries by file path or font family in order to pare down the number of fonts that need to be scored by `FontManager.findfont`.

### DIFF
--- a/kiva/fonttools/_database.py
+++ b/kiva/fonttools/_database.py
@@ -82,6 +82,9 @@ class FontDatabase:
     def fonts_for_family(self, families):
         """ Returns all fonts which best match a particular family query or
         all possible fonts if exact families are not matched.
+
+        `families` is a list of real and generic family names. An iterable
+        of `FontEntry` instances is returned.
         """
         flat_list = (lambda it: list(itertools.chain.from_iterable(it)))
 
@@ -93,6 +96,7 @@ class FontDatabase:
             return entries
 
         # Return all entries if no families found
+        # Yes, self._entries is a set. Consumers should only expect an iterable
         return self._entries
 
     def __len__(self):
--- a/kiva/fonttools/_database.py
+++ b/kiva/fonttools/_database.py
@@ -1,0 +1,115 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+import itertools
+import os
+
+from kiva.fonttools._constants import preferred_fonts
+
+
+class FontEntry(object):
+    """ A class for storing properties of fonts which have been discovered on
+    the local machine.
+
+    `__hash__` is implemented so that set() can be used to prune duplicates.
+    """
+    def __init__(self, fname="", family="", style="normal", variant="normal",
+                 weight="normal", stretch="normal", size="medium",
+                 face_index=0):
+
+        self.fname = fname
+        self.family = family
+        self.style = style
+        self.variant = variant
+        self.weight = weight
+        self.stretch = stretch
+        self.face_index = face_index
+
+        try:
+            self.size = str(float(size))
+        except ValueError:
+            self.size = size
+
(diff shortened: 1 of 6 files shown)


kiva/fonttools/font_manager.py
```diff
@@ -74,17 +74,18 @@ class FontManager:
...
+    __version__ = 10
 
     def __init__(self, size=None, weight="normal"):
         self._version = self.__version__
 
         self.__default_weight = weight
         self.default_size = size if size is not None else 12.0
-
-        paths = []
+        self.default_family = "sans-serif"
```
jwiggins, line 84: This is basically equivalent to the previous value. Bitstream Vera and Helvetica are both sans-serif families. But since not every system may have them, we can use the generic family name for the default and get good results thanks to the `preferred_fonts` dictionary listing several options for sans-serif.

kiva/fonttools/font_manager.py
```diff
@@ -199,24 +198,25 @@ def __repr__(self):
...
 
-        best_score = 1e64
-        best_font = None
+        # Narrow the search
+        if directory is not None:
+            fontlist = font_db.fonts_for_directory(directory)
+        else:
+            fontlist = font_db.fonts_for_family(prop.get_family())
 
+        best_score = 20.0
```
jwiggins, line 217: The highest possible score is only 15, so this is a reasonable starting point.

kiva/fonttools/font_manager.py
```diff
@@ -199,24 +198,25 @@ def __repr__(self):
...
+        best_font = None
         for font in fontlist:
-            fname = font.fname
-            if (directory is not None
-                    and os.path.commonprefix([fname, directory]) != directory):
-                continue
```
jwiggins, line 219: This is now handled by `font_db.fonts_for_directory` above.

kiva/fonttools/tests/test_font_manager.py
```diff
@@ -21,54 +21,18 @@
...
-    @mock.patch("kiva.fonttools._scan_parse.TTCollection",
-                side_effect=RuntimeError)
-    def test_ttc_exception_on_TTCollection(self, m_TTCollection):
-        # When
-        with self.assertLogs("kiva"):
-            fontlist = create_font_list([self.ttc_fontpath])
-
-        # Then
-        self.assertEqual(len(fontlist), 0)
-        self.assertEqual(m_TTCollection.call_count, 1)
```
jwiggins, line 63: These tests are already in `test_scan_sys.py` but were never cleaned up in #707

kiva/fonttools/_database.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Translate generic families into lists of families
+        fams = flat_list(preferred_fonts.get(fam, [fam]) for fam in families)
+        # Then collect all entries for those families
+        entries = flat_list(self._family_map.get(fam, []) for fam in fams)
+        if entries:
+            return entries
+
+        # Return all entries if no families found
+        return self._entries
```
jwiggins, line 96: This should possibly be a `list`? It doesn't break because it's still an iterable, but returning either a `list` or a `set` from a method might be weird. Although this is also only meant to be consumed by `FontManager.findfont`, so it's probably OK.

kiva/fonttools/_database.py
```diff
@@ -0,0 +1,115 @@
...
+        all possible fonts if exact families are not matched.
+        """
+        flat_list = (lambda it: list(itertools.chain.from_iterable(it)))
+
+        # Translate generic families into lists of families
+        fams = flat_list(preferred_fonts.get(fam, [fam]) for fam in families)
+        # Then collect all entries for those families
+        entries = flat_list(self._family_map.get(fam, []) for fam in fams)
+        if entries:
+            return entries
```
jwiggins, line 93: We're taking advantage of how `score_family` works in the context of `FontManager.findfont` here. The score for a candidate font will always be >= 10 if the family doesn't match. Therefore we can return only the subset of fonts covered by `families` here and it won't change the result of `FontManager.findfont`.

I think a comment is in order in any case.

rahulporuri (approved):
LGTM

jwiggins:
Thanks for the review

PR #717, opened 2021-03-15
Add documentation for mouse events

### LINKED ISSUE #625: Enable: document mouse events
We have documentation for keyboard and drag/drop events, but nothing for mouse events.  We can ignore blob events for now.

jwiggins:
Closes #625

### DIFF
--- a/docs/source/enable/mouse_events.rst
+++ b/docs/source/enable/mouse_events.rst
@@ -1,12 +1,13 @@
 Enable Mouse Events
 ===================
 
-Enable mouse events are represented by the :class:`MouseEvent` type and their
-event names (which are the suffixes used by :py:meth:`dispatch`) can be
-divided into two groups: mouse clicks and mouse movements. The mouse click
-events have names ending in ``_down``, ``_up``, or ``_dclick`` and names
-beginning with ``left``, ``right``, or ``middle``. This means that Enable only
-supports three mouse buttons (plus wheel events).
+Enable mouse events are represented by the :class:`~.MouseEvent` type and their
+event names (which are the suffixes used by
+:py:meth:`enable.interactor.Interactor.dispatch`) can be divided into two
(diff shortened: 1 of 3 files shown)


docs/source/enable/mouse_events.rst
```diff
@@ -0,0 +1,52 @@
+Enable Mouse Events
+===================
+
+Enable mouse events are represented by the :class:`MouseEvent` type and their
```
aaronayres35, line 4: This doesn't link to the api docs but we may want it to

docs/source/enable/mouse_events.rst
```diff
@@ -0,0 +1,52 @@
...
+event names (which are the suffixes used by :py:meth:`dispatch`) can be
```
aaronayres35, line 5: same as above

aaronayres35 (approved):
LGTM

I know eventually we plan to re-evaluate doc organization at the end, so no need to change anything here.  But eventually I think this content and the Keyboard events could be on the same page just different sections

docs/source/enable/mouse_events.rst
jwiggins: I'm not sure how to do that.

jwiggins: Figured it out.

docs/source/enable/mouse_events.rst
jwiggins: Done.

docs/source/enable/mouse_events.rst
aaronayres35: I believe you can replace ":class:`MouseEvent`" with ":class:`~enable.events.MouseEvent`"

jwiggins: ":class:`~.MouseEvent`" also works if there's only one `MouseEvent` (which there is).

jwiggins:
Thanks for the feedback

PR #718, opened 2021-03-15
Add documentation for AbstractWindow

### LINKED ISSUE #626: Enable: document window classes
Give overview of `AbstractWindow` and what it provides.

jwiggins:
Closes #626

### DIFF
--- a/docs/source/enable/abstract_window.rst
+++ b/docs/source/enable/abstract_window.rst
@@ -1,0 +1,28 @@
+Top-level Windows
+=================
+When a component is shown on screen via a GUI toolkit, its :attr:`window` trait
+contains an instance of :class:`AbstractWindow` which serves as a delegate
+between the underlying window system and the component.
+
+For the most part, code doesn't need to interact with the underlying window.
+However one common exception is tools which want to set a custom cursor. This
+is accomplished via the :py:meth:`set_pointer` method.
+
+AbstractWindow
+--------------
+The following methods are the public interface of :class:`AbstractWindow`.
+
+.. automethod:: enable.abstract_window.AbstractWindow.get_pointer_position
+  :noindex:
+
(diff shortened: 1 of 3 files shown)


docs/source/enable/abstract_window.rst
```diff
@@ -0,0 +1,28 @@
+Top-level Windows
+=================
+When a component is shown on screen via a GUI toolkit, its :attr:`window` trait
+contains an instance of :class:`AbstractWindow` which serves as a delegate
```
aaronayres35, line 4: This could link to the api docs

That would also make the following section a little redundant, but I don't think it hurts to call it out explicitly like is done here

aaronayres35 (approved):
LGTM

Just one request:
could we change # to #: in the trait defs for the `AbstractWindow` class so they are also picked up by the api docs, not just the methods

jwiggins:
> could we change # to #: in the trait defs for the `AbstractWindow` class so they are also picked up by the api docs

I actually prefer for them not to show up. Can you think of any traits that should be exposed to user code?

docs/source/enable/abstract_window.rst
jwiggins: Done.

aaronayres35:
> > could we change # to #: in the trait defs for the `AbstractWindow` class so they are also picked up by the api docs
> 
> I actually prefer for them not to show up. Can you think of any traits that should be exposed to user code?

I guess that makes sense.  It seems most (all?) the traits will be dealt with by enable inner machinery.

I think I am still a bit unclear on when we anticipate a user needing to mess with the `AbstractWindow`.  Is the custom cursor scenario pretty much the only one that is common?
We are putting this in the user manual but I get the vibe that we don't really want users to touch or have to think about it

jwiggins:
Thanks for the feedback

PR #721, opened 2021-03-15
Document Enable trait types

### LINKED ISSUE #631: Enable: document trait types
Enable defines a number of trait types in various places.  Many of these are intended for external use.  We should document them, if only to mention their existence and briefly state their purpose.

jwiggins:
Closes #631

### DIFF
--- a/docs/source/enable/traits.rst
+++ b/docs/source/enable/traits.rst
@@ -1,0 +1,45 @@
+Enable Custom Traits
+====================
+Enable defines several `trait types <https://docs.enthought.com/traits>`_ which
+can be used to simplify the generation of values to pass to various APIs
+throughout Enable (and Kiva).
+
+
+bounds_trait
+------------
+:class:`~.bounds_trait` represents the bounds of an object. It is a list of two
+values: width and height.
+
+ColorTrait
+----------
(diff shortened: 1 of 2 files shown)


docs/source/enable/traits.rst
```diff
@@ -0,0 +1,45 @@
+Enable Custom Traits
+====================
+Enable defines several `trait types <https://docs.enthought.com/traits>`_ which
```
aaronayres35, line 3: Maybe link to the custom traits section of the traits docs?
https://docs.enthought.com/traits/traits_user_manual/custom.html

I actually am unsure which would make more sense, so feel free to leave if you prefer the home page

aaronayres35 (approved):
LGTM

Most of the links appear to be dead.  Trait definitions are strange though so I think they may not even have api docs correctly generated.  
see https://docs.enthought.com/enable/api/enable.api.html#enable-trait-types

seems like the markers do so they could be linked, but that is probably unnecessary

docs/source/enable/traits.rst
jwiggins: I'll leave it like this

jwiggins:
Thanks for the review

PR #723, opened 2021-03-16
Remove default_size and __default_weight from FontManager

### LINKED ISSUE #709: Do something about FontManager.default_size
See discussion on #707.

The `default_size` attribute comes into play if you call `FontProperties.get_size_in_points` or pass a `FontProperties` with `size` equal to a string from the `font_scalings` dictionary. However, I'm pretty sure no code actually does this.

In both of the above cases, you will fail _unless_ you explicitly set `FontManager.default_size` to a value which can be multiplied with numbers (ie: not None). `_load_from_cache_or_rebuild` is explicitly resetting `default_size` to `None` every time a `FontManager` is loaded from cache.

jwiggins:
Resolves #709

One more disruptive change for `kiva.fonttools`. This removes the `default_size` attribute from `FontManager`, plus all related code. Additionally the unused `__default_weight` attribute is also removed and `FontProperties` has been renamed to `FontQuery` to better reflect how it is actually used (internally, I'll add).


rahulporuri (approved):
LGTM.

IIUC, the user-facing behavior that changes in this PR is the fact that the user can no longer use relative sizes to set font size. Is that correct? But IIUC, the user was never actually able to use relative sizes for fonts because the code has been broken for a while now.

jwiggins:
Yes, that's correct.

PR #724, opened 2021-03-16
Create an independent renderer for draw_marker_at_points

### LINKED ISSUE #708: Implement a backend-agnostic version of draw_marker_at_points
Currently, `kiva.agg` is the only Kiva backend which implements the `draw_marker_at_points` method correctly. This is mostly OK, because 99% of the time that's the only backend an application uses.

However, given #414 and our desire to live in a future without `kiva.agg`, it would be useful have an implementation of `draw_marker_at_points`, written in a fast language, that could draw to any buffer of bytes that was passed to it. Chaco scatter plot renderers would benefit considerably if they can keep their fast drawing path.

`draw_marker_at_points` has built-in limitations in its only existent implementation (no rotation allowed, line widths aren't scaled), so it might be OK to keep those limitations if it means we get a solution sooner.

jwiggins:
Closes #708

This was surprisingly straightforward. Basically, since marker rendering is a much more primitive form of rendering, it's not too bad to just copy the subset of AGG which implements it.

As was the case with #392, there's a bunch of code from AGG here which is just copied in. Well, copied and given a new `agg24markers` namespace so that there's no danger of symbol collisions. (also LF line endings and stripped whitespace) You can basically ignore it for the purpose of reviewing and concentrate on the Python and Cython code.

I've added `draw_marker_at_points` to the celiagg backend to show what an implementation might look like.

### DIFF
--- a/kiva/_marker_renderer.pxd
+++ b/kiva/_marker_renderer.pxd
@@ -25,17 +25,17 @@ cdef extern from "marker_renderer.h" namespace "agg24markers":
 
 
 cdef extern from "marker_renderer.h" namespace "kiva_markers":
-    # This is just here for the type signature/
+    # This is just here for the type signature
     cdef enum marker_type:
         pass
 
     # Abstract base class
     cdef cppclass marker_renderer_base:
-        void draw_marker_at_points(double* pts, unsigned Npts,
-                                   unsigned size, marker_type marker,
-                                   double* fill, double* stroke)
-        void transform(double sx, double shy,
-                       double shx, double sy,
+        bool draw_markers(double* pts, unsigned Npts,
+                          unsigned size, marker_type marker,
+                          double* fill, double* stroke)
+        void transform(double sx, double sy,
+                       double shx, double shy,
                        double tx, double ty)
 
     # Template class
--- a/kiva/_marker_renderer.pxd
+++ b/kiva/_marker_renderer.pxd
@@ -1,0 +1,44 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+from libcpp cimport bool
+
+cdef extern from "marker_renderer.h" namespace "agg24markers":
+    cdef cppclass pixfmt_abgr32:
+        pass
+    cdef cppclass pixfmt_argb32:
+        pass
+    cdef cppclass pixfmt_bgra32:
+        pass
+    cdef cppclass pixfmt_rgba32:
+        pass
+    cdef cppclass pixfmt_bgr24:
+        pass
+    cdef cppclass pixfmt_rgb24:
+        pass
+
+
+cdef extern from "marker_renderer.h" namespace "kiva_markers":
+    # This is just here for the type signature/
+    cdef enum marker_type:
+        pass
+
+    # Abstract base class
+    cdef cppclass marker_renderer_base:
+        void draw_marker_at_points(double* pts, unsigned Npts,
+                                   unsigned size, marker_type marker,
+                                   double* fill, double* stroke)
+        void transform(double sx, double shy,
+                       double shx, double sy,
+                       double tx, double ty)
+
+    # Template class
+    cdef cppclass marker_renderer[pixfmt_T]:
+        marker_renderer(unsigned char* buf, unsigned width, unsigned height,
+                        int stride, bool bottom_up)
--- a/kiva/_marker_renderer.pyx
+++ b/kiva/_marker_renderer.pyx
@@ -14,13 +14,6 @@ from numpy cimport uint8_t
 cimport _marker_renderer
 
 ctypedef _marker_renderer.marker_renderer_base renderer_base_t
-# Template specializations
-ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_abgr32] renderer_abgr32_t
-ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_argb32] renderer_argb32_t
-ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_bgra32] renderer_bgra32_t
-ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_rgba32] renderer_rgba32_t
-ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_bgr24] renderer_bgr24_t
-ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_rgb24] renderer_rgb24_t
 
 @cython.internal
 cdef class MarkerRendererBase:
@@ -38,9 +31,17 @@ cdef class MarkerRendererBase:
         # so that it lives as long as this object
         self.py_array = image
 
-    def draw_marker_at_points(self, points, size, marker, fill, stroke):
-        """draw_marker_at_points(points, size, marker, fill, stroke)
+    def draw_markers(self, points, size, marker, fill, stroke):
+        """draw_markers(points, size, marker, fill, stroke)
         Draw markers at a collection of points.
+
+        :param points: An Nx2 iterable of (x, y) points for marker positions
+        :param size: An integer pixel size for each marker
+        :param marker: A Kiva marker enum integer
+        :param fill: Fill color given as an iterable of 4 numbers (R, G, B, A)
+        :param stroke: Line color given as an iterable of 4 numbers (R, G, B, A)
+
+        :returns: True if any markers were drawn, False otherwise
         """
         cdef:
             double[:,::1] _points = np.asarray(points, dtype=np.float64, order='c')
@@ -52,36 +53,47 @@ cdef class MarkerRendererBase:
         if _points.shape[1] != 2:
             msg = "points argument must be an iterable of (x, y) pairs."
             raise ValueError(msg)
+        if _stroke.shape[0] != 4:
+            msg = "stroke argument must be an iterable of 4 numbers."
+            raise ValueError(msg)
+        if _fill.shape[0] != 4:
+            msg = "fill argument must be an iterable of 4 numbers."
+            raise ValueError(msg)
 
-        self._this.draw_marker_at_points(
-            &_points[0][0],
-            _points.shape[0],
-            _size, _marker,
-            &_fill[0],
-            &_stroke[0]
+        return self._this.draw_markers(
+            &_points[0][0], _points.shape[0], _size, _marker,
+            &_fill[0], &_stroke[0]
         )
 
-    def transform(self, sx, shy, shx, sy, tx, ty):
-        """transform(sx, shy, shx, sy, tx, ty)
-        Set the transform to be applied to the markers.
+    def transform(self, sx, sy, shy, shx, tx, ty):
+        """transform(sx, sy, shy, shx, tx, ty)
+        Set the transform to be applied to the marker points and size.
 
         :param sx: Scale in X
+        :param sy: Scale in Y
         :param shy: Shear in X
         :param shx: Shear in Y
-        :param sy: Scale in Y
         :param tx: Translation in X
         :param ty: Translation in Y
         """
         cdef:
             double _sx = <double>sx
+            double _sy = <double>sy
             double _shy = <double>shy
             double _shx = <double>shx
-            double _sy = <double>sy
             double _tx = <double>tx
             double _ty = <double>ty
 
-        self._this.transform(_sx, _shy, _shx, _sy, _tx, _ty)
+        self._this.transform(_sx, _sy, _shx, _shy, _tx, _ty)
 
+
+# Template specializations
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_abgr32] renderer_abgr32_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_argb32] renderer_argb32_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_bgra32] renderer_bgra32_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_rgba32] renderer_rgba32_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_bgr24] renderer_bgr24_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_rgb24] renderer_rgb24_t
 
 cdef class MarkerRendererABGR32(MarkerRendererBase):
     def __cinit__(self, uint8_t[:,:,::1] image, bottom_up=True):
--- a/kiva/_marker_renderer.pyx
+++ b/kiva/_marker_renderer.pyx
@@ -65,22 +65,22 @@ cdef class MarkerRendererBase:
             &_fill[0], &_stroke[0]
         )
 
-    def transform(self, sx, sy, shy, shx, tx, ty):
-        """transform(sx, sy, shy, shx, tx, ty)
+    def transform(self, sx, sy, shx, shy, tx, ty):
+        """transform(sx, sy, shx, shy, tx, ty)
         Set the transform to be applied to the marker points and size.
 
         :param sx: Scale in X
         :param sy: Scale in Y
-        :param shy: Shear in X
-        :param shx: Shear in Y
+        :param shx: Shear in X
+        :param shy: Shear in Y
         :param tx: Translation in X
         :param ty: Translation in Y
         """
         cdef:
             double _sx = <double>sx
             double _sy = <double>sy
-            double _shy = <double>shy
             double _shx = <double>shx
+            double _shy = <double>shy
             double _tx = <double>tx
             double _ty = <double>ty
 
--- a/kiva/_marker_renderer.pyx
+++ b/kiva/_marker_renderer.pyx
@@ -1,0 +1,126 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+import cython
+import numpy as np
+from numpy cimport uint8_t
+
+cimport _marker_renderer
+
+ctypedef _marker_renderer.marker_renderer_base renderer_base_t
+# Template specializations
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_abgr32] renderer_abgr32_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_argb32] renderer_argb32_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_bgra32] renderer_bgra32_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_rgba32] renderer_rgba32_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_bgr24] renderer_bgr24_t
+ctypedef _marker_renderer.marker_renderer[_marker_renderer.pixfmt_rgb24] renderer_rgb24_t
+
+@cython.internal
+cdef class MarkerRendererBase:
+    cdef renderer_base_t* _this
+    cdef object py_array
+
+    def __dealloc__(self):
+        del self._this
+
+    cdef int base_init(self, image) except -1:
+        if image is None:
+            raise ValueError('image argument must not be None.')
+
+        # Retain a reference to the memory view supplied to the constructor
+        # so that it lives as long as this object
+        self.py_array = image
+
+    def draw_marker_at_points(self, points, size, marker, fill, stroke):
+        """draw_marker_at_points(points, size, marker, fill, stroke)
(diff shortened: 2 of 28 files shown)


kiva/_marker_renderer.pxd
```diff
@@ -0,0 +1,44 @@
...
+    cdef cppclass marker_renderer_base:
+        void draw_marker_at_points(double* pts, unsigned Npts,
+                                   unsigned size, marker_type marker,
+                                   double* fill, double* stroke)
+        void transform(double sx, double shy,
+                       double shx, double sy,
+                       double tx, double ty)
+
+    # Template class
+    cdef cppclass marker_renderer[pixfmt_T]:
```
aaronayres35, line 42: where does `pixfmt_T` come from?  I look in marker_renderer.h and I see `pixfmt_t`

aaronayres35 (approved):
LGTM

I ran the benchmark and the celiagg `draw_marker_at_points` results look correct. 
I did not look through most of the changes line by line as you suggested (skipped bulk of the .h files).  I looked through celiagg implementation, test, etc. and everything made sense. I walked through the cython code and it seemed to all make sense, but given my limited exposure it might be good to get another pair of eyes on those changes specifically

kiva/_marker_renderer.pxd
jwiggins: It's analogous to the C++ `template<typename pixfmt_T>`. It's just a stand in for the type that the class is templated on.

kiva/markers/marker_renderer.h
```diff
@@ -0,0 +1,130 @@
...
+// All rights reserved.
+//
+// This software is provided without warranty under the terms of the BSD
+// license included in LICENSE.txt and may be redistributed only under
+// the conditions described in the aforementioned license. The license
+// is also available online at http://www.enthought.com/licenses/BSD.txt
+//
+// Thanks for using Enthought open source!
+#ifndef KIVA_MARKER_RENDERER_H
+#define KIVA_MARKER_RENDERER_H
```
jwiggins, line 11: This file is new code. Only the code in `kiva/markers/agg` is copied.

kiva/markers/agg/agg_renderer_markers.h
```diff
@@ -0,0 +1,711 @@
...
+                case marker_dash:              do { dash             (int(*x), int(*y), int(r)); ++x; ++y; } while(--n); break;
+                case marker_dot:               do { dot              (int(*x), int(*y), int(r)); ++x; ++y; } while(--n); break;
+                case marker_pixel:             do { pixel            (int(*x), int(*y), int(r)); ++x; ++y; } while(--n); break;
+                case end_of_markers: break;
+            }
+        }
+
+        //--------------------------------------------------------------------
+        template<class T>
+        void markers(int n, const T* x, const T* y, const T* r, marker_e type)
```
jwiggins, line 623: Using this method would give us a `draw_marker_at_points` where `size` could be an array (it's `r` in the signature).

kiva/markers/agg/agg_renderer_markers.h
```diff
@@ -0,0 +1,711 @@
...
+                case marker_dash:              do { dash             (int(*x), int(*y), int(*r)); ++x; ++y; ++r; } while(--n); break;
+                case marker_dot:               do { dot              (int(*x), int(*y), int(*r)); ++x; ++y; ++r; } while(--n); break;
+                case marker_pixel:             do { pixel            (int(*x), int(*y), int(*r)); ++x; ++y; ++r; } while(--n); break;
+                case end_of_markers: break;
+            }
+        }
+
+        //--------------------------------------------------------------------
+        template<class T>
+        void markers(int n, const T* x, const T* y, const T* r, const color_type* fc, marker_e type)
```
jwiggins, line 652: And this version would give per-marker fill color selection (in addition to per-marker sizes).

kiva/markers/agg/agg_renderer_markers.h
```diff
@@ -0,0 +1,711 @@
...
+                case marker_dash:              do { base_type::fill_color(*fc); dash             (int(*x), int(*y), int(*r)); ++x; ++y; ++r; ++fc; } while(--n); break;
+                case marker_dot:               do { base_type::fill_color(*fc); dot              (int(*x), int(*y), int(*r)); ++x; ++y; ++r; ++fc; } while(--n); break;
+                case marker_pixel:             do { base_type::fill_color(*fc); pixel            (int(*x), int(*y), int(*r)); ++x; ++y; ++r; ++fc; } while(--n); break;
+                case end_of_markers: break;
+            }
+        }
+
+        //--------------------------------------------------------------------
+        template<class T>
+        void markers(int n, const T* x, const T* y, const T* r, const color_type* fc, const color_type* lc, marker_e type)
```
jwiggins, line 681: Variable fill and stroke colors.

jwiggins (commented):
There are several additional `markers` methods in AGG that we could use to speed up various code paths in chaco's scatterplot (size and color variations). With the marker rendering existing independently of the kiva backend, we can even entertain such an idea...

kiva/_marker_renderer.pyx
```diff
@@ -0,0 +1,126 @@
...
+        cdef:
+            double[:,::1] _points = np.asarray(points, dtype=np.float64, order='c')
+            double[::1] _fill = np.asarray(fill, dtype=np.float64, order='c')
+            double[::1] _stroke = np.asarray(stroke, dtype=np.float64, order='c')
+            unsigned _size = <unsigned>size
+            _marker_renderer.marker_type _marker = <_marker_renderer.marker_type>marker
+
+        if _points.shape[1] != 2:
+            msg = "points argument must be an iterable of (x, y) pairs."
+            raise ValueError(msg)
```
jwiggins, line 54: We should probably also check that `_fill` and `_stroke` are the right shape. The C++ code which consumes them only gets a `double *` and has no way to know if it's reading garbage.

kiva/_marker_renderer.pyx
```diff
@@ -0,0 +1,126 @@
...
+            raise ValueError('image argument must not be None.')
+
+        # Retain a reference to the memory view supplied to the constructor
+        # so that it lives as long as this object
+        self.py_array = image
+
+    def draw_marker_at_points(self, points, size, marker, fill, stroke):
+        """draw_marker_at_points(points, size, marker, fill, stroke)
+        Draw markers at a collection of points.
+        """
```
jwiggins, line 44: Should probably fill out the docstring here.

kiva/_marker_renderer.pyx
```diff
@@ -0,0 +1,126 @@
...
+        self._this = <renderer_base_t*> new renderer_bgr24_t(
+            &image[0][0][0], image.shape[1], image.shape[0], image.strides[0], bottom_up
+        )
+
+cdef class MarkerRendererRGB24(MarkerRendererBase):
+    def __cinit__(self, uint8_t[:,:,::1] image, bottom_up=True):
+        self.base_init(image)
+        self._this = <renderer_base_t*> new renderer_rgb24_t(
+            &image[0][0][0], image.shape[1], image.shape[0], image.strides[0], bottom_up
+        )
```
jwiggins, line 126: You think there's a nicer way to do these specializations?

kiva/celiagg.py
```diff
@@ -826,6 +835,33 @@ def draw_path_at_points(self, points, path, mode=constants.FILL_STROKE):
...
+
+        # Grab the fill and stroke colors (where possible)
+        fill = (0.0, 0.0, 0.0, 0.0)
+        stroke = (0.0, 0.0, 0.0, 1.0)
+        if isinstance(self.fill_paint, agg.SolidPaint):
+            fp = self.fill_paint
+            fill = (fp.r, fp.g, fp.b, fp.a)
+        if isinstance(self.stroke_paint, agg.SolidPaint):
+            sp = self.stroke_paint
+            stroke = (sp.r, sp.g, sp.b, sp.a)
```
jwiggins, line 858: Maybe it would be better to add `set_stroke_color` and `set_fill_color` methods to `MarkerRenderer*` so that we can just pass through colors from `GraphicsContext.set_*_color` methods? This particular block of code is just here in case a gradient fill color was set.

corranwebster:
Absolute +1 for the ability to provide arrays of sizes, markers shapes and colors. That would permit _huge_ simplification of scatterplot/cmap_scatterplot code and potentially speed things up as well, since currently differences in color/marker/size mean rendering calls in Python loops.

kiva/tests/test_marker_rendering.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+
+from kiva import constants
+from kiva.marker_renderer import MarkerRendererRGB24
+
+
+class TestMarkerDrawing(unittest.TestCase):
+    def setUp(self):
+        self.buffer = np.ones((300, 300, 3), dtype=np.uint8)
+        self.gc = MarkerRendererRGB24(self.buffer)
```
jwiggins, line 22: Might be a good idea to test all variations of the renderer here?

jwiggins:
Something I left out here: I didn't commit the 30k lines of Cython-generated C++ code like what was done for `kiva._cython_speedups`. We don't do that for `kiva.quartz` either, but I know there's been discussion about that in the past.

kiva/celiagg.py
```diff
@@ -89,8 +95,11 @@ def __init__(self, size, *args, **kwargs):
         self.pix_format = kwargs.get('pix_format', 'rgba32')
 
         shape = (self._height, self._width, 4)
+        buffer = np.zeros(shape, dtype=np.uint8)
         canvas_klass = pix_format_canvases[self.pix_format]
-        self.gc = canvas_klass(np.zeros(shape, dtype=np.uint8), bottom_up=True)
+        self.gc = canvas_klass(buffer, bottom_up=True)
+        marker_ren_klass = pix_format_marker_renderers[self.pix_format]
+        self.marker_gc = marker_ren_klass(buffer, bottom_up=True)
```
corranwebster, line 102: Are there potential issues with keeping this synchronized with the main gc?  In particular, is there a risk of this holding an old buffer because something (eg. screen resize, code that is rummaging around in the internals of the GC when it shouldn't be, etc.) changed the original buffer.

I don't know about the performance implications, but if creating the marker gc is fast and lightweight enough it might make sense to create and dispose of as needed in `draw_marker_at_points`.

Or it may be that there isn't any way that the buffer can get changed, so this doesn't matter.

corranwebster:
With the code copied from Agg, was this from the version in Kiva, or from the updated non-GPL Agg repo?

kiva/celiagg.py
jwiggins: The GC size never changes. If the parent widget is resized, a new GC is instantiated.

jwiggins:
> was this from the version in Kiva, or from the updated non-GPL Agg repo?

From Kiva, but we updated that version from the upstream SVN repo in #288

corranwebster:
QPainter prefers premultiplied ARGB32 for its internal buffer format, I think - is that going to make using this for the QPainter backend difficult?  Don't have the insight into Agg's internal buffer formats to know if there is a mismatch here.

jwiggins:
>  is that going to make using this for the QPainter backend difficult?

AGG has pixel formats for everything, basically:
https://github.com/enthought/enable/pull/724/files#diff-ffe2dc1d178b04a7e935d0f03d7562024ef0782721d67026dea38d28dc061efdR2750-R2753

We can make more versions of MarkerRenderer as needed. I believe Quartz uses premultiplied alpha as well.

rahulporuri (commented):
Can you add `kiva/_marker_renderer.cpp` to the git ignore list given that we don't want to include it in the git repo?

setup.py
```diff
@@ -292,6 +292,23 @@ def base_extensions():
...
+                os.path.join('kiva', 'markers', 'agg'),
+                os.path.join('kiva', 'markers'),
+                numpy.get_include(),
+            ],
+            define_macros=[
+                ("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"),
+            ],
+            language='c++',
+        ),
+
```
rahulporuri, line 311: nit : not sure if the new line here was intentional
```suggestion
```

rahulporuri (approved):
LGTM with a couple of comments/questions (some nitpicky)

kiva/celiagg.py
rahulporuri: i'm not sure i fully understand what you're recommending here.

kiva/tests/test_marker_rendering.py
rahulporuri: that's definitely a good idea!

kiva/_marker_renderer.pyx
rahulporuri: _has no understanding whatsoever about templates/specializations so cannot provide constructive comments_ Should we ask internally if anyone knows better?

rahulporuri:
Also, I don't think we want to implement `draw_marker_at_points` for any of the other backends - but it'll probably be useful for aaron (or someone new) to implement the method for new backend to get comfortable with the code. We definitely need an issue to track this work though.

kiva/_marker_renderer.pyx
```diff
@@ -0,0 +1,126 @@
...
+
+        self._this.draw_marker_at_points(
+            &_points[0][0],
+            _points.shape[0],
+            _size, _marker,
+            &_fill[0],
+            &_stroke[0]
+        )
+
+    def transform(self, sx, shy, shx, sy, tx, ty):
```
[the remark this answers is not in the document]
jwiggins: I can change the order. I guess I'd imagined passing it through as an array at first.

jwiggins:
> Also, I don't think we want to implement draw_marker_at_points for any of the other backends

You mean in this PR, right? No complaints from my side there.

Adding this to an AGG backend is fairly trivial because we're already in control of the backing store there. It might be a bit more challenging for the Quartz and QPainter backends.

kiva/_marker_renderer.pyx
jwiggins: There are options: https://martinralbrecht.wordpress.com/2017/07/23/adventures-in-cython-templating/

rahulporuri:
I'm testing this branch (and enable main branch) against an internal application - and there's one problem i'm running into when I use the celiagg backend.

The internal application uses the `MarkerTrait` - https://github.com/enthought/enable/blob/2ce5387fb2ab2e743065fde096a7dd69e024b6b4/enable/markers.py#L486-L489 - and when I try to use the following markers, i don't see anything in the plot - `left_triangle`, `right_triangle`, `pentagon`, `hexagon`, `hexagon2`, `star`, `cross_plus`.

rahulporuri:
Also, there are a *ton* of log messages now from `fontTools`, most of them of the form - 

```
2021-03-18 14:38:11,681 DEBUG    [fontTools.ttLib.ttFont:381] Reading 'name' table from disk
2021-03-18 14:38:11,681 DEBUG    [fontTools.ttLib.ttFont:390] Decompiling 'name' table
```

corranwebster:
> when I try to use the following markers, i don't see anything in the plot - `left_triangle`, `right_triangle`, `pentagon`, `hexagon`, `hexagon2`, `star`, `cross_plus`.

These markers don't have native agg versions, and so are rendered by paths only; so `draw_marker_at_points` doesn't work with them (in their definition you will see that they have a `kiva_marker` value of `NO_MARKER`

rahulporuri:
> These markers don't have native agg versions, and so are rendered by paths only; so `draw_marker_at_points` doesn't work with them (in their definition you will see that they have a `kiva_marker` value of `NO_MARKER`)

But this worked fine with the agg backend, which is why im slightly concerned

corranwebster:
Oh, OK - maybe the Agg version has a codepath for markers defined by paths?

Additionally, _some_  of these correspond to agg markers, so they could be supported:

Agg markers are (checked if it is exposed by Kiva):

- [x] marker_square
- [x] marker_diamond
- [x] marker_circle
- [x] marker_crossed_circle
- [ ] marker_semiellipse_left
- [ ] marker_semiellipse_right
- [ ] marker_semiellipse_up
- [ ] marker_semiellipse_down
- [ ] marker_triangle_left
- [ ] marker_triangle_right
- [x] marker_triangle_up
- [x] marker_triangle_down
- [ ] marker_four_rays
- [x] marker_cross
- [x] marker_x
- [ ] marker_dash
- [x] marker_dot
- [x] marker_pixel
```

File: kiva/celiagg.py [hunk 14]
```diff
@@ -826,6 +835,33 @@ def draw_path_at_points(self, points, path, mode=constants.FILL_STROKE):
             fill=self.fill_paint,
         )
 
+    def draw_marker_at_points(self, points, size,
```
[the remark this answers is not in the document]
jwiggins: I was just trying to keep it on a single line.

kiva/celiagg.py
jwiggins: Add `set_*_color` methods to `MarkerRenderer`, then in the `set_*_color` methods of `GraphicsContext` call `self.marker_gc.set_*_color(color)` so that the marker renderer stays in sync with the simple fill and stroke colors. If you set a gradient on the celiagg backend and then call `draw_marker_at_points`, the fill color of the markers will be transparent. One could argue that it's a bug in the code doing the drawing, but it's a silent failure.

corranwebster:
For the record here, issue was that Chaco code expects `draw_marker_at_points` to return `False` or `0` when it can't draw because it was passed `NO_MARKER`.

kiva/celiagg.py
jwiggins: I won't doing anything about it in this PR

kiva/_marker_renderer.pyx
```diff
@@ -0,0 +1,126 @@
...
+    def draw_marker_at_points(self, points, size, marker, fill, stroke):
+        """draw_marker_at_points(points, size, marker, fill, stroke)
+        Draw markers at a collection of points.
+        """
+        cdef:
+            double[:,::1] _points = np.asarray(points, dtype=np.float64, order='c')
+            double[::1] _fill = np.asarray(fill, dtype=np.float64, order='c')
+            double[::1] _stroke = np.asarray(stroke, dtype=np.float64, order='c')
+            unsigned _size = <unsigned>size
+            _marker_renderer.marker_type _marker = <_marker_renderer.marker_type>marker
```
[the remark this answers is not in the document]
jwiggins: Cython does this for us. I'm also exercising it in the unit tests.

jwiggins:
Thanks for all the great feedback and testing. This should be good to go once CI is happy and y'all have taken a look at my new changes.

kiva/api.py
```diff
@@ -148,3 +160,7 @@
 )
 from ._cython_speedups import points_in_polygon
 from .fonttools import add_application_fonts, Font
+from .marker_renderer import (
+    MarkerRendererABGR32, MarkerRendererARGB32, MarkerRendererBGR24,
+    MarkerRendererBGRA32, MarkerRendererRGB24, MarkerRendererRGBA32
+)
```
jwiggins, line 166: Oops. This is wrong now.

kiva/_marker_renderer.pyx
```diff
@@ -0,0 +1,138 @@
...
+            msg = "fill argument must be an iterable of 4 numbers."
+            raise ValueError(msg)
+
+        return self._this.draw_markers(
+            &_points[0][0], _points.shape[0], _size, _marker,
+            &_fill[0], &_stroke[0]
+        )
+
+    def transform(self, sx, sy, shy, shx, tx, ty):
+        """transform(sx, sy, shy, shx, tx, ty)
```
rahulporuri, line 69: IIUC this is a mistake - and the order does matter.
```suggestion
    def transform(self, sx, sy, shx, shy, tx, ty):
        """transform(sx, sy, shx, shy, tx, ty)
```

kiva/_marker_renderer.pyx
```diff
@@ -0,0 +1,138 @@
...
+        Set the transform to be applied to the marker points and size.
+
+        :param sx: Scale in X
+        :param sy: Scale in Y
+        :param shy: Shear in X
+        :param shx: Shear in Y
```
rahulporuri, line 75:
```suggestion
        :param shx: Shear in X
        :param shy: Shear in Y
```

kiva/_marker_renderer.pyx
```diff
@@ -0,0 +1,138 @@
...
+        if _fill.shape[0] != 4:
+            msg = "fill argument must be an iterable of 4 numbers."
+            raise ValueError(msg)
+
+        return self._this.draw_markers(
+            &_points[0][0], _points.shape[0], _size, _marker,
+            &_fill[0], &_stroke[0]
+        )
+
+    def transform(self, sx, sy, shy, shx, tx, ty):
```
rahulporuri, line 68: Do we need to exercise `transform` in the testsuite? I guess not because it's a thin wrapper around the underlying agg code.

kiva/markers/marker_renderer.h
```diff
@@ -0,0 +1,130 @@
...
+                m_transform.transform(&mx, &my);
+                m_renderer.marker(int(mx), int(my), size * scale, marker);
+            }
+
+            return true;
+        }
+
+        void transform(const double sx, const double sy,
+                       const double shy, const double shx,
+                       const double tx, const double ty)
```
rahulporuri, line 92: i'm sorry but `transform` is freaking me out now - the signature for `transform` on `marker_renderer_base` does not match this - again, the `shy` and `shx` are switched around. I apologize if im missing something obvious.

kiva/_marker_renderer.pyx
jwiggins: Yeah. I can add that.

kiva/markers/marker_renderer.h
jwiggins: I'm glad you're looking close enough to spot it. Hopefully my latest commit has it right now. 😅

corranwebster:
Not going to have enough time to do this myself today, but if we haven't we should make sure that the various Chaco scatterplot types work with this as expected.  They are the primary users.

For fun it might be also worth comparing how fast rendering large numbers of "pixel" markers is on agg and celiagg.  Based on really old numbers I would expect ~100k to 1m points to be reasonably performant.

rahulporuri (approved):
:shipit:

PR #729, opened 2021-03-18
Make celiagg a little bit more like kiva.agg

jwiggins:
This makes some small changes to the celiagg backend so that a celiagg `GraphicsContext` looks a bit more like a `kiva.agg` `GraphicsContext`. This covers [this comment](https://github.com/enthought/enable/issues/414#issuecomment-737519739)  on #414 by adding a `bmp_array` attribute to celiagg's `GraphicsContext`. Additionally, "bgra32" is now the default pixel format.


aaronayres35 (approved):
LGTM

PR #738, opened 2021-03-22
Clarify __init__s which allow `component` as a pos. arg.

### LINKED ISSUE #736: Redundant code in BaseTool.__init__
The first two lines of `BaseTool.__init__` are redundant and will never be run.  The can safely be removed.

https://github.com/enthought/enable/blob/84f75d36d012093e5aee78f80a209dbb1665b0eb/enable/base_tool.py#L126-L130

### LINKED ISSUE #737: AbstractOverlay.__init__ is redundant
https://github.com/enthought/enable/blob/84f75d36d012093e5aee78f80a209dbb1665b0eb/enable/abstract_overlay.py#L61-L64

There seems to be no need for special-handling of the `component` trait - it can just be added to the keywords and passed through.

### LINKED ISSUE #739: Audit __init__ methods
As seen in #737 and #736 there are some classes which are doing things like setting traits in the init method (particularly when before the `super()` call) or not passing values through.  There can be legitimate reasons for doing this (eg. to guarantee that certain traits hold valid values before their trait change notifiers fire), but in some cases these are interfering with the expected behaviour of the classes.

Additional classes that should be investigated include `ViewportToolbar` and `ViewportZoomTool `.  There are likely to be others.

jwiggins:
Fixes #736
Fixes #737
Fixes #739

@corranwebster I see that `ViewportToolbar` and `ViewportZoomTool` also have `component=None` in their `__init__` method signatures.

### DIFF
--- a/enable/base_tool.py
+++ b/enable/base_tool.py
@@ -170,5 +170,5 @@ class BaseTool(Interactor):
     def deactivate(self, component=None):
         """ Handles this component no longer being the active tool.
         """
-        # Compatibility with new AbstractController interface
+        # Compatibility with [Chaco's] AbstractController interface
         self._deactivate()
--- a/enable/base_tool.py
+++ b/enable/base_tool.py
@@ -123,11 +123,11 @@ class BaseTool(Interactor):
     # Concrete methods
     # ------------------------------------------------------------------------
 
-    def __init__(self, component=None, **kwtraits):
-        if "component" in kwtraits:
-            component = kwtraits["component"]
-        super(BaseTool, self).__init__(**kwtraits)
-        self.component = component
+    # Implement __init__ so `component` trait can be passed as a positional arg
+    def __init__(self, component=None, **traits):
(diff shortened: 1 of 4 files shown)


enable/base_tool.py
```diff
@@ -170,5 +170,5 @@ def _deactivate(self):
     def deactivate(self, component=None):
         """ Handles this component no longer being the active tool.
         """
-        # Compatibility with new AbstractController interface
+        # Compatibility with [Chaco's] AbstractController interface
```
corranwebster, line 173: Square brackets around `[Chaco's]`?  A search and replace gone wrong?
```suggestion
        # Compatibility with Chaco's AbstractController interface
```

Occurs a couple of other places too.

jwiggins: No, that was intentional.

rahulporuri (approved):
LGTM

jwiggins:
Thanks for the feedback

PR #740, opened 2021-03-22
Modify PR workflow for better cache utilization

### LINKED ISSUE #601: GitHub Actions CI keeps hitting and missing the edm package cache
I'm not sure why the GitHub Actions CI isn't able to use the cache on any platform. See for example jobs on the PR https://github.com/enthought/enable/pull/599

- Linux + PyQt5 https://github.com/enthought/enable/runs/1937004829 
- Mac OS + PyQt5- https://github.com/enthought/enable/pull/599/checks?check_run_id=1937005006
- Windows + PyQt5 https://github.com/enthought/enable/pull/599/checks?check_run_id=1937005141

On all three platforms, messages of the form below can be seen in the Cache EDM Packages step of the jobs.

```
Cache not found for input keys: macOS-pyqt5-ad8fc51d531ddbde3c34c7c037cc0b3105c6d1a1dd735c4cf83f06ea624cf4ed
```

jwiggins:
Resolves #601

This is actually pretty ugly... See [this discussion](https://github.community/t/how-to-perform-an-early-clean-exit-of-a-job/17531) to understand why we have an `if:` added to everything after the `actions/cache` runs.

Basically, we need to run CI jobs on the master branch each time the hash key changes. Otherwise PRs will always build a cache on their first run of the workflow, because no cache was found for the feature branch or the master branch. The Actions documentation explains it [here](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

### DIFF
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -53,6 +53,8 @@ jobs:
         run: edm --config ci/.edm.yaml install -y wheel click coverage
       - name: Install test environment
         if: github.event_name == 'pull_request' || steps.cacheStep.outputs.cache-hit != 'true'
+        env:
+          PIP_CACHE_DIR: ~/.cache/pip
         run: edm run -- python ci/edmtool.py install --toolkit=${{ matrix.toolkit }}
       - name: Run tests
         if: github.event_name == 'pull_request' || steps.cacheStep.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/test-with-edm.yml
```diff
@@ -32,19 +37,26 @@ jobs:
...
         run: sudo apt-get install libglu1-mesa-dev
       - name: Cache EDM packages
+        id: cacheStep
         uses: actions/cache@v2
+        env:
+          cache-name: ${{ runner.os }}-${{ matrix.toolkit }}-${{ hashFiles('ci/edmtool.py') }}
         with:
           path: ~/.cache
-          key: ${{ runner.os }}-${{ matrix.toolkit }}-${{ hashFiles('ci/edmtool.py') }}
+          key: ${{ env.cache-name }}
```
jwiggins, line 46: This change isn't really necessary.

jwiggins:
We might need to give up here... If you look at the most recent CI run for this branch, _the cache was not hit_. Possibly due to eviction? That's detailed [here](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).

Maybe the only thing worth keeping here is the `PIP_CACHE_DIR` addition?

Thaddeuswider555 (approved):

PR #726, opened 2021-03-17
Add support for text styles on Win32+Celiagg

jwiggins:
Fixes #643

Celiagg 2.1.0 introduces a new font object for win32 which allows specification of bold/italic styles.

Issue #643, opened 2021-02-26
Celiagg backend font selection is wrong on Windows


jwiggins:
Due to Celiagg using Windows font rendering, the selection of fonts needs to work a bit differently on Windows.

With FreeType, a `celiagg.Font` gets passed the path to the font file being loaded (along with a face index if it's a font collection). But with Windows, the first argument of `celiagg.Font` is a face name that gets passed to [`CreateFont`](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-createfonta).

To work around this, all we need to do is use `kiva.Font.findfontname()` on Windows and continue using `kiva.Font.findfont()` on other platforms.

jwiggins:
The foundation for this has been laid by #605, celiagg/celiagg#84 and celiagg/celiagg#85. Once a new version of celiagg has been released, some small changes to the `set_font` method will be needed to apply the font style.

PR #741, opened 2021-03-22
Pick #726 into the 5.1.0 release branch

jwiggins:
Now that #643 is closed, it should be included in 5.1.0. I've also unset the `IS_RELEASED` flag until we're ready to cut rc2.


aaronayres35 (approved):
LGTM

PR #744, opened 2021-03-23
[WIP] Run testsuite against latest pillow (7.2.0)

rahulporuri:


rahulporuri:
Closing the PR as starting with version 7.0.0, pillow stopped supporting PyQt4 - https://pillow.readthedocs.io/en/7.2.0/releasenotes/7.0.0.html#pyqt4-and-pyside. This is a problem because we (ETS in general) still wants to support PyQt4.

PR #745, opened 2021-03-23
Don't test drawing on Qt4

jwiggins:
Our Pillow requirement has been upgraded to version 7.2.0 in EDM. This version removed support for Qt4. To get CI working again, we should skip tests of QPainter drawing on Qt4.

### DIFF
--- a/kiva/tests/test_qpainter_drawing.py
+++ b/kiva/tests/test_qpainter_drawing.py
@@ -11,13 +11,7 @@ import sys
 import unittest
 
 try:
-    from pyface.qt import QtGui
-except ImportError:
-    QT_NOT_AVAILABLE = True
-else:
-    QT_NOT_AVAILABLE = False
-try:
-    from pyface.qt import is_qt5
+    from pyface.qt import is_qt5, QtGui
 except ImportError:
     is_qt5 = False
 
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
Code changes LGTM but I don't know if there are any documentation changes that are necessary. I'm not entirely sure what all backends will continue to support Qt4 and Qt5 once we start depending on pillow >= 7.2.0. I _think_ the agg and celiagg backends will continue supporting both Qt4 and Qt5. Is that correct?

jwiggins:
Yes, the AGG backends will continue to support Qt4 and Qt5. We can also just "vendor" the `QImage` generation code so that we don't need to get it from Pillow. All backends which paint an offscreen GC into a QWidget would be able to use such code.

rahulporuri:
>  We can also just "vendor" the QImage generation code so that we don't need to get it from Pillow.

I don't think we need that at the moment. We can announce with the 5.1.0 release that Qt4 is only supported by the agg backends if the edm eggs are used (because itll depend on pillow >= 7.2.0) and if someone complains, we can consider vendoring that code. Otherwise, i think it's an unnecessary maintenance overhead.

jwiggins:
Thanks for the review

PR #746, opened 2021-03-23
Update constrained layout documentation

### LINKED ISSUE #706: Enable: Fill holes in constraints layout documentation
The constraints layout documentation is pretty good, but not so good that everyone gets it right the first time they try to use it. Some holes:

* `hbox` points out that it cares about the height of components but `horizontal` does not mention that it ignores height (and might zero out the height of your components absent other constraints).
* Ditto for `vbox` and `vertical` (but replace "height" with "width")
* `auto_size` should probably be ignored when using `ConstraintsContainer`.
* Elaborate on the relationship between `bounds`, `position`, and the constraints variables (`CoordinateBox._constraints_vars`)
* Note that components might end up with a size of 0 if they are incompletely constrained.

jwiggins:
Resolves #706

### DIFF
--- a/docs/source/enable/constraints_layout.rst
+++ b/docs/source/enable/constraints_layout.rst
@@ -12,7 +12,7 @@ Using Constraints
 -----------------
 
 :class:`ConstraintsContainer` is a :class:`Container` subclass which uses the
-Cassowary_ constraint solver to determine the layout of its child
+Kiwisolver_ constraint solver to determine the layout of its child
 :class:`Component` instances. This is achieved by adding constraint variables
 to the :class:`Component` class which define a simple box model:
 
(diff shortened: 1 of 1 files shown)


docs/source/enable/constraints_layout.rst
```diff
@@ -90,43 +90,54 @@ also available in Enable. The layout helpers are:
 
 :data:`spacer`: Creates space between two adjacent components.
 
-.. function:: horizontal(*components[, spacing=10])
+.. function:: hbox(*components[, spacing=10, margins=...])
```
notmatthancock, line 93: Why `...` instead of 0?

notmatthancock (commented):
Looks good. Thanks for the additional explanations. Just one question.

docs/source/enable/constraints_layout.rst
```diff
@@ -12,7 +12,7 @@ Using Constraints
 -----------------
 
 :class:`ConstraintsContainer` is a :class:`Container` subclass which uses the
```
aaronayres35, line 14: Nit: These don't link to api docs

aaronayres35 (approved):
Changes LGTM

I notice that _none_ of the links to classes, traits, etc. in this file actually successfully link to api docs which would be a nice improvement. 

Also, doesn't need to go in to this PR by any means, but a full fledged example and/or some screenshots showing different layouts and code to achieve them might make the documentation much easier/quicker to understand

docs/source/enable/constraints_layout.rst
jwiggins: Because `margins` has several possible values which are enumerated below this line

docs/source/enable/constraints_layout.rst
jwiggins: Fixed

docs/source/enable/constraints_layout.rst
notmatthancock: Yes, I thought this line was meant to indicate the default value that the parameter takes on, not to indicate the range of types that it accepts, which is why I thought it might read 0 instead.

It doesn't matter too much; I was just curious since I hadn't seen ellipsis used like this.

jwiggins:
Thanks for the reviews

PR #748, opened 2021-03-23
Deprecate the str_to_font in enable.base

### LINKED ISSUE #1: Revisit or deprecate Enable's string-to-font stuff
In enable/base.py, there is some string-to-font logic/code starting around line 180 that should probably be removed at this point. Is the functionality there already covered by Kiva?

https://svn.enthought.com/enthought/ticket/355

jwiggins:
Resolves #1 (it only took 10 years 😎)

### DIFF
--- a/enable/api.py
+++ b/enable/api.py
@@ -16,7 +16,6 @@ Enable Base
 ===========
 
 - :class:`~.IDroppedOnHandler`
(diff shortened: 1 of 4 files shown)


rahulporuri (approved):
LGTM

jwiggins:
Thanks for the review

PR #751, opened 2021-03-24
Catch the exception thrown by AGG

### LINKED ISSUE #750: Handle std::overflow_error exception from AGG
#288 brought in AGG changes from Matplotlib. It also introduced the one and only `throw` from AGG code. Unfortunately that except is not being caught, so it will result in `abort()` being called if it is ever triggered.

Related issues:
* enthought/chaco#474
* celiagg/celiagg#96 (fixed in celiagg/celiagg#97)

jwiggins:
Resolves #750

This is basically the same change as celiagg/celiagg#97, but with SWIG exception handling instead of Cython. A test has also been added to show that the error is now recoverable.

(Celiagg keeps its rasterizer instance around between draw calls, so it needed an additional change to the AGG code which is not needed here)

### DIFF
--- a/kiva/agg/src/graphics_context.i
+++ b/kiva/agg/src/graphics_context.i
@@ -722,10 +722,26 @@ namespace kiva {
 
             void clear_clip_path();
             void clear(agg24::rgba& value=_clear_color);
+
+            // Exception handler for cell block overflow
+            %exception {
+                try
+                {
+                    $action
+                }
+                catch (const std::overflow_error& exn)
+                {
+                    PyErr_SetString(PyExc_OverflowError, exn.what());
+                    return NULL;
+                }
+            }
+
             void stroke_path();
             void fill_path();
             void eof_fill_path();
             void draw_path(draw_mode_e mode=FILL_STROKE);
+
+            %exception;  // clear exception handlers
 
             void draw_rect(double rect[4],
                            draw_mode_e mode=FILL_STROKE);
--- a/kiva/agg/src/graphics_context.i
+++ b/kiva/agg/src/graphics_context.i
@@ -732,6 +732,13 @@ namespace kiva {
                 catch (const std::overflow_error& exn)
                 {
                     PyErr_SetString(PyExc_OverflowError, exn.what());
+
+                    // XXX: This is a hacky, but `arg1` is the graphics context
+                    // object in the SWIG-generated code.
+                    // The path needs to be cleared by drawing calls and this
+                    // guarantees that it is even if an exception is thrown.
+                    (arg1)->begin_path();
+
                     return NULL;
                 }
             }
--- a/kiva/tests/drawing_tester.py
+++ b/kiva/tests/drawing_tester.py
@@ -184,8 +184,8 @@ class DrawingImageTester(DrawingTester):
 
         filename = "{0}.png".format(self.filename)
         self.gc.save(filename)
-        image = Image.open(filename)
-        dpi = image.info['dpi']
+        with Image.open(filename) as image:
+            dpi = image.info['dpi']
         return dpi[0]
 
     @contextlib.contextmanager
--- a/kiva/tests/test_agg_drawing.py
+++ b/kiva/tests/test_agg_drawing.py
@@ -46,3 +46,27 @@ class TestAggDrawing(DrawingImageTester, unittest.TestCase):
         with open(filename, 'wb') as fp:
             fp.write(stream)
         self.assertImageSavedWithContent(filename)
+
+    def test_rasterizer_cell_overflow(self):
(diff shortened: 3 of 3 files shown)


kiva/agg/src/graphics_context.i
```diff
@@ -722,11 +722,27 @@ namespace kiva {
...
+                {
+                    PyErr_SetString(PyExc_OverflowError, exn.what());
+                    return NULL;
+                }
+            }
+
             void stroke_path();
             void fill_path();
             void eof_fill_path();
             void draw_path(draw_mode_e mode=FILL_STROKE);
```
rahulporuri, line 742: Correct me if i'm wrong but exceptions in these 4 calls are handled by the surrounding `%exception` definitions?

rahulporuri (approved):
LGTM

kiva/agg/src/graphics_context.i
jwiggins: That's my understanding

kiva/tests/test_agg_drawing.py
```diff
@@ -46,3 +46,27 @@ def test_ipython_repr_png(self):
...
+            return arr
+
+        # This many points definitely generates more than 2^22 cells
+        count = 10000
+        points = genpoints(count)
+        self.gc.lines(points)
+        with self.assertRaises(OverflowError):
+            self.gc.stroke_path()
+
+        self.gc.begin_path()
```
jwiggins, line 66: Not sure if we want to do this, but without this `begin_path` the next draw will trigger the exception again because throwing an exception caused the path to not be cleared by the draw.

rahulporuri:
I tested this branch against the code in https://github.com/enthought/chaco/issues/474 and it works as expected. On the main branch, python just halted quietly (with some additional information when we use the `-Xfaulthandler` flag). On this branch, I get a much cleaner error -

```
  File "c:\users\rporuri\work\github\ets\enable\enable\qt4\base_window.py", line 243, in paintEvent
    self.handler.paintEvent(event)
  File "c:\users\rporuri\work\github\ets\enable\enable\qt4\base_window.py", line 78, in paintEvent
    self._enable_window._paint(event)
  File "c:\users\rporuri\work\github\ets\enable\enable\abstract_window.py", line 538, in _paint
    self.component.draw(gc, view_bounds=(0, 0, size[0], size[1]))
  File "c:\users\rporuri\work\github\ets\enable\enable\component.py", line 418, in draw
    self._draw(gc, view_bounds, mode)
  File "c:\users\rporuri\work\github\ets\enable\enable\component.py", line 799, in _draw
    self._dispatch_draw(layer, gc, view_bounds, mode)
  File "c:\users\rporuri\work\github\ets\enable\enable\container.py", line 280, in _dispatch_draw
    component._dispatch_draw(layer, gc, new_bounds, mode)
  File "c:\users\rporuri\work\github\ets\enable\enable\component.py", line 817, in _dispatch_draw
    handler(gc, view_bounds, mode)
  File "c:\users\rporuri\work\github\ets\chaco\chaco\base_xy_plot.py", line 463, in _draw_plot
    self._draw_component(gc, view_bounds, mode)
  File "c:\users\rporuri\work\github\ets\chaco\chaco\base_xy_plot.py", line 471, in _draw_component
    self._render(gc, pts)
  File "c:\users\rporuri\work\github\ets\chaco\chaco\lineplot.py", line 325, in _render
    render(gc, points, self.orientation)
  File "c:\users\rporuri\work\github\ets\chaco\chaco\lineplot.py", line 336, in _render_normal
    gc.stroke_path()
  File "c:\users\rporuri\work\github\ets\enable\kiva\agg\agg.py", line 1360, in stroke_path
    return _agg.GraphicsContextArray_stroke_path(self, *args)
OverflowError: Exceeded cell block limit
```

kiva/tests/test_agg_drawing.py
jwiggins: The issue could be fixed by adding a path reset to the exception handler.

jwiggins: Done

kiva/tests/drawing_tester.py
```diff
@@ -184,8 +184,8 @@ def save_and_return_dpi(self):
 
         filename = "{0}.png".format(self.filename)
         self.gc.save(filename)
-        image = Image.open(filename)
-        dpi = image.info['dpi']
+        with Image.open(filename) as image:
+            dpi = image.info['dpi']
```
jwiggins, line 188: Drive-by fix. This caused a `ResourceWarning` when I was running the tests (which I hadn't seen before).

jwiggins:
Thanks for the review/testing

PR #755, opened 2021-03-24
Require a minimum of Traits 6.2.0

jwiggins:
Resolves #733

PR #735, opened 2021-03-22
Set encoding when saving image from SVG backend.

rahulporuri:
fixes #734 

This PR fixes an issue with the SVG backend where saving to an image when running the enable benchmark was failing with an encoding issue. This was fixed by simply setting the encoding when saving the image.

PR #732, opened 2021-03-19
Rearrange the documentation

jwiggins:
Closes #720

PR #730, opened 2021-03-19
Add documentation for ComponentEditor

jwiggins:
Closes #627

PR #756, opened 2021-03-24
More cherry picks for 5.1.0

jwiggins:
These are cherry-picks of recent PRs which I think should be included in 5.1.0. Please advise if you disagree or have other suggestions. PRs: #735, #745, #749, #755, #751


rahulporuri:
@jwiggins aren't there more PRs that we need to cherrypick to the maintenance branch? What about #748 , #746 , #732, #729  and #730 ?

jwiggins:
I wasn't sure if the documentation PRs mattered or not... I suppose it's better practice to do the release docs build from the release tag. The two code change PRs you reference are probably safe as well.

rahulporuri (approved):
LGTM. It looks like you'll need to update the changelog again 😅

PR #757, opened 2021-03-24
Don't try to import from enable in setup.py

### LINKED ISSUE #486: In setup.py __version__, __requirements__, etc. are being imported
This line is the problem:
https://github.com/enthought/enable/blob/b328edaeb32587078a21bbde1cbc61a521c29807/setup.py#L238

This is fragile, because it requires the setup.py to be run from the top level source directory or it will fail (no `enable` in the environment) or worse provide the wrong information (a different `enable` installed in the environment).

A better approach is to explicitly read in the data from the file. See how TraitsUI does it: https://github.com/enthought/traitsui/blob/master/setup.py#L223-L287

jwiggins:
Resolves #486

I didn't do this exactly like TraitsUI (as suggested in the issue). Instead, I opted to use `runpy.run_path` everywhere.

### DIFF
--- a/enable/__init__.py
+++ b/enable/__init__.py
@@ -10,7 +10,10 @@
 """ A multi-platform object drawing library.
     Part of the Enable project of the Enthought Tool Suite.
 """
-from ._version import full_version as __version__
+try:
+    from enable._version import full_version as __version__
+except ImportError:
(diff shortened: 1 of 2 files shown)


rahulporuri (approved):
LGTM.

PR #760, opened 2021-03-25
Release 5.1.0rc2

jwiggins:
One more and then we probably do the real 5.1.0.

I've already added the tag for this prerelease.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,8 +23,8 @@ from setuptools.command.install import install as base_install
 MAJOR = 5
 MINOR = 1
 MICRO = 0
-PRERELEASE = "rc2"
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM. I'm guessing this won't be squash merged?

jwiggins:
That's right. Always rebase & merge on the release branches

jwiggins:
Doh! Rebase and Merge really does rebase (unsurprisingly), but that means our 5.1.0rc2 tag is pointing at a different commit. The result will still be the same if you build it, but it's awkward.

NOTE: **Don't tag before merging**

PR #761, opened 2021-03-25
Create a Blend2D Kiva backend

jwiggins:
This is a bit experimental and unfinished, mostly due to the incomplete state of the blend2d-python wrappers. That said, there's some promise here. I'm only exposing the backend via the `null` toolkit, but it's covered by the `enable.gcbench` suite. It fails most of the benchmarks. But its performance for the `draw_path` benchmark should encourage us to fill out the remaining functionality and see how it stacks up against the other backends.

### DIFF
--- a/kiva/blend2d.py
+++ b/kiva/blend2d.py
@@ -65,6 +65,8 @@ class GraphicsContext(object):
         self.base_scale = kwargs.pop("base_pixel_scale", 1)
         self.gc.translate(0, size[1])
         self.gc.scale(self.base_scale, -self.base_scale)
+        # Lock it in
+        self.gc.user_to_meta()
 
     # ----------------------------------------------------------------
     # Size info
@@ -119,7 +121,8 @@ class GraphicsContext(object):
     def get_ctm(self):
         """ Return the current coordinate transform matrix.
         """
-        raise NotImplementedError()
+        # XXX: Not a useful return value
+        return self.gc.user_matrix()
 
     # ----------------------------------------------------------------
     # Save/Restore graphics state.
@@ -255,7 +258,7 @@ class GraphicsContext(object):
     def begin_path(self):
         """ Clear the current drawing path and begin a new one.
         """
-        self.path.reset()
+        self.path.clear()
 
     def move_to(self, x, y):
         """ Start a new drawing subpath at place the current point at (x, y).
@@ -289,13 +292,13 @@ class GraphicsContext(object):
     def rect(self, x, y, sx, sy):
         """ Add a rectangle as a new subpath.
         """
-        self.path.rect(x, y, sx, sy)
+        self.path.add_rect(x, y, sx, sy)
 
     def rects(self, rects):
         """ Add multiple rectangles as separate subpaths to the path.
         """
         for rect in rects:
-            self.path.rect(rect)
+            self.path.add_rect(rect)
 
     def draw_rect(self, rect, mode=constants.FILL_STROKE):
         """ Draw a rect.
@@ -323,14 +326,14 @@ class GraphicsContext(object):
         self.path.quadric_to(cpx, cpy, x, y)
 
     def arc(self, x, y, radius, start_angle, end_angle, clockwise=False):
-        # XXX: Only circles are supported for now
-        if math.fabs(end_angle - start_angle) >= math.pi:
-            self.path.ellipse(x, y, radius, radius)
-        else:
-            raise NotImplementedError()
+        self.path.arc_to(
+            x, y, radius, radius,
+            start_angle, math.fabs(end_angle-start_angle),
+            forceMoveTo=True
+        )
 
     def arc_to(self, x1, y1, x2, y2, radius):
-        raise NotImplementedError()
+        self.path.arc_quadrant_to(x1, y1, x2, y2)
 
     # ----------------------------------------------------------------
     # Getting infomration on paths
@@ -339,17 +342,19 @@ class GraphicsContext(object):
     def is_path_empty(self):
         """ Test to see if the current drawing path is empty
         """
-        raise NotImplementedError()
+        return self.path.empty()
 
     def get_path_current_point(self):
         """ Return the current point from the graphics context.
         """
-        raise NotImplementedError()
+        return self.path.get_last_vertex()
 
     def get_path_bounding_box(self):
         """ Return the bounding box for the current path object.
         """
-        raise NotImplementedError()
+        # XXX: Returns a blend2d.Rect which is sort of useless...
+        self.path.get_bounding_box()
+
 
     # ----------------------------------------------------------------
     # Clipping path manipulation
@@ -370,7 +375,7 @@ class GraphicsContext(object):
 
             Region should be a 4-tuple or a sequence.
         """
-        raise NotImplementedError()
+        self.gc.clip_to_rect(blend2d.Rect(x, y, w, h))
 
     def clip_to_rects(self, rects):
         """ Clip context to a collection of rectangles
@@ -477,13 +482,12 @@ class GraphicsContext(object):
             warnings.warn(msg.format(type(img)))
             return
 
-        # XXX: Unscaled!
         # XXX: Upside down!
-        x, y, *_ = rect
+        dst_rect = blend2d.Rect(*rect)
         w, h = img.width, img.height
         image = blend2d.Image(img_array)
         rect = blend2d.Rect(0, 0, w, h)
-        self.gc.blit_image((x, y), image, rect)
+        self.gc.blit_scaled_image(dst_rect, image, rect)
 
     # ----------------------------------------------------------------
     # Drawing Text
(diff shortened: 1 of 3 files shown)


jwiggins:
Pushed a small update which fixed a few problems:
* Now uses a bottom left origin
* Fixed image drawing
* Fixed color selection in jwiggins/blend2d-python#17
* Fixed saving (BGRA -> RGBA)

The performance is quite good (when supported):

Draw Function | kiva.agg | blend2d
-- | -- | --
draw_image | 1.00 | 1.30
draw_image_2x | 1.00 | 1.33
draw_marker_at_points | 1.00 | ✘
draw_marker_at_points_2x | 1.00 | ✘
draw_path | 1.00 | 1.18
draw_path_2x | 1.00 | 1.21
draw_path_at_points | 1.00 | ✘
draw_path_at_points_2x | 1.00 | ✘
draw_path_linear_gradient | 1.00 | ✘
draw_path_linear_gradient_2x | 1.00 | ✘
draw_rect | 1.00 | 1.26
draw_rect_2x | 1.00 | 1.52
show_text | 1.00 | 0.07
show_text_2x | 1.00 | 0.08
show_text_radial_gradient | 1.00 | ✘
show_text_radial_gradient_2x | 1.00 | ✘

jwiggins:
Fixed a few missing features. Note that blend2d gains performance on AGG as the output gets bigger.

Draw Function | kiva.agg | blend2d
-- | -- | --
draw_image | 1.00 | 1.14
draw_image_2x | 1.00 | 1.29
draw_marker_at_points | 1.00 | ✘
draw_marker_at_points_2x | 1.00 | ✘
draw_path | 1.00 | 1.09
draw_path_2x | 1.00 | 1.11
draw_path_at_points | 1.00 | ✘
draw_path_at_points_2x | 1.00 | ✘
draw_path_linear_gradient | 1.00 | 1.71
draw_path_linear_gradient_2x | 1.00 | 2.79
draw_rect | 1.00 | 1.34
draw_rect_2x | 1.00 | 1.77
show_text | 1.00 | 0.20
show_text_2x | 1.00 | 0.24
show_text_radial_gradient | 1.00 | ✘
show_text_radial_gradient_2x | 1.00 | ✘

jwiggins:
Baby steps:


Draw Function | kiva.agg | blend2d
-- | -- | --
draw_image | 1.00 | 1.29
draw_image_2x | 1.00 | 1.30
draw_marker_at_points | 1.00 | ✘
draw_marker_at_points_2x | 1.00 | ✘
draw_path | 1.00 | 1.17
draw_path_2x | 1.00 | 1.21
draw_path_at_points | 1.00 | ✘
draw_path_at_points_2x | 1.00 | ✘
draw_path_linear_gradient | 1.00 | 1.69
draw_path_linear_gradient_2x | 1.00 | 2.88
draw_rect | 1.00 | 1.38
draw_rect_2x | 1.00 | 1.64
show_text | 1.00 | 0.19
show_text_2x | 1.00 | 0.22
show_text_radial_gradient | 1.00 | 0.05
show_text_radial_gradient_2x | 1.00 | 0.06

kiva/blend2d.py
```diff
@@ -0,0 +1,780 @@
...
+
+    def set_alpha(self, alpha):
+        self.gc.set_alpha(alpha)
+
+    # ----------------------------------------------------------------
+    # Gradients
+    # ----------------------------------------------------------------
+
+    def linear_gradient(self, x1, y1, x2, y2, stops, spread_method,
+                        units="userSpaceOnUse"):
```
rahulporuri, line 424: It looks like the interface centers around the needs of agg/celiagg and some/most backends don't care about some of the arguments passed to methods on the interface - `blend2d` doesn't care about `units`.

kiva/blend2d.py
```diff
@@ -0,0 +1,780 @@
...
+            img = Image.fromarray(img.bmp_array)
+            img = normalize_image(img)
+            img_array = np.array(img)
+        else:
+            msg = "Cannot render image of type '{}' into blend2d context."
+            warnings.warn(msg.format(type(img)))
+            return
+
+        # XXX: Upside down!
+        dst_rect = blend2d.Rect(*rect)
```
rahulporuri, line 487: I don't think this will work in the default case where `rect=None` - I don't know if that default case is ever exercised in reality.

kiva/blend2d.py
```diff
@@ -0,0 +1,780 @@
...
+        """ Set the font for the current graphics context.
+        """
+        self.set_font(Font(face_name, size=size, style=style))
+
+    def set_font(self, font):
+        """ Set the font for the current graphics context.
+        """
+        spec = font.findfont()
+        self._kiva_font = font
+        self.font = blend2d.Font(spec.filename, font.size)
```
rahulporuri, line 507: dunno if `blend2d.Font` supports `family` which is being introduced in a separate PR.

kiva/blend2d.py
```diff
@@ -0,0 +1,780 @@
...
+        rect - a tuple (x, y, w, h)
+        """
+        from PIL import Image
+
+        def normalize_image(img):
+            if not img.mode.startswith("RGB"):
+                img = img.convert("RGB")
+            return img
+
+        if isinstance(img, np.ndarray):
```
rahulporuri, line 464: This code is now more or less repeated across multiple backends - apart from a few extra clauses I think. It might be worth thinking about refactoring them into a single place - maybe put somewhere in `kiva.abstract_graphics_context`.

kiva/blend2d.py
```diff
@@ -0,0 +1,780 @@
...
+        self.path.clear()
+
+    def move_to(self, x, y):
+        self.path.move_to(x, y)
+
+    def arc(self, x, y, r, start_angle, end_angle, clockwise=False):
+        self.path.arc(x, y, r, start_angle, end_angle, clockwise)
+
+    def arc_to(self, x1, y1, x2, y2, r):
+        raise NotImplementedError()
```
rahulporuri, line 728: It's not clear to me why `GraphicsContext.arc` uses `Path.arc_to` but `CompiledPath.arc` uses `Path.arc` and `CompiledPath.arc_to` is unimplemented even though `Path.arc_to` exists.

I'm sorry if this is obvious.

kiva/blend2d.py
```diff
@@ -0,0 +1,780 @@
...
+    def copy(self):
+        raise NotImplementedError()
+
+    def begin_path(self):
+        self.path.clear()
+
+    def move_to(self, x, y):
+        self.path.move_to(x, y)
+
+    def arc(self, x, y, r, start_angle, end_angle, clockwise=False):
```
rahulporuri, line 724: nit : looks like agg uses `cw` instead of `clockwise` for this interface. Again, not sure what we want to do about it - if we want to switch this to `cw` or if we want to force everyone to use `clockwise` (i'd prefer the latter but that might be a lengthy deprecation/move)

rahulporuri (approved):
Mostly LGTM with a bunch of questions/comments.

Also, just to confirm, we're not exposing the blend2d backend to end users at the moment i.e. they can't select the backend using `ETS_TOOLKIT` or `toolkit_object`. Right?

kiva/blend2d.py
```diff
@@ -0,0 +1,780 @@
...
+        """ Start a new drawing subpath at place the current point at (x, y).
+        """
+        self.path.move_to(x, y)
+
+    def line_to(self, x, y):
+        """ Add a line from the current point to the given point (x, y).
+
+            The current point is moved to (x, y).
+        """
+        self.path.line_to(x, y)
```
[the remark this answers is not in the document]
jwiggins: This should probably be done elsewhere.

kiva/blend2d.py
jwiggins: `units` is relevant. It's only ignored by backends because it's tricky to implement. For `"objectBoundingBox"` you need to compute the bounding box of the path being filled by the gradient and then set the transformation matrix for the fill to fit that box.

kiva/blend2d.py
jwiggins: You're right, that will raise an exception.

kiva/blend2d.py
jwiggins: It's OK, it's not relevant here. I think `family` is only useful for APIs which handle the font lookup for us.

kiva/blend2d.py
jwiggins: We should open an issue.

kiva/blend2d.py
jwiggins: I believe I fixed this. The arc functions provided by blend2d are slightly different than what Kiva expects, so they're not implemented yet.

kiva/blend2d.py
jwiggins: This backend started as a copy of `celiagg` (which was a copy of `qpainter`) so we'll need to unify those. This should be added to #648

rahulporuri:
Still LGTM, barring the issues that need to be opened.

jwiggins:
Alright then. Issues to follow. Thanks for the feedback.

jwiggins:
#771 opened for blend2d-specific issues

Issue #762, opened 2021-03-26
Provide better support for mixed writing systems


jwiggins:
This is mainly a problem of the AGG backends. Quartz and QPainter backends already have the correct behavior.

Basically, user code should be able to call `show_text` on the following string: `"Kiva Graphics一番😎"` and have it render correctly even if the currently selected font only supports Latin characters.

To get to this point, we need to do a few things:
- [ ] Collect writing system information for entries in our font database
- [ ] Build fallback lists for font families and styles
- [ ] Find a library to use, or failing that, write our own function which breaks a string up into chunks which share the same writing system
- [ ] Make low level text drawing functions return the text cursor position after drawing a run of glyphs (or work around the absence by calling `get_text_extent` on every chunk of a string before drawing)
- [ ] Bring everything together in the `show_text` method so that mixed strings can be drawn
- [ ] Bonus: Support bidirectional text mixing

This is roughly what Qt does, based on a quick skim of the code: https://code.qt.io/cgit/qt/qtbase.git/
* `QFreeTypeFontDatabase::addTTFile` (qtbase.git/tree/src/gui/text/freetype/qfreetypefontdatabase.cpp):
  Scans a font for the following information: weight, style, fixed-width, supported writing systems (unicode range, codepage range), family name
* `QPlatformFontDatabase::fallbacksForFamily` (qtbase.git/tree/src/gui/text/qfontdatabase.cpp):
  Takes a style and script ID and returns a list of fonts which support that script with that style (or just support the script)
* `QPainter::drawText` (qtbase.git/tree/src/gui/painting/qpainter.cpp): Basically Qt's `show_text`.
  Uses `QStackTextEngine` for shaping, breaking of input string. Breaks into `QScriptItem` objects. Picks the font _per_ item and draws it.
* `QStackTextEngine`/`QScriptItem`/`QTextItemInt` (qtbase.git/tree/src/gui/text/qtextengine.cpp)
  These are the components which break up a string into chunks which can be shaped and drawn as a unit.

jwiggins:
Lovely: https://raphlinus.github.io/rust/skribo/text/2019/04/04/font-fallback.html

jwiggins:
Copying from #767 so it's easier to find:

[...]

PR #763, opened 2021-03-29
Add support for determining the language of a font

jwiggins:
First step towards #762

### DIFF
--- a/kiva/fonttools/_util.py
+++ b/kiva/fonttools/_util.py
@@ -44,10 +44,80 @@ _name_ids = {
     5: "version",
     6: "postscript_name",
 }
+# OpenType 'OS/2' `ulCodePageRange` bit meanings
+_ot_code_page_masks = {
+    "Latin": 0x93,
+    "Cyrillic": 0x4,
+    "Greek": 0x8,
+    "Hebrew": 0x20,
+    "Arabic": 0x40,
+    "Vietnamese": 0x100,
+    "Thai": 0x10000,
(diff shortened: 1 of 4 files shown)


kiva/fonttools/_util.py
```diff
@@ -44,10 +44,80 @@
     5: "version",
     6: "postscript_name",
 }
+# OpenType 'OS/2' `ulCodePageRange` bit meanings
```
rahulporuri, line 47: can you add a reference here - something like this https://docs.microsoft.com/en-us/typography/opentype/spec/os2#cpr

kiva/fonttools/_util.py
```diff
@@ -44,10 +44,80 @@
...
+    "Arabic": 0x40,
+    "Vietnamese": 0x100,
+    "Thai": 0x10000,
+    "Japanese": 0x20000,
+    "Simplified Chinese": 0x40000,
+    "Traditional Chinese": 0x100000,
+    "Korean": 0x280000,
+    "Symbol": 0x80000000,
+}
+# OpenType 'OS/2' table `ulUnicodeRange` bit meanings
```
rahulporuri, line 62: and another reference here, like - https://docs.microsoft.com/en-us/typography/opentype/spec/os2#ur

kiva/fonttools/_util.py
jwiggins: Yeah, I was thinking the same

jwiggins:
Thanks for the feedback

PR #764, opened 2021-03-29
Basic analysis of the script content of strings

jwiggins:
This is another part of #762

There are two pieces here, and a ton of machine-generated code (don't fear the diff):
* A small program which fetches http://www.unicode.org/Public/UNIDATA/Scripts.txt, parses it, and writes out the `kiva.fonttools.text._data` module.
* A new class `UnicodeAnalyzer` which uses the data from `Scripts.txt` and returns the languages and slices for a given input string.

`UnicodeAnalyzer` is pretty basic right now. I'd like to keep it that way in this PR. For instance, Emoji ligatures are not great:
```python
In [3]: s = "👩‍👩‍👧‍👧" 

In [4]: s                                                                       
Out[4]: '👩\u200d👩\u200d👧\u200d👧'

In [5]: an.languages(s)                                                         
Out[5]: 
[(0, 1, 'Common'),
 (1, 2, 'Inherited'),
 (2, 3, 'Common'),
 (3, 4, 'Inherited'),
 (4, 5, 'Common'),
 (5, 6, 'Inherited'),
 (6, 7, 'Common')]
```


kiva/fonttools/_util.py
```diff
@@ -102,16 +102,16 @@
...
     79: "Runic",
     80: "Khmer",
     81: "Mongolian",
     86: "Gothic",
     87: "Deseret",
     93: "Limbu",
-    94: "Tai Le",
-    95: "New Tai Lue",
+    94: "Tai_Le",
+    95: "New_Tai_Lue",
```
jwiggins, line 114: These were changed to match the names in `kiva.fonttools.text._data`.

kiva/fonttools/text/_unicode_lookup.py
```diff
@@ -0,0 +1,46 @@
...
+                last_lang = lang
+                last_start = idx
+
+        result.append((last_start, idx + 1, last_lang))
+
+        return result
+
+    def _lookup_codepoint(self, cp):
+        comps = self.ranges - ord(cp)
+        index = ((comps[:, 0] <= 0) == (comps[:, 1] >= 0)).argmax()
```
rahulporuri, line 45: This looks like the most important detail in the PR - how we're selecting the entry given a code point - and it'd be useful if you could elaborate on how we're doing it.

rahulporuri (approved):
LGTM

kiva/fonttools/text/_unicode_lookup.py
jwiggins: Fair enough

rahulporuri:
still LGTM

jwiggins:
Thanks for the review

PR #766, opened 2021-03-30
Fix listeners hooked up using observe but unhooked using on_trait_change

rahulporuri:
This PR fixes two listeners that are hooked up using `observe` but are then unhooked using `on_trait_change`.

### DIFF
--- a/enable/scrolled.py
+++ b/enable/scrolled.py
@@ -487,13 +487,13 @@ class Scrolled(Container):
     def _release_sb(self, sb):
         if sb is not None:
             if sb == self._vsb:
-                sb.on_trait_change(
+                sb.observe(
                     self._handle_vertical_scroll,
                     "scroll_position",
                     remove=True,
                 )
             if sb == self._hsb:
-                sb.on_trait_change(
+                sb.observe(
                     self._handle_horizontal_scroll,
                     "scroll_position",
                     remove=True,


aaronayres35 (approved):
These changes LGTM

Looking closer at what is hooked/unhooked it looks like we do hook things up to the `_handle_mouse_thumb` method but I don't see any removals 🤔 (w/ on_trait_change or with observe).  In any case, that could be addressed in a separate PR

rahulporuri:
Im looking at the `_handle_mouse_thumb` method and im not sure who is responsible for unhooking it. It looks like it is getting hooked up alongside  `_handle_vertcal_scroll` and `_handle_vertical_scroll` but im not sure if `_release_sb` is the right place/time to unhook the listeners for `_handle_mouse_thumb`. I'm merging this for now.

aaronayres35:
> Im looking at the `_handle_mouse_thumb` method and im not sure who is responsible for unhooking it. It looks like it is getting hooked up alongside `_handle_vertcal_scroll` and `_handle_vertical_scroll` but im not sure if `_release_sb` is the right place/time to unhook the listeners for `_handle_mouse_thumb`. I'm merging this for now.

SGTM, I was in the same boat.  It seems like `_release_sb` makes the most sense as they are handlers hooked up to scroll bars, but I too wasn't certain. We can address this separately

PR #767, opened 2021-03-30
Create a mapping from script to font language

jwiggins:
This is another step for #762

The PR adds support for converting the script name generated by `UnicodeAnalyzer` and mapping it to a language name which a font reports support for.

### DIFF
--- a/kiva/fonttools/text/_language.py
+++ b/kiva/fonttools/text/_language.py
@@ -1,0 +1,74 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM

kiva/fonttools/text/_language.py
```diff
@@ -0,0 +1,74 @@
...
+        locale_lang = "en_US"
+
+    # Pick a language to use for "Han" script
+    han_lang = "Traditional Chinese"  # Default
+    if locale_lang in ("zh_CN", "zh_SG"):
+        han_lang = "Simplified Chinese"
+    elif locale_lang.startswith("ja"):
+        han_lang = "Japanese"
+    elif locale_lang.startswith("ko"):
+        han_lang = "Korean"
```
jwiggins, line 48: Having played with this a bit more, we should only use this choice when it's not otherwise clear from the context. For instance if a string already contains Hiragana or Katakana, then Han should be mapped to "Japanese". If Hangul is encountered, Han maps to "Korean". Only if the Han is mixed with some non-CJK language should we fall back to this locale-based guess.

jwiggins:
Thanks for the review

PR #768, opened 2021-03-30
Add support for font fallback lookups

jwiggins:
This is a big part of #762: Give a way to lookup a font which "matches" a query and supports a particular language. Builds on top of the work done in #763

### DIFF
--- a/kiva/fonttools/font_manager.py
+++ b/kiva/fonttools/font_manager.py
@@ -76,7 +76,7 @@ class FontManager:
     # Increment this version number whenever the font cache data
     # format or behavior has changed and requires a existing font
     # cache files to be rebuilt.
-    __version__ = 11
+    __version__ = 12
 
     def __init__(self):
         self._version = self.__version__
@@ -120,6 +120,8 @@ class FontManager:
 
         self.ttf_lookup_cache = {}
         self.afm_lookup_cache = {}
+        self.ttf_fallback_cache = {}
+        self.afm_fallback_cache = {}
 
     def update_fonts(self, paths):
         """ Update the font lists with new font files.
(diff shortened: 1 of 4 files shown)


kiva/fonttools/font_manager.py
```diff
@@ -120,6 +120,8 @@ def __init__(self):
 
         self.ttf_lookup_cache = {}
         self.afm_lookup_cache = {}
+        self.ttf_fallback_cache = {}
+        self.afm_fallback_cache = {}
```
rahulporuri, line 124: I should've bought this up earlier but there's no good reason to make these public attributes. Is there?

kiva/fonttools/font_manager.py
```diff
@@ -290,6 +329,25 @@ def __repr__(self):
         return result
 
 
+class FontSpec(object):
```
rahulporuri, line 332: Maybe make this a private class - i don't think there are users of the class outside this module or outside `kiva.fonttools.*`

rahulporuri (approved):
LGTM with a couple of minor suggestions

kiva/fonttools/tests/test_font.py
```diff
@@ -54,6 +54,18 @@ def test_find_font_name(self):
         # Name should be nonempty.
         self.assertGreater(len(name), 0)
 
+    def test_find_font_for_language(self):
```
rahulporuri, line 57: it'd also be useful to have a test that exercises `font_fallback`

kiva/fonttools/font_manager.py
jwiggins: No reason, just following what was already there.

kiva/fonttools/font_manager.py
jwiggins: Good point

jwiggins:
Thanks for the feedback

PR #778, opened 2021-04-16
Handle preferred size computation when container isnt resizable

rahulporuri:
Ref ~https://github.com/enthought/chaco/pull/673/~. https://github.com/enthought/chaco/pull/673 got closed in favor of https://github.com/enthought/chaco/pull/674

It looks like these lines were commented out from the original commit when this code was added to enable in https://github.com/enthought/enable/commit/75fb3ac9e478751992865d98d8f27791e7a988df
and subsequently, the commented out code was removed in the PR #466

### DIFF
--- a/enable/stacked_layout.py
+++ b/enable/stacked_layout.py
@@ -10,8 +10,6 @@
 """ Routines for stacked layout of components in a container
 """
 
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM

I see the same code in chaco.  Unclear to me why it had been commented out.

PR #780, opened 2021-04-16
Fix alignment trait on "HStackedContainer"

rahulporuri:
This PR fixes the alignment trait and `_do_layout` method on the `HStackedContainer`. This might have been a simple copy-paste mistake. This was discovered when working on https://github.com/enthought/chaco/pull/674.

This PR simply renames the `halign` triat with `valign` and updates its usage in the `_do_layout` method. See related https://github.com/enthought/chaco/blob/a88b7b96a981dc96ad9f0d615c21d699f93851d1/chaco/plot_containers.py#L289-L324, which is possibly where the `HStackedContainer` class originated.

### DIFF
--- a/enable/stacked_container.py
+++ b/enable/stacked_container.py
@@ -52,10 +52,10 @@ class HStackedContainer(StackedContainer):
     # Overrides StackedPlotContainer.
     stack_index = 0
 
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM

Issue #716, opened 2021-03-13
CustomMarker add_path and get_compiled_path likely to fail on some backends


corranwebster:
This issue could alternatively be "Compiled Path API inconsistent between backends".

The issue here is that both of these methods on `CustomMarker` try to provide a copy of the custom marker's path scaled appropriately for the size of the marker.  This relies on a call to `CompiledPath.scale_ctm` which is not available on all backends (eg. QPainter backend, celiagg backend) and in some cases (eg. QPainter) there doesn't look to be an easy way to provide that support.

As a result the call will fail.

In some cases `add_path` may still work in use because it is being used by passing in a GC instead of a path.

jwiggins:
I don't think `CompiledPath` has any reason to contain a transformation matrix.

That said, I think this should give the same result across backends:
```python
from math import pi
with gc:
    gc.set_line_width(7)
    gc.set_stroke_color((1.0, 0.0, 1.0, 1.0))

    with gc:
        gc.translate_ctm(50, 50)
        for i in range(0, 12):
            gc.translate_ctm(20, 0)
            gc.rotate_ctm(2*pi/12.0)
            gc.move_to(0, 0)
            gc.line_to(20, 0)
        gc.stroke_path()
```

That is, when you're modifying the implicit path in a `GraphicsContext` the current transformation matrix is _absolutely relevant_. But `kiva.agg` can't have nice things (#390) because it does path transformation too early.

I don't think this necessarily precludes `AbstractMarker.add_to_path` from scaling markers.

jwiggins:
Right. I played around with this a bit in celiagg... I don't think it's actually possible to do this for the implicit path. We could possibly add it to `CompiledPath`, but even that might be too confusing. Based on the intractability of #390 for `kiva.agg`, I see this as a design bug.

PR #782, opened 2021-04-16
Update custom marker as scale_ctm expects floats

aaronayres35:
Needed to close https://github.com/enthought/chaco/issues/232

Down in agg the method that gets called wants double inputs.  If an int gets passed it things break.  Changing the inputs to floats seems to resolve the problem.

### DIFF
--- a/enable/tests/test_markers.py
+++ b/enable/tests/test_markers.py
@@ -1,0 +1,48 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+import unittest
+
+import numpy as np
+
+from enable.compiled_path import CompiledPath
+from enable.kiva_graphics_context import GraphicsContext
+from enable.markers import CustomMarker
+
+
+class TestCustomMarker(unittest.TestCase):
+
+    def test_add_to_path(self):
+        path = CompiledPath()
+        path.begin_path()
+        path.move_to(-5, -5)
+        path.line_to(-5, 5)
+        path.line_to(5, 5)
+        path.line_to(5, -5)
+        path.line_to(-5, -5)
+
+        marker = CustomMarker(path=path)
+
+        gc = GraphicsContext((50, 50))
+        # should not fail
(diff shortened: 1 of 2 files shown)


rahulporuri:
I don't see any unittests for the `CustomMarker` code so let's take advantage of this opportunity to add some.

rahulporuri (approved):
LGTM. Ref https://github.com/enthought/enable/blob/a1a60592accb921b625cc5883ee4f02416ac85e2/kiva/agg/src/kiva_compiled_path.h#L124.

I'd still like it if we could add a few tests.

aaronayres35:
CI is failing with 
```
ERROR: test_get_compiled_path (enable.tests.test_markers.TestCustomMarker)
  File "/Users/runner/work/enable/enable/.edm/envs/enable-test-3.6-wx/lib/python3.6/site-packages/enable/tests/test_markers.py", line 50, in test_get_compiled_path
    marker.get_compiled_path(np.int64(0))
  File "/Users/runner/work/enable/enable/.edm/envs/enable-test-3.6-wx/lib/python3.6/site-packages/enable/markers.py", line 424, in get_compiled_path
    newpath.scale_ctm(float(size), float(size))
AttributeError: 'kiva.quartz.ABCGI.CGMutablePath' object has no attribute 'scale_ctm'

```
perhaps this test just needs to be skipped on some combination of quartz / macos / wx?  I am marking this as a WIP until I get a chance to look into this more

enable/tests/test_markers.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# Thanks for using Enthought open source!
+import unittest
+
+import numpy as np
+
+from enable.compiled_path import CompiledPath
+from enable.kiva_graphics_context import GraphicsContext
+from enable.markers import CustomMarker
+from enable.tests._testing import skip_if_not_qt
```
rahulporuri, line 17: not an actionable comment - i was wondering if the enable testsuite should always import from `enable.api` instead of importing from the specific modules. Note that this is a more general question for ETS/package development.

rahulporuri (approved):
Still LGTM

enable/tests/test_markers.py
aaronayres35: Yeah I notice some of the test files do but others don't.  I think it probably makes sense to use api everywhere in tests

jwiggins:
> perhaps this test just needs to be skipped on some combination of quartz / macos / wx? I am marking this as a WIP until I get a chance to look into this more

This behavior of `CompiledPath` is only possible with `kiva.agg` due to a design defect (see #716)

PR #784, opened 2021-04-19
Convert # to #: to improve docstrings in markers

rahulporuri:
This PR converts `#` comments for traits to `#:` comments to improve the documentation in the `enable.markers` module. This PR also removes unnecessary trait definitions `antialias` for some markers which don't modify the default that they inherit from `AbstractMarker`.

### DIFF
--- a/enable/markers.py
+++ b/enable/markers.py
@@ -31,19 +31,18 @@ class AbstractMarker(HasTraits):
     """ Abstract class for markers.
     """
 
-    # How this marker is to be stroked (from kiva.api).
+    #: How this marker is to be stroked (from kiva.api).
     # Since this needs to be a class variable, it can't be a trait.
     draw_mode = STROKE
-    # draw_mode = Enum(FILL, EOF_FILL, STROKE, FILL_STROKE, EOF_FILL_STROKE)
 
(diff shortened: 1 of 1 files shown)


enable/markers.py
```diff
@@ -323,7 +322,7 @@ class StarMarker(AbstractMarker):
     """ A marker that is a (filled) star.
     """
 
-    # How this marker is to be stroked. (Overrides AbstractMarker.)
+    #: How this marker is to be stroked. (Overrides AbstractMarker.)
     draw_mode = FILL_STROKE
     # Do not render anti-aliased. (Overrides AbstractMarker.)
     antialias = True
```
rahulporuri, line 328: this - and more such examples are ignored in this PR - because the comment strings say one thing (i.e. `antialias = False`) but the actual code says something else (i.e. `antialias = True`). This needs to be investigated further, separately.

rahulporuri: Update : These markers were added in https://github.com/enthought/enable/pull/222 - which flipped `antialias` back to `True` but the PR forgot to remove the outdated comments.

aaronayres35 (approved):
LGTM

PR #785, opened 2021-04-20
Remove Python 2 or 3 only conditionals

rahulporuri:
This PR removes Python 2/3 specific conditionals and `__future__` imports.


aaronayres35 (approved):
LGTM

PR #789, opened 2021-04-21
Update "super" usage

rahulporuri:
This PR updates `super` usage using automated regex-based search and replace. After the changes were made, we reviewed each of the changes to ensure that we were calling the right class before.

See https://github.com/enthought/traits/pull/1280

### DIFF
--- a/enable/container.py
+++ b/enable/container.py
@@ -431,7 +431,7 @@ class Container(Component):
         should be set to True, and contained components will not be called
         with the event.
         """
-        # super(Container, self)._dispatch_stateful_event(event, suffix)
+        # super()._dispatch_stateful_event(event, suffix)
         Component._dispatch_stateful_event(self, event, suffix)
 
     def get_event_transform(self, event=None, suffix=""):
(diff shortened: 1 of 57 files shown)


enable/container.py
```diff
@@ -574,12 +574,12 @@ def _fit_window_changed(self, old, new):
     def _bounds_changed(self, old, new):
         # crappy... calling our parent's handler seems like a common traits
         # event handling problem
```
aaronayres35, line 576: I was thinking this same thing when reviewing recent chaco PRs...

enable/scrolled.py
```diff
@@ -612,7 +612,7 @@ def _container_handle_mouse_event(self, event, suffix):
     # -------------------------------------------------------------------------
 
     def __getstate__(self):
-        state = super(Scrolled, self).__getstate__()
+        state = super().__getstate__()
         for key in ["alternate_vsb", "_vsb", "_hsb"]:
             if key in state:
                 del state[key]
```
aaronayres35, line 618: orthogonal but these could be updated to use `transient` metadata

aaronayres35 (approved):
LGTM

enable/scrolled.py
rahulporuri: i'll open a separate issue/PR for this.

rahulporuri: i created https://github.com/enthought/enable/issues/795

PR #790, opened 2021-04-21
Update super usage in potentially buggy cases

rahulporuri:
See related #789 

This PR updates a few calls to `super` that were potentially buggy. In these two cases for example, we were calling `super(A, self)` from class `B` where `B` inherits from `A`. We first update the `super` calls to use the correct class argument - and then update the `super` call to remove the arguments altogether.

### DIFF
--- a/enable/canvas.py
+++ b/enable/canvas.py
@@ -159,7 +159,7 @@ class Canvas(Container):
                     gc.move_to(x, 0)
                     gc.line_to(x2, 0)
                     gc.stroke_path()
-        super(Container, self)._draw_underlay(gc, view_bounds, mode)
+        super(Canvas, self)._draw_underlay(gc, view_bounds, mode)
 
     def _transform_view_bounds(self, view_bounds):
         # Overload the parent class's implementation to skip visibility test
--- a/enable/canvas.py
+++ b/enable/canvas.py
@@ -159,7 +159,7 @@ class Canvas(Container):
                     gc.move_to(x, 0)
                     gc.line_to(x2, 0)
                     gc.stroke_path()
-        super(Canvas, self)._draw_underlay(gc, view_bounds, mode)
+        super()._draw_underlay(gc, view_bounds, mode)
 
     def _transform_view_bounds(self, view_bounds):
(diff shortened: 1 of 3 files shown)


enable/canvas.py
```diff
@@ -159,7 +159,7 @@ def _draw_underlay(self, gc, view_bounds=None, mode="default"):
                     gc.move_to(x, 0)
                     gc.line_to(x2, 0)
                     gc.stroke_path()
-        super(Container, self)._draw_underlay(gc, view_bounds, mode)
+        super(Canvas, self)._draw_underlay(gc, view_bounds, mode)
```
rahulporuri, line 162:
```suggestion
        super()._draw_underlay(gc, view_bounds, mode)
```

enable/text_grid.py
```diff
@@ -94,7 +94,7 @@ class TextGrid(Component):
     # ------------------------------------------------------------------------
 
     def __init__(self, **kwtraits):
-        super(Component, self).__init__(**kwtraits)
+        super(TextGrid, self).__init__(**kwtraits)
```
rahulporuri, line 97:
```suggestion
        super().__init__(**kwtraits)
```

enable/tests/tools/test_hover_tool.py
```diff
@@ -53,7 +53,7 @@ class GuiTestAssistant(object):
 class HoverToolTestCase(EnableTestAssistant, GuiTestAssistant,
                         unittest.TestCase):
     def setUp(self):
-        super(HoverToolTestCase, self).setUp()
+        GuiTestAssistant.setUp(self)
```
aaronayres35, line 56: Same question I had on the other PR, why exactly do we purposefully avoid super for setUp

aaronayres35 (approved):
LGTM

enable/tests/tools/test_hover_tool.py
rahulporuri: answered elsewhere - part convention, part best practice - especially because we have cases where we need to call `setUp` on more than one base class.

PR #799, opened 2021-04-26
Skip markers tests if not using agg

aaronayres35:
Previously the test was skipped when not on qt.  This worked because it avoided the scenario of using quartz as tthe default backend on wx.  See:
https://github.com/enthought/traits/blob/249d64813b633390616af4e57ceb19a553c250ef/traits/etsconfig/etsconfig.py#L313-L320

Realistically, these tests are specific to the old agg. see comment: https://github.com/enthought/enable/pull/782#issuecomment-825074184

Note we can't yet skip if not "oldagg" because oldagg isn't actually used yet.  Doing so the test would no longer be run.  I left a comment to replace 'image' with 'oldagg' in the `skipIf` once `image` uses celiagg.

### DIFF
--- a/enable/tests/test_markers.py
+++ b/enable/tests/test_markers.py
@@ -11,13 +11,18 @@ import unittest
 
 import numpy as np
 
+from traits.etsconfig.api import ETSConfig
+
 from enable.compiled_path import CompiledPath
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM

PR #800, opened 2021-04-26
Document EnableTestAssistant

### LINKED ISSUE #774: Document EnableTestAssistant
`enable.testing.EnableTestAssistant` is valuable for developers using `enable` who want to test their code. We should add some documentation around the test assistant by extracting information from the enable testsuite.

aaronayres35:
closes #774 

This PR adds a brief section to the documentation describing the `EnableTestAssistant`.  This is effectively meant as a starting point, and we will likely want to expand this documentation going forward.  In any case, just mentioning its existence in the docs is a significant improvement over it being missing entirely.

A next step (in a future PR) would be to add an example test using the test assistant to showcase how to use it and some of the useful features.  I was going to simply link to an existing test, but none of them felt really suited to be a good example for the documentation IMO.  I think it will be useful to write a test specifically for the docs, possibly writing a test for on of the existing examples.

### DIFF
--- a/docs/source/enable/testing.rst
+++ b/docs/source/enable/testing.rst
@@ -1,0 +1,24 @@
+=========================
+Testing Enable Components
+=========================
+
+In order to assist in the testing of enable :class:`~.Component` and
+:class:`~.Interactor` objects, Enable provides the
+:class:`~.EnableTestAssistant` class. This is a mixin class intended to work
+along side :class:`python:unittest.TestCase`. Often times in order to
+effectively test an enable component and its supported interactivity, one needs
+to simulate user interactions such as moving/clicking/dragging the mouse, using
+keys, etc. This involves manually creating the corresponding events with
+appropriate state and dispatching them appropriately. This can be rather
+tedious and :class:`~.EnableTestAssistant` provides a number of helper methods
+to greatly simplify the process, making tests both faster/cleaner to write and
+easier to read. Furthermore, in test scenarios it is ofen unnecessary to launch
+the full application window and as such :class:`~.EnableTestAssistant` provides
+a :meth:`~.create_mock_window` method for simply
+mocking out the window itself. This allows for specifically testing the
+component of interest alone, as is the goal in a unit test. Please see the api
+docs (by clicking :class:`~.EnableTestAssistant`) for the full list of methods
+available.
+
+.. Todo: Add example test. I was going to refer to an existing test, but none
+   of the exissting tests seem very helpful for documentation purposes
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
Mostly LGTM as it is.

I too was looking through enable for uses of `EnableTestAssistant` and the usage i find most compelling is interestingly enough `test_test_assistant` - you could rewrite those tests to demonstrate `EnableTestAssistant` instead of testing it IMO.

docs/source/enable/testing.rst
```diff
@@ -0,0 +1,61 @@
...
+keys, etc. This involves manually creating the corresponding events with
+appropriate state and dispatching them appropriately. This can be rather
+tedious and :class:`~.EnableTestAssistant` provides a number of helper methods
+to greatly simplify the process, making tests both faster/cleaner to write and
+easier to read. Furthermore, in test scenarios it is ofen unnecessary to launch
+the full application window and as such :class:`~.EnableTestAssistant` provides
+a :meth:`~.create_mock_window` method for simply
+mocking out the window itself. This allows for specifically testing the
+component of interest alone, as is the goal in a unit test. Please see the api
+docs (by clicking :class:`~.EnableTestAssistant`) for the full list of methods
```
rahulporuri, line 20:
```suggestion
docs (:class:`~.EnableTestAssistant`) for the full list of methods
```

docs/source/enable/testing.rst
```diff
@@ -0,0 +1,61 @@
...
+available.
+
+.. Todo: Add example test. I was going to refer to an existing test, but none
+   of the exissting tests seem very helpful for documentation purposes
+
```
rahulporuri, line 25:
```suggestion
```

docs/source/enable/testing.rst
```diff
@@ -0,0 +1,61 @@
...
+The following is a Dummy TestCase to showcase some basics of the
+:class:`~.EnableTestAssistant` functionality.  It is not testing things
+users would likely want to test, but it showcases the capabilities / usefulness
+of the test assistant.
```
rahulporuri, line 29: I think this paragraph is too verbose and unnecessary. You should be able to replace all of this with just "here is an example" and be done with it

docs/source/enable/testing.rst
```diff
@@ -0,0 +1,61 @@
...
+
+::
+
+    import unittest
+    from unittest import mock
+
+    from enable.component import Component
```
rahulporuri, line 36: can we use `enable.api` here?

docs/source/enable/testing.rst
```diff
@@ -0,0 +1,61 @@
...
+    from enable.testing import EnableTestAssistant, _MockWindow
```
rahulporuri, line 37: Maybe don't import and use `_MockWindow`. Let's not unnecessarily advertise private objects to end users in the docs.

rahulporuri (approved):
LGTM

PR #804, opened 2021-04-28
remove __getstate__ methods and use transient metadata

### LINKED ISSUE #795: Mark traits as "transient" instead of explicitly overriding traits __getstate__
https://github.com/enthought/enable/blob/685038503500ecf8ffc30f3da68c7e848ae02e6f/enable/scrolled.py#L614-L619

https://github.com/enthought/enable/blob/588cde79a0e7099e2c0368d3e8e2e3664a296e61/enable/abstract_window.py#L555-L560

aaronayres35:
closes #795 

This PR simply replaces the use of `__getstate__` methods by setting `transient` metadata on relevant traits.  `HasTraits` handles this with its own `__getstate__` method appropriately.

EDIT: It also removes the unused `_scroll_origin` attribute on `AbstractWindow`.

### DIFF
--- a/enable/abstract_window.py
+++ b/enable/abstract_window.py
@@ -102,8 +102,6 @@ class AbstractWindow(HasTraits):
     # When exceeding this, the entire window is marked damaged to save memory
     MAX_DAMAGED_REGIONS = 100
 
-    _scroll_origin = Tuple()
-
     # -------------------------------------------------------------------------
     #  Abstract methods that must be implemented by concrete subclasses
     # -------------------------------------------------------------------------
@@ -207,7 +205,6 @@ class AbstractWindow(HasTraits):
     # ------------------------------------------------------------------------
 
     def __init__(self, **traits):
-        self._scroll_origin = (0.0, 0.0)
         self._update_region = None
         self._gc = None
         self._pointer_owner = None
--- a/enable/abstract_window.py
+++ b/enable/abstract_window.py
@@ -42,27 +42,27 @@ class AbstractWindow(HasTraits):
 
     # A reference to the nested component that has focus.  This is part of the
     # manual mechanism for determining keyboard focus.
-    focus_owner = Instance(Interactor)
+    focus_owner = Instance(Interactor, transient=True)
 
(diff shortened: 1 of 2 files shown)


enable/abstract_window.py
```diff
@@ -552,13 +554,6 @@ def _paint(self, event=None):
 
         self._update_region = []
 
-    def __getstate__(self):
-        attribs = ("component", "bgcolor", "overlay", "_scroll_origin")
```
aaronayres35, line 556: `_scroll_origin` is set to (0,0) in the `__init__` method, but I don't see it used anywhere.  A search of the chaco and enable codebases yields 0 results.  Perhaps it can just be removed entirely?

enable/abstract_window.py
```diff
@@ -86,22 +86,24 @@ class AbstractWindow(HasTraits):
...
 
     # The previous component that handled an event.  Used to generate
     # mouse_enter and mouse_leave events.  Right now this can only be
     # None, self.component, or self.overlay.
-    _prev_event_handler = Instance(Component)
+    _prev_event_handler = Instance(Component, transient=True)
 
     # (dx, dy) integer size of the Window.
-    _size = Trait(None, Tuple)
+    _size = Trait(None, Tuple, transient=True)
```
aaronayres35, line 97: I added `transient=True` for all traits that were previously explicitly listed in `__getstate__` to preserve behavior.  Perhaps we should re-evaluate though what actually should/shouldn't be transient

enable/abstract_window.py
jwiggins: Yeah, it should probably just be removed. I also don't see it being used in some older internal codebases.

enable/abstract_window.py
```diff
@@ -81,15 +81,15 @@ class AbstractWindow(HasTraits):
...
 
     # The regions to update upon redraw
     _update_region = Any
```
rahulporuri, line 95:
```suggestion
    _update_region = Any(transient=True)
```

rahulporuri (approved):
LGTM with one minor comment

PR #805, opened 2021-04-28
move example helper modules into examples

### LINKED ISSUE #793: Consider moving "enable.example_*" modules into "enable.examples" submodule
The following modules are used only by the examples so we should consider moving them into `enable.examles` submodule instead of letting them live in the top-level `enable` namespace.

- [`enable.example_application`](https://github.com/enthought/enable/blob/3d138a235547240c98ed8aba1046a3f1c016d42b/enable/example_application.py)
- [`enable.example_canvas`](https://github.com/enthought/enable/blob/3d138a235547240c98ed8aba1046a3f1c016d42b/enable/example_canvas.py)
- [`enable.example_support`](https://github.com/enthought/enable/blob/3d138a235547240c98ed8aba1046a3f1c016d42b/enable/example_support.py)

aaronayres35:
closes #793 
This PR simply moves the `enable.example_application`, `enable.example_canvas`, and `enable.example_support` modules to sit in the `enable/examples` directory.  I then did a VSCode find and replace for `enable.example_support` with `enable.examples.example_support`, etc.

~Note, as mentioned in this comment: https://github.com/enthought/enable/issues/793#issuecomment-828526455 this may be a breaking change elsewhere as it appears we have down stream users who had directly been importing from these modules.  As a result we may not want to push ahead with these changes...~

~Perhaps the old modules should be left around, but simply import from the new modules in `enable/examples` and raise some sort of warning?  A deprecation warning doesn't really feel correct as it seems to me its more of a "you shouldn't be importing from here" kind of warning.~

EDIT: I have added stub modules at the old file locations which simply import objects from the new modules and raise a deprecation warning.  In the deprecation warning, I did not mention these modules were moved, as we don't want users importing them from anywhere! So I just said importing from here is deprecated and the module will be removed in a future release. Perhaps I should say explicitly don't do this?


enable/example_application.py
```diff
@@ -12,64 +12,17 @@
 ===========================
 
 This module provides a simple Pyface application that can be used by examples
-in places where a DemoFrame is insufficient.
+in places where a DemoFrame is insufficient. Note this has been moved to
+sit in enable/examples.  This module is kept for backwards compatibility.
```
rahulporuri, line 16: Let's go with the following for the module docstring - and the warning message as well.

```
Importing from this module is deprecated. Please use `enable.examples.example_application`
instead. This backwards compatibility stub will be removed in Enable 6.0.
```

We need an issue to track the removal of these stubs in Enable 6.0

rahulporuri: let's use the above template for the module docstring and warning message in the other stub modules as well.

enable/example_application.py
```diff
@@ -12,64 +12,17 @@
...
 
 """
+import warnings
 
-from pyface.api import ApplicationWindow, GUI
+from enable.examples.example_application import DemoApplication
```
rahulporuri, line 21:
```suggestion
from enable.examples.example_application import DemoApplication, demo_main
```

enable/examples/example_support.py
```diff
@@ -0,0 +1,57 @@
...
+if not ETSConfig.toolkit:
+    for toolkit, toolkit_module in (("wx", "wx"), ("qt4", "PyQt4")):
+        try:
+            exec("import " + toolkit_module)
+            ETSConfig.toolkit = toolkit
+            break
+        except ImportError:
+            pass
+    else:
+        raise RuntimeError("Can't load wx or qt4 backend for Chaco.")
```
rahulporuri, line 35: I suspect this isn't necessary now.

rahulporuri (approved):
Almost LGTM - with a bunch of comments/suggestions

enable/examples/example_application.py
```diff
@@ -0,0 +1,75 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
```
rahulporuri, line 1: I just realized that these should probably be private modules i.e. `enable.examples._example_application`, `enable.examples._example_support` and `enable.examples._example_canvas`. I say that because I notice external users of these modules on GitHub - https://github.com/search?q=org%3Aenthought+example_support+extension%3Apy&type=code - which I doubt was the original intention of the devs who introduced these modules.

What do you think @aaronayres35 ?

aaronayres35: Yeah I agree, I will rename them on this PR

enable/example_application.py
aaronayres35: Should we say `Please use `enable.examples.example_application` instead.` if we are making the modules private?  Theoretically the suggested change is "stop doing that" as opposed to "update your imports"

rahulporuri: Ahh. You're right. Let's update it to

```
This module is meant for internal use only and it is not meant for use in library code.
Importing from this module is deprecated and it will be removed in Enable 6.0.
We highly recommend that you update your code and vendorize what is necessary.
```

PR #806, opened 2021-04-28
use non deprecated font families styles etc

### LINKED ISSUE #803: wx.DEFAULT is deprecated in favor of wx.FONTFAMILY_DEFAULT
https://github.com/enthought/enable/blob/f292da25818cc9642b27beae282fd2ed303e6b80/kiva/trait_defs/ui/wx/kiva_font_editor.py#L56-L63
https://github.com/enthought/enable/blob/f292da25818cc9642b27beae282fd2ed303e6b80/kiva/trait_defs/ui/wx/kiva_font_editor.py#L86-L93

Ref https://wxpython.org/Phoenix/docs/html/MigrationGuide.html#font-pen-and-brush-styles

aaronayres35:
Closes #803 
This PR directly follows the migration guide linked in the issue: https://wxpython.org/Phoenix/docs/html/MigrationGuide.html#font-pen-and-brush-styles

### DIFF
--- a/kiva/trait_defs/ui/wx/kiva_font_editor.py
+++ b/kiva/trait_defs/ui/wx/kiva_font_editor.py
@@ -51,16 +51,20 @@ class ToolkitEditorFactory(EditorFactory):
         import kiva.constants as kc
 
         font = editor.value
-        weight = (wx.NORMAL, wx.BOLD)[font.weight == kc.BOLD]
-        style = (wx.NORMAL, wx.ITALIC)[font.style == kc.ITALIC]
+        weight = (
+            wx.FONTSTYLE_NORMAL, wx.FONTWEIGHT_BOLD
+        )[font.weight == kc.BOLD]
+        style = (
+            wx.FONTSTYLE_NORMAL, wx.FONTSTYLE_ITALIC
+        )[font.style == kc.ITALIC]
         family = {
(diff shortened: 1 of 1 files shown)


kiva/trait_defs/ui/wx/kiva_font_editor.py
```diff
@@ -51,16 +51,20 @@ def to_wx_font(self, editor):
         import kiva.constants as kc
 
         font = editor.value
-        weight = (wx.NORMAL, wx.BOLD)[font.weight == kc.BOLD]
-        style = (wx.NORMAL, wx.ITALIC)[font.style == kc.ITALIC]
+        weight = (
+            wx.FONTSTYLE_NORMAL, wx.FONTWEIGHT_BOLD
+        )[font.weight == kc.BOLD]
```
rahulporuri, line 56: It took me a second to understand why a `True`/`False` was being used to index into a tuple. 

I was going to ask that we rewrite this to be more readable - but given that this is the wx toolkit, let's not.

rahulporuri (approved):
:shipit:

kiva/trait_defs/ui/wx/kiva_font_editor.py
aaronayres35: Yeah... very strange

PR #808, opened 2021-04-28
Add pyproject.toml to specify cython and numpy as build deps

aaronayres35:
1st part of #769 
This PR simply includes a `pyproject.toml` file specifying `cython` and `numpy` as build dependencies.  

I ran `python setup.py sdist`, created a fresh environment and ran `pip install enable-5.1.0.dev280.tar.gz`

previously I saw:
```
$ pip install enable-5.1.0.dev280.tar.gz
Processing ./enable-5.1.0.dev280.tar.gz
    ERROR: Command errored out with exit status 1:
     command: /Users/aayres/.edm/envs/dumb/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-req-build-e1pybp8n/setup.py'"'"'; __file__='"'"'/private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-req-build-e1pybp8n/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-req-build-e1pybp8n/pip-egg-info
         cwd: /private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-req-build-e1pybp8n/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-req-build-e1pybp8n/setup.py", line 19, in <module>
        from Cython.Distutils import build_ext
    ModuleNotFoundError: No module named 'Cython'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

Now I see the far more long winded:
<details>

```
$ pip install enable-5.1.0.dev280.tar.gz
Processing ./enable-5.1.0.dev280.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Collecting pillow
[...]

PR #811, opened 2021-05-05
verify swig version in setup.py

### LINKED ISSUE #769: Make the user experience of installing enable from PyPI better
At the moment, installing `enable` from PyPI can be incredibly complicated and bug-ridden. For example, here is what happens when I create a new edm environment and do `python -m pip install enable`

1. Installation fails with the error `ModuleNotFoundError: No module named 'Cython'`
2. When I edm install `cython` and retry pip installing `enable`, installation fails with the error `ModuleNotFoundError: No module named 'numpy'`
3. When I edm install `numpy` and retry pip installing `enable`, installation proceeds slightly further but it fails again with the error `error: command 'swig.exe' failed: No such file or directory`.
4. `swig` cannot be installed from pypi so the user will need to use the operating system package manager or a third party package manager to install `swig`, assuming that the user doesn't want to use edm. Just asking the package manager for `swig` will probably install the latest version of swig i.e. version 4. Pip installing `enable` after installing `swig` will work - but it will be subtly broken because `enable` needs swig version 3, not version 4. Ref issue https://github.com/enthought/enable/issues/360#issuecomment-625464941

These issues are not documented well and put together, create a high barrier for users who try to `pip install enable`.

We could improve the user experience by doing two things - 
1. Use `pyproject.toml` to specify the build dependencies i.e. `cython` and `numpy`
2. We cannot specify `swig` as a build dependency in the toml file but we can check the version of swig being used to build `enable`/`kiva` in the `setup.py` file - and raise an exception if swig version 3 isnt being used.

aaronayres35:
This is a follow up to #808 which closes #769 

This PR checks the version of swig via a subprocess call (a little ugly but seems to get the job done) in setutp.py.
If swig is not around or we have a bad version (not version 3.x) we raise an exception.  This follows the second point at the end of the 769 issue description

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -106,7 +106,6 @@ dependencies = {
     "traitsui",
     "pyface",
     "pypdf2",
-    "swig",
     "unittest2",
 }
 
--- a/setup.py
+++ b/setup.py
@@ -429,6 +429,18 @@ def macos_extensions():
 
 
 if __name__ == "__main__":
+
+    msg = ("SWIG is a required build dependency of Enable. Furthermore, there "
+          "is currently a known issue with SWIG 4.0, see enthought/enable#360."
+          " Please download SWIG 3.0.x (see http://www.swig.org/).")
+    try:
+        with subprocess.Popen(["swig", "-version"], stdout=subprocess.PIPE, encoding='utf-8') as proc:
+            swig_version_line = proc.stdout.read().split('\n')[1]
+            swig_version = swig_version_line.split(' ')[2]
+            if  not swig_version.startswith('3'):
+                raise Exception(msg)
+    except FileNotFoundError:
+        raise Exception(msg)
     # Write version modules as needed
     enable_version_path = os.path.join('enable', '_version.py')
     write_version_py(filename=enable_version_path)


setup.py
```diff
@@ -429,6 +429,18 @@ def macos_extensions():
...
 if __name__ == "__main__":
+
+    msg = ("SWIG is a required build dependency of Enable. Furthermore, there "
+          "is currently a known issue with SWIG 4.0, see enthought/enable#360."
+          " Please download SWIG 3.0.x (see http://www.swig.org/).")
+    try:
+        with subprocess.Popen(["swig", "-version"], stdout=subprocess.PIPE, encoding='utf-8') as proc:
+            swig_version_line = proc.stdout.read().split('\n')[1]
+            swig_version = swig_version_line.split(' ')[2]
+            if  not swig_version.startswith('3'):
```
aaronayres35, line 440: I guess as #360 mentions swig 4 breaks Agg backend specifically.  Maybe this should just be a warning here instead of raising?  Saying beware agg won't work but if you wanna use non agg, go right ahead as you are?

ci/edmtool.py
```diff
@@ -106,7 +106,6 @@
     "traitsui",
     "pyface",
     "pypdf2",
-    "swig",
```
rahulporuri, line 109: i'm not sure why you removed `swig` from the dependencies list here. Without this, `swig` won't get installed via edm into the python environment.

setup.py
```diff
@@ -429,6 +429,18 @@ def macos_extensions():
 
 
 if __name__ == "__main__":
+
+    msg = ("SWIG is a required build dependency of Enable. Furthermore, there "
+          "is currently a known issue with SWIG 4.0, see enthought/enable#360."
+          " Please download SWIG 3.0.x (see http://www.swig.org/).")
```
rahulporuri, line 435:
```suggestion
          " Please install SWIG 3.0.x (see http://www.swig.org/).")
```

setup.py
```diff
@@ -429,6 +429,18 @@ def macos_extensions():
...
+    try:
+        with subprocess.Popen(["swig", "-version"], stdout=subprocess.PIPE, encoding='utf-8') as proc:
```
rahulporuri, line 437: it looks like we don't run flake8 on ci.

setup.py
rahulporuri: > Maybe this should just be a warning here instead of raising?

I think this should be an error. Warnings during installation might be ignored.

setup.py
```diff
@@ -429,6 +429,18 @@ def macos_extensions():
...
+            swig_version_line = proc.stdout.read().split('\n')[1]
+            swig_version = swig_version_line.split(' ')[2]
+            if  not swig_version.startswith('3'):
+                raise Exception(msg)
+    except FileNotFoundError:
```
rahulporuri, line 442: i'm not entirely sure what raises the `FileNotFoundError`

setup.py
```diff
@@ -429,6 +429,18 @@ def macos_extensions():
...
 
 if __name__ == "__main__":
+
+    msg = ("SWIG is a required build dependency of Enable. Furthermore, there "
+          "is currently a known issue with SWIG 4.0, see enthought/enable#360."
+          " Please download SWIG 3.0.x (see http://www.swig.org/).")
+    try:
+        with subprocess.Popen(["swig", "-version"], stdout=subprocess.PIPE, encoding='utf-8') as proc:
+            swig_version_line = proc.stdout.read().split('\n')[1]
+            swig_version = swig_version_line.split(' ')[2]
```
rahulporuri, line 439:
```suggestion
            # We expect text along the lines of "SWIG Version 3.0.11"
            swig_version = swig_version_line.split(' ')[2]
```

ci/edmtool.py
aaronayres35: it was listed twice, see line 104, I should have mentioned that

setup.py
```diff
@@ -429,6 +429,19 @@ def macos_extensions():
 
 
 if __name__ == "__main__":
+
+    msg = ("SWIG is a required build dependency of Enable. Furthermore, there "
+          "is currently a known issue with SWIG 4.0, see enthought/enable#360."
```
jwiggins, line 434: This is complex enough to belong in its own function

setup.py
```diff
@@ -429,6 +429,19 @@ def macos_extensions():
...
+          " Please install SWIG 3.0.x (see http://www.swig.org/).")
+    try:
+        with subprocess.Popen(["swig", "-version"], stdout=subprocess.PIPE, encoding='utf-8') as proc:
+            swig_version_line = proc.stdout.read().split('\n')[1]
+            # We expect text along the lines of "SWIG Version 3.0.11"
+            swig_version = swig_version_line.split(' ')[2]
```
jwiggins, line 440: Did you consider using a regex here?

setup.py
aaronayres35: if `swig` is not installed, then `swig -version` causes this:
```
$ brew uninstall swig
Uninstalling /usr/local/Cellar/swig/4.0.2... (725 files, 5.4MB)

(dumb)~/Desktop/enable/dist aayres  (verify-swig-version) $ swig -version
bash: /usr/local/bin/swig: No such file or directory

(dumb)~/Desktop/enable/dist aayres  (verify-swig-version) $ pip install enable-5.1.0.dev287.tar.gz
Processing ./enable-5.1.0.dev287.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /Users/aayres/.edm/envs/dumb/bin/python3 /Users/aayres/.edm/envs/dumb/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/tmp5st4ag4b
       cwd: /private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-req-build-w6n7h3n2
  Complete output (29 lines):
    File "setup.py", line 439, in verify_swig_version
      stdout=subprocess.PIPE, encoding='utf-8') as proc:
    File "/Users/aayres/.edm/envs/dumb/lib/python3.6/subprocess.py", line 729, in __init__
      restore_signals, start_new_session)
    File "/Users/aayres/.edm/envs/dumb/lib/python3.6/subprocess.py", line 1364, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  FileNotFoundError: [Errno 2] No such file or directory: 'swig': 'swig'
  
    File "/Users/aayres/.edm/envs/dumb/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 257, in <module>
      main()
    File "/Users/aayres/.edm/envs/dumb/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 240, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/Users/aayres/.edm/envs/dumb/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 91, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-build-env-017ig1uv/overlay/lib/python3.6/site-packages/setuptools/build_meta.py", line 155, in get_requires_for_build_wheel
      config_settings, requirements=['wheel'])
    File "/private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-build-env-017ig1uv/overlay/lib/python3.6/site-packages/setuptools/build_meta.py", line 135, in _get_build_requires
      self.run_setup()
    File "/private/var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/pip-build-env-017ig1uv/overlay/lib/python3.6/site-packages/setuptools/build_meta.py", line 150, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 449, in <module>
      verify_swig_version()
    File "setup.py", line 446, in verify_swig_version
      raise Exception(msg)
  Exception: SWIG is a required build dependency of Enable. Furthermore, there is currently a known issue with SWIG 4.0, see enthought/enable#360. Please install SWIG 3.0.x (see http://www.swig.org/).
  ----------------------------------------
ERROR: Command errored out with exit status 1: /Users/aayres/.edm/envs/dumb/bin/python3 /Users/aayres/.edm/envs/dumb/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/j9/ps8s26f91_1_lgf7mtd4smyr0000gn/T/tmp5st4ag4b Check the logs for full command output.
```

ci/edmtool.py
```diff
@@ -106,7 +106,6 @@
...
     "unittest2",
```
rahulporuri, line 109: we don't depend on `unittest2` anymore. we should be able to remove this - maybe in a later PR

ci/edmtool.py
rahulporuri: ahh. i should have noticed that things were not in alphabetical order - which i what i was kinda expecting.

can you rearrange the set alphabetically, for readability? this could be done in a later PR

rahulporuri (approved):
LGTM.

setup.py
rahulporuri: >  The current code is a little bit hacky looking, although I'm not sure a regex would look much cleaner 🤔

I think it would look cleaner. Instead of splitting lines and then checking that `3` exists in the specific line, you can just use the regex on the stdout directly.

jwiggins:
Do we not run flake8 as part of CI in this project?

setup.py
```diff
@@ -428,7 +428,29 @@ def macos_extensions():
...
+    """
+    msg = ("SWIG is a required build dependency of Enable. Furthermore, there "
+          "is currently a known issue with SWIG 4.0, see enthought/enable#360."
+          " Please install SWIG 3.0.x (see http://www.swig.org/).")
+    try:
+        with subprocess.Popen(["swig", "-version"], \
+                stdout=subprocess.PIPE, encoding='utf-8') as proc:
+            # We expect text along the lines of "SWIG Version 3.X.Y"
+            swig_version_match = re.search(
+                '(SWIG Version 3)\.\d{1,2}\.\d{1,2}',
```
jwiggins, line 442: Shouldn't this be `r"\(SWIG Version 3\)\.\d{1,2}\.\d{1,2}"`? Did you test this?

aaronayres35: I tested and everything seemed to work as I had expected (I got an error with swig 4, rolled back to swig 3 and install succeeded)
why the `\(`, and `\)` as we don't want literal "(" or ")"?
The leading r is probably safer, I think I may have just lucked out here that it works without it 
I will update

PR #813, opened 2021-05-12
Sierpinski's Triangle example

aaronayres35:
This PR introduces a new example for visualizing Sierpinski's Triangle.

We currently have a kiva tutorial and maybe eventually it could be useful to turn this into a comparable enable tutorial?  It will need to be cleaned up first. 

~Some things of note: Sierpinski's Triangle can be coded in a cleaner way (and this can be down with enabl/kiva) where you have a draw triangle method and a midpoint method where basically you draw a triangle and then another one using the midpoints of the original as vertices for the next, and then use the midpoints of each of the 3 resulting triangles to draw 3 more and so on.~

~I was trying to make use of `draw_marker_at_points` here (initially) and so was explicitly calculating the marker locations relative to a previous marker location.  I ultimately could not use `draw_marker_at_points` because I realized the kiva `INVERTED_TRIANGLE_MARKER` was not drawing how I expected based on the `_add_to_path` method for the corresponding enable marker.  It did not appear to be drawing equilateral triangles, and they were shifted slightly.  Because it was more obvious to see exactly how the path was being drawn via the `_add_to_path` method I simply used `draw_path_at_points` (it is more generally available IIRC across backends anyway).~

~As a part of this motivation to use  `draw_path_at_points` in my recursive method to find where to draw triangles, rather than drawing them in that method (which is what I had originally done), I instead stored the locations / sizes in a `point_contexts` dictionary (naming things is hard and the current naming in this PR is very ugly).  The number of triangles to draw increases by a power of 3 every iteration, so this way we can make `self.iterations` calls to  `draw_path_at_points` rather than once call for every 3 drawn triangles.~

~There are still improvements I want to make, the biggest one being adding zoom which I am not exactly sure how to do currently...~


jwiggins (commented):
Thanks for the example. Sierpinski's triangle is one of my favorite fractals.

I feel like this could be done without a marker object, but I'm still sore about #390 and #716. I think you've basically got the math already for drawing a triangle of arbitrary scale given a center point. You can still minimize draw calls by recursively building a path object and then drawing it in one call.

(don't forget to flake this before merging)

aaronayres35:
@jwiggins thank you for the review!

enable/examples/demo/enable/sierpinski.py
```diff
@@ -0,0 +1,146 @@
...
+    bounds_trait, Component, ComponentEditor, Inverted_TriangleMarker
+)
+
+
+m = Inverted_TriangleMarker()
+
+
+class SierpinskiTriangle(Component):
+
+    base_width = Int(500)
```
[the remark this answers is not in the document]
aaronayres35: I had originally been thinking to do that but I might hold off and do it as a later improvement in a future PR

enable/examples/demo/enable/sierpinski_zoom.py
```diff
@@ -0,0 +1,207 @@
...
+        self.viewport = Viewport(
+            component=canvas, enable_zoom=True, stay_inside=True
+        )
+        self.viewport.view_position = [0, 0]
+        self.viewport.view_bounds = [
+            self.base_width, self.base_width*(SQRT3/2)
+        ]
+        self.viewport.tools.append(
+            ViewportPanTool(self.viewport, drag_button="right")
+        )
```
aaronayres35: `Viewport` is very strange... It has a `zoom_tool` trait so zoom is handled differently from other tools, eg `ViewportPanTool`.  Additionally, with this line:
https://github.com/enthought/enable/blob/7f3231d14d118b7a7966af0cd4f0c257340569b9/enable/viewport.py#L97-L98
we create a `ViewportZoomTool` instance by default if `enable_zoom` is True, passing in `self` (the viewport) as the component.  That is all fine and good, but I was struggling to see how I could pass in a zoom_tool other than this... If I want to pass in my own zoom tool when instantiation a viewport, eg:
```
my_zoom_tool = ViewportZoomTool(...)
my_viewport = Viewport(..., zoom_tool=my_zoom_tool)
```
I was running into a chicken and egg scenario as I needed the `my_viewport` instance to pass in as the component for the `ViewportZoomTool`, but it isn't defined when I define `my_zoom_tool`.  If I try to switch the order:
```
my_viewport = Viewport(..., zoom_tool=my_zoom_tool)
my_zoom_tool = ViewportZoomTool(...)
```
Then the above line gets run so I end up using the default zoom tool, not `my_zoom_tool`.  Thus I resorted to just setting the trait I wanted to change on the default zoom tool that got created, rather than trying to instantiate my own.
This all works as I want it to.  However, I do not understand when a user would want / be able to set the `zoom_tool` trait.  Maybe if the `ViewportZoomTool` was working with a second `Viewport` or something?
However, as far as I understand these are always with a `Canvas` (that very well may not actually be true, but there isn't documentation on this / there aren't many examples) , and `Canvas` has the comment:
https://github.com/enthought/enable/blob/7f3231d14d118b7a7966af0cd4f0c257340569b9/enable/canvas.py#L34-L35
so that doesn't seem possible either.


Additionally, `ViewportZoomTool` has traits `max_zoom` / `min_zoom`, but its base class `BaseZoomTool` has traits `max_zoom_in_factor` and `min_zoom_in_factor` which appear to have the same purpose.  

There are still many things about `Viewport` and `ViewportZoomTool` I do not fully understand / I was unable to get to work how I expected.  e.g. the `{horizontal/vertical}_anchor` traits, `stay_inside` / `view_bounds`.


In any case, this example effectively works as I want it to.  Perhaps I should open some issues about these things / an issue about adding documentation on how to use zoom in enable generally.  I imagine enable zoom is a pretty niche thing as chaco handles zoom on its own, but even so.

enable/examples/demo/enable/sierpinski_zoom.py
```diff
@@ -0,0 +1,204 @@
...
+        canvas.overlays.append(self.triangle)
+
+        self.viewport = Viewport(
+            component=canvas, enable_zoom=True, stay_inside=True
+        )
+        self.viewport.zoom_tool.min_zoom = 1.0
+        self.viewport.tools.append(
+            ViewportPanTool(self.viewport, drag_button="right")
+        )
```
aaronayres35, line 164: `Viewport` is very strange... It has a `zoom_tool` trait so zoom is handled differently from other tools, eg `ViewportPanTool`.  Additionally, with this line:
https://github.com/enthought/enable/blob/7f3231d14d118b7a7966af0cd4f0c257340569b9/enable/viewport.py#L97-L98
we create a `ViewportZoomTool` instance by default if `enable_zoom` is True, passing in `self` (the viewport) as the component.  That is all fine and good, but I was struggling to see how I could pass in a zoom_tool other than this... If I want to pass in my own zoom tool when instantiation a viewport, eg:
```
my_zoom_tool = ViewportZoomTool(...)
my_viewport = Viewport(..., zoom_tool=my_zoom_tool)
```
I was running into a chicken and egg scenario as I needed the `my_viewport` instance to pass in as the component for the `ViewportZoomTool`, but it isn't defined when I define `my_zoom_tool`.  If I try to switch the order:
```
my_viewport = Viewport(..., zoom_tool=my_zoom_tool)
my_zoom_tool = ViewportZoomTool(...)
```
Then the above line gets run so I end up using the default zoom tool, not `my_zoom_tool`.  Thus I resorted to just setting the trait I wanted to change on the default zoom tool that got created, rather than trying to instantiate my own.
This all works as I want it to.  However, I do not understand when a user would want / be able to set the `zoom_tool` trait.  Maybe if the `ViewportZoomTool` was working with a second `Viewport` or something?
However, as far as I understand these are always with a `Canvas` (that very well may not actually be true, but there isn't documentation on this / there aren't many examples) , and `Canvas` has the comment:
https://github.com/enthought/enable/blob/7f3231d14d118b7a7966af0cd4f0c257340569b9/enable/canvas.py#L34-L35
so that doesn't seem possible either.


Additionally, `ViewportZoomTool` has traits `max_zoom` / `min_zoom`, but its base class `BaseZoomTool` has traits `max_zoom_in_factor` and `min_zoom_in_factor` which appear to have the same purpose.  

There are still many things about `Viewport` and `ViewportZoomTool` I do not fully understand / I was unable to get to work how I expected.  e.g. the `{horizontal/vertical}_anchor` traits, `stay_inside` / `view_bounds`.


In any case, this example effectively works as I want it to.  Perhaps I should open some issues about these things / an issue about adding documentation on how to use zoom in enable generally.  I imagine enable zoom is a pretty niche thing as chaco handles zoom on its own, but even so.

aaronayres35:
I've removed the WIP label from this PR.  I think there are certainly still improvements that can be made, but I think I will leave them for follow up PRs now.

enable/examples/demo/enable/sierpinski_zoom.py
```diff
@@ -0,0 +1,203 @@
...
+            iterations=self.iterations,
+            max_iters=self.max_iters,
+            base_width=self.base_width,
+        )
+        canvas.overlays.append(self.triangle)
+
+        self.viewport = Viewport(
+            component=canvas, enable_zoom=True, stay_inside=True
+        )
+        self.viewport.zoom_tool.min_zoom = 1.0
```
aaronayres35, line 161: this works as I want it to initially (eg when base_width is 500), in that you can not zoom out passed when you see the full triangle.  However, if base_width is changed to 1000 for example now you can not zoom out enough.  

I want to only be able to zoom out as far as bounds.  I have had trouble understanding / getting things to work with the `stay_inside` trait, `view_bounds`, etc.

aaronayres35:
@jwiggins no rush at all, but if you do have a chance to take another look at this PR I would greatly appreciate your feedback!

enable/examples/demo/enable/sierpinski_zoom.py
jwiggins: Not sure how much I can help. I have basically no experience with `Viewport`.

enable/examples/demo/enable/sierpinski_zoom.py
```diff
@@ -0,0 +1,203 @@
...
+        canvas = Canvas(draw_axes=True, bgcolor="gray")
+        self.triangle = SierpinskiTriangle(
+            canvas,
+            position=[0.0, 0.0],
+            bounds=[self.base_width, self.base_width*(SQRT3/2)],
+            iterations=self.iterations,
+            max_iters=self.max_iters,
+            base_width=self.base_width,
+        )
+        canvas.overlays.append(self.triangle)
```
jwiggins, line 156: Is this a `Canvas` thing to use `overlays`? Is there no "main layer"?

enable/examples/demo/enable/sierpinski_zoom.py
```diff
@@ -0,0 +1,203 @@
...
+        gc.begin_path()
+        path = gc.get_empty_path()
+
+        if self.iterations >= 1:
+            with gc:
+                self.add_triangle_to_path(
+                    path,
+                    (self.base_width/4, (self.base_width/4)*SQRT3),
+                    self.base_width/2
+                )
```
jwiggins, line 59: Why use `with gc` at this level? You don't appear to be modifying the state.

enable/examples/demo/enable/sierpinski_zoom.py
```diff
@@ -0,0 +1,203 @@
...
+        """ Recursive method to determine size and location of triangles to
+        add to input path object, based on the current iteration n.
+
+        This method recursively adds to the path object which can be drawn upon
+        completion of the full recursive call.
+        """
+        size = self.base_width/4 * (1/2)**(n - 1)
+        if n == self.iterations:
+            return
+        else:
```
jwiggins, line 79: The `else` isn't strictly necessary here.

enable/examples/demo/enable/sierpinski_zoom.py
jwiggins: Yeah.. I'm not entirely sure what use case this stuff was initially built for but it likely doesn't exist anymore. The quirks are the interface at this point, unfortunately.

enable/examples/demo/enable/sierpinski_zoom.py
```diff
@@ -0,0 +1,203 @@
...
+            viewport_component=self.viewport,
+            always_show_sb=True,
+            continuous_drag_update=True,
+        )
+        return scrolled
+
+    @observe("base_width")
+    def _update_base_width(self, event):
+        self.triangle.bounds = [self.base_width, self.base_width*(SQRT3/2)]
+        self.triangle.base_width = event.new
```
jwiggins, line 179: It reads a little weird to mix `self.base_width` and `event.new` in these two lines

jwiggins (approved):
This is good enough as is to merge now. If you're looking for follow-up work to do, I would consider caching the path which is generated so that you don't have to do the expensive recursive call for every draw. Performance gets really sluggish if you zoom in and crank the iterations up.

enable/examples/demo/enable/sierpinski_zoom.py
aaronayres35: Ah you are right, this just didn't get thrown out when I updated the code.  Previously it had modified state.

aaronayres35:
@jwiggins thank you for the reviews!

enable/examples/demo/enable/sierpinski_zoom.py
aaronayres35: No worries.
I am curious if this (viewport, canvas, etc.) gets much (_any_) use out the wild

enable/examples/demo/enable/sierpinski_zoom.py
aaronayres35: Ah good point this is not necessary. I had struggled to get things working and just followed one of the examples.  I believe it is working now with `SierpinskiTriangle` subclassing `Component` not `AbstractOverlay`, and just calling `canvas.add` here

Issue #815, opened 2021-05-12
ColorPicker is used but no longer exists in code base


aaronayres35:
searching the code base for `ColorPicker` the only results are https://github.com/enthought/enable/blob/21c299e1229d74e9db701487a613934e332a7225/enable/trait_defs/ui/wx/enable_rgba_color_editor.py#L20 and https://github.com/enthought/enable/blob/21c299e1229d74e9db701487a613934e332a7225/enable/trait_defs/ui/wx/enable_rgba_color_editor.py#L139

but I dont see the class defined anywhere

rahulporuri:
I can't find any users of `enable.ColorPicker` on GitHub - https://github.com/search?q=%22from+enable+import+ColorPicker%22&type=code and https://github.com/search?q=%22enable.ColorPicker%22+extension%3Apy&type=code

I couldn't find any users of `enable_rgba_color_editor` either - https://github.com/search?p=1&q=%22enable_rgba_color_editor%22+extension%3Apy&type=Code - so i think we can nuke that module.

From the looks of it, this actually got moved into traitsui as the color editor module.

PR #816, opened 2021-05-12
Remove AbstractWindow.bg_color trait

### LINKED ISSUE #812: Remove deprecated "AbstractWindow.bg_color" trait
https://github.com/enthought/enable/blob/5d130a1462c6082b1485d0a76fd45ce1d8abd7bb/enable/abstract_window.py#L71-L74

This backwards compatibility seems to have been added in the commit - https://github.com/enthought/enable/commit/1f5e90f241ea61b1cedf2026a022cca509df9301 - which has existed in chaco since version 3.3.0 of chaco.

aaronayres35:
closes #812

This PR simply removes the old deprecated trait.  Note searching the codebase for `bg_color` is what motivated issue #815. Perhaps `enable/trait_defs/ui/wx/enable_rgba_color_editor.py` should just be deleted?  given 815 I imagine it is completely broken

### DIFF
--- a/enable/abstract_window.py
+++ b/enable/abstract_window.py
@@ -68,11 +68,6 @@ class AbstractWindow(HasTraits):
     # painted with this color before the component gets to draw.
     bgcolor = ColorTrait("sys_window")
 
-    # Unfortunately, for a while, there was a naming inconsistency and the
-    # background color trait named "bg_color".  This is still provided for
-    # backwards compatibility but should not be used in new code.
-    bg_color = Alias("bgcolor")
-
     alt_pressed = Bool(False)
     ctrl_pressed = Bool(False)
     shift_pressed = Bool(False)


rahulporuri (approved):
LGTM. I think we can remove this outright without worrying about breaking any code. Can you also search chaco for uses of `bg_color`, just to be safe?

Also, I think we can remove the `enable_rgba_color_editor.py` module but let's do that in a separate PR

aaronayres35:
> Also, I think we can remove the `enable_rgba_color_editor.py` module but let's do that in a separate PR

👍 I've opened #817

aaronayres35:
> LGTM. I think we can remove this outright without worrying about breaking any code. Can you also search chaco for uses of `bg_color`, just to be safe?

After searching chaco for `bg_color` there are a few uses, but the majority of which `bg_color` is used as a temp variable that is passed to `bgcolor`.  eg `bgcolor=bg_color`.
There are 2 problem uses tho:
https://github.com/enthought/chaco/blob/54a1b929d3e3932bc79e17de5995658bd30e1b37/examples/demo/basic/contour_cmap_plot.py#L65

https://github.com/enthought/chaco/blob/54a1b929d3e3932bc79e17de5995658bd30e1b37/examples/demo/basic/contour_plot.py#L60
All uses occur in example code, not in library code.

PR #820, opened 2021-05-20
Fix artefact in Qt caused by QRectF having a slightly wrong size

### LINKED ISSUE #819: Missing pixels with Qt and image backend
When drawing with Enable + Qt, some pixels seem to have gone missing, causing broken lines in an image.
e.g.
Screenshot 2021-05-20 at 17 40 58

One can see, with eagle eyes, that the lines are broken along the middle row and column.

This issue is more obvious in a non-HiDPI situation than in a HiDPI situation. The above image is from a non-HiDPI setting. The following image is from a HiDPI setting:
Screenshot 2021-05-20 at 17 42 54

To reproduce:
```
from traits.api import HasTraits, Instance
from traitsui.api import Item, View
from enable.api import Label, Component, ComponentEditor


TEXT = "\n".join(["HELLO" * 1000] * 1000)


class DemoFrame(HasTraits):

    component = Instance(Component)

    traits_view = View(
        Item("component", editor=ComponentEditor(), show_label=False),
        resizable=True,
    )

    def _component_default(self):
        return self._create_component()

    def _create_component(self):
        return Label(text=TEXT, rotate_angle=45.0, position=(200, 0))


if __name__ == "__main__":
    DemoFrame().configure_traits()

```

Enable version: 5.1.0
Platform: OSX (with retina display and without, see description above)
Qt backend: PySide2

Note: A lot of existing examples in Enable uses rectangles, so such misalignments would not be obvious visually (a rectangle may simply be a pixel shorter).

kitchoi:
Closes #819

The issue is tested (manually) to affect these backends: `qt4.image`, `qt4.celiagg` and `qt4.cairo`.
For these backends, the `GraphicsContext` is defined to be slightly larger than the requested size:
https://github.com/enthought/enable/blob/1c3805cdf8e47085a2ae91cdcdc2ed0d12ba33e6/enable/qt4/cairo.py#L20-L23

There is another `QRectF` defined for `qt4.quartz`, but I could not reproduce the issue there. For this backend, the GraphicsContext was defined without the plus-one:
https://github.com/enthought/enable/blob/685038503500ecf8ffc30f3da68c7e848ae02e6f/enable/qt4/quartz.py#L35

Note that `base_scale` was not defined on the `GraphicsContext` for the cairo backend, and so it has to be added.

Apparently `base_scale` is not defined on the `qpainter`'s `GraphicsContext` either, but that would be an orthogonal issue as #819 does not affect the `qpainter` backend.

### DIFF
--- a/enable/qt4/cairo.py
+++ b/enable/qt4/cairo.py
@@ -36,6 +36,8 @@ class Window(BaseWindow):
 
         image = QtGui.QImage(data, w, h, QtGui.QImage.Format_ARGB32)
 
-        rect = QtCore.QRectF(0, 0, self.control.width(), self.control.height())
+        rect = QtCore.QRectF(
+            0, 0, w / self._gc.base_scale, h / self._gc.base_scale
+        )
         painter = QtGui.QPainter(self.control)
         painter.drawImage(rect, image)
--- a/enable/qt4/celiagg.py
+++ b/enable/qt4/celiagg.py
@@ -48,7 +48,9 @@ class Window(BaseWindow):
         image = QtGui.QImage(
             self._shuffle_buffer, w, h, QtGui.QImage.Format_RGB32
         )
-        rect = QtCore.QRectF(0, 0, self.control.width(), self.control.height())
+        rect = QtCore.QRectF(
+            0, 0, w / self._gc.base_scale, h / self._gc.base_scale
+        )
         painter = QtGui.QPainter(self.control)
         painter.drawImage(rect, image)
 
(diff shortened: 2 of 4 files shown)


rahulporuri:
@jwiggins any chance you can make time to review this PR today/monday? It looked good to me but i wanted a more experienced pair of eyes. The idea is to make a bug fix release ASAP.

jwiggins (approved):
LGTM

rahulporuri:
@kitchoi i might be doing something wrong but have you/someone else on your team tested this fix on windows?

rahulporuri (approved):
LGTM

kitchoi:
Thanks for the review and testing!

PR #833, opened 2021-06-01
Explicitly require pyface>=7.2 since we use pyface.undo

aaronayres35:
This has actually been true since PR #507 

Adding it now simply to be more explicit


rahulporuri (approved):
LGTM

PR #837, opened 2021-06-01
Enable Release Candidate 5.2.0rc2

aaronayres35:
The merge commit from this PR will be tagged as the 5.2.0 release candidate (this time for real)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ from setuptools.command.install import install as base_install
 MAJOR = 5
 MINOR = 2
 MICRO = 0
-PRERELEASE = "rc1"
+PRERELEASE = "rc2"
 IS_RELEASED = True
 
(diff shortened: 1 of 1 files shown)


rahulporuri (commented):
This doesn't look right. We already have a commit in the `maint/5.2` branch which is an rc1 release. This needs to be an rc2 release because we shouldn't have two commits which mark the same version.

aaronayres35:
> This doesn't look right. We already have a commit in the `maint/5.2` branch which is an rc1 release. This needs to be an rc2 release because we shouldn't have two commits which mark the same version.

Ah okay I see, I wasn't thinking at a commit level, I was thinking since the tag for the other was removed we could simply tag this instead. I will update this to be `"rc2"`.

rahulporuri (approved):
LGTM

PR #839, opened 2021-06-05
Make Kiva compileable with Swig 4

### LINKED ISSUE #360: SWIG 4.0 appears to break the kiva.agg backend

jvkersch:
Fixes #360.

The Swig wrappers used a historic construct (using `_swig_setattr` instead of direct attribute assignment) that broke with the relase of Swig 4 (which removed `_swig_setattr`). 

There are a lot of historical oddities with the Swig wrappers (as is well-known); I didn't touch anything beyond the strictly necessary, expecting the Swig layer to go away with #414.

### DIFF
--- a/kiva/agg/src/font_type.i
+++ b/kiva/agg/src/font_type.i
@@ -84,7 +84,7 @@ def unicode_safe_init(self, _name="Arial", _size=12, _family=0, _style=0,
                                _encoding, _face_index, validate)
 
     self.this = obj
-    self.thisown = 1                
+    self.thisown = 1
 
 # This is a crappy way of overriding the constructor
 AggFontType.__init__ = unicode_safe_init
--- a/kiva/agg/src/font_type.i
+++ b/kiva/agg/src/font_type.i
@@ -82,8 +82,9 @@ def unicode_safe_init(self, _name="Arial", _size=12, _family=0, _style=0,
             _name = _name.decode()
     obj = _agg.new_AggFontType(_name, _size, _family, _style,
                                _encoding, _face_index, validate)
-    _swig_setattr(self, AggFontType, "this", obj)
-    _swig_setattr(self, AggFontType, "thisown", 1)
+
+    self.this = obj
+    self.thisown = 1                
 
 # This is a crappy way of overriding the constructor
 AggFontType.__init__ = unicode_safe_init
--- a/kiva/agg/src/graphics_context.i
+++ b/kiva/agg/src/graphics_context.i
@@ -362,12 +362,8 @@ namespace kiva {
                 # Apply base scale for a HiDPI context
                 _agg.GraphicsContextArray_scale_ctm(obj, base_pixel_scale, base_pixel_scale)
 
-                _swig_setattr(self, GraphicsContextArray, 'this', obj)
-                # swig 1.3.28 does not have real thisown, thisown is mapped
-                # to this.own() but with previous 'self.this=obj' an
-                # attribute 'own' error is raised. Does this workaround
-                # work with pre-1.3.28 swig?
-                _swig_setattr(self, GraphicsContextArray, 'thisown2', 1)
+                self.this = obj
+                self.thisown2 = 1
 
                 self.bmp_array = ary
                 self.base_scale = base_pixel_scale
--- a/kiva/agg/src/graphics_context.i
+++ b/kiva/agg/src/graphics_context.i
@@ -363,6 +363,10 @@ namespace kiva {
                 _agg.GraphicsContextArray_scale_ctm(obj, base_pixel_scale, base_pixel_scale)
 
                 self.this = obj
+                # swig 1.3.28 does not have real thisown, thisown is mapped
+                # to this.own() but with previous 'self.this=obj' an
+                # attribute 'own' error is raised. Does this workaround
+                # work with pre-1.3.28 swig?				
                 self.thisown2 = 1
 
                 self.bmp_array = ary
--- a/kiva/agg/tests/test_affine_matrix.py
+++ b/kiva/agg/tests/test_affine_matrix.py
@@ -31,7 +31,7 @@ class AffineMatrixTestCase(unittest.TestCase):
         try:
             agg.AffineMatrix(a)
         except (NotImplementedError, TypeError):
-            # NotImplementedError in Swig 3, TypeError in Swig 4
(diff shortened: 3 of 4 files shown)


aaronayres35 (approved):
LGTM (although it may be beneficial to have a pair of more SWIG-savvy eyes on this), thank you for this PR @jvkersch !

In an environment with swig 4 installed, if I comment out this line:
https://github.com/enthought/enable/blob/6e1c93e9df2a6e37eb79b92a81696ac757392dce/setup.py#L454
and I run `pip install -e .` on master things seems to work, however when running the test suite many tests fail with the `NameError: name '_swig_setattr' is not defined`.

Doing the same on this branch all the tests pass! 🎉 🙌  (aside from a couple failing with `ModuleNotFoundError: No module named 'celiagg'`, but that is orthogonal - they should be skipped if missing celiagg).

I did some digging in the swig release notes / changelog but struggled to find specific mentions about this.  These PRs seem _potentially_ relevant:
~https://github.com/swig/swig/pull/718~
https://github.com/swig/swig/pull/1261
https://github.com/swig/swig/commit/4301fa548d1e4b9a53811d766e86301b9d473d9d

TBH I do not have a great understanding on why `_swig_setattr` used to be needed.


Once this is merged, a follow up PR can be added to remove the recently added `verify_swig_version` in `setup.py`.  

~Additionally, given we have dropped support for python 2, I believe we should be passing the `-py3` flag as a swig option in setup.py? see https://github.com/swig/swig/issues/695~

kiva/agg/tests/test_affine_matrix.py
```diff
@@ -30,7 +30,8 @@ def test_init_from_array1(self):
         a = ones(6, "D")
         try:
             agg.AffineMatrix(a)
-        except NotImplementedError:
+        except (NotImplementedError, TypeError):
+            # NotImplementedError in Swig 3, TypeError in Swig 4
```
aaronayres35, line 34: I _think_ this is relevant to this?: https://github.com/swig/swig/pull/1261
changelog entry for the above PR:
"2018-07-31: wsfulton
            [Python] #1293 Overloaded C++ function wrappers now raise a TypeError instead
            of NotImplementedError when the types passed are incorrect. This change means
            there is now consistency with non-overloaded function wrappers which have always
            raised TypeError when the incorrect types are passed. The error message remains
            the same and is for example now:

              TypeError: Wrong number or type of arguments for overloaded function 'f'.
                Possible C/C++ prototypes are:
                  f(int)
                  f(char const *)

            instead of:

              NotImplementedError: Wrong number or type of arguments for overloaded function 'f'.
                Possible C/C++ prototypes are:
                  f(int)
                  f(char const *)

            *** POTENTIAL INCOMPATIBILITY ***"

kiva/agg/src/font_type.i
```diff
@@ -82,8 +82,9 @@ def unicode_safe_init(self, _name="Arial", _size=12, _family=0, _style=0,
             _name = _name.decode()
     obj = _agg.new_AggFontType(_name, _size, _family, _style,
                                _encoding, _face_index, validate)
-    _swig_setattr(self, AggFontType, "this", obj)
-    _swig_setattr(self, AggFontType, "thisown", 1)
+
+    self.this = obj
+    self.thisown = 1
```
rahulporuri, line 87:
```suggestion
    self.thisown = 1
```

kiva/agg/src/graphics_context.i
```diff
@@ -362,12 +362,8 @@ namespace kiva {
                 # Apply base scale for a HiDPI context
                 _agg.GraphicsContextArray_scale_ctm(obj, base_pixel_scale, base_pixel_scale)
 
-                _swig_setattr(self, GraphicsContextArray, 'this', obj)
-                # swig 1.3.28 does not have real thisown, thisown is mapped
-                # to this.own() but with previous 'self.this=obj' an
-                # attribute 'own' error is raised. Does this workaround
-                # work with pre-1.3.28 swig?
-                _swig_setattr(self, GraphicsContextArray, 'thisown2', 1)
```
jwiggins, line 370: Getting rid of this comment makes it a bit confusing as to why this class is using `thisown2` instead of just `thisown`

jwiggins (commented):
I'm basically 👍 on this. Glad the change wasn't complicated.

kiva/agg/src/graphics_context.i
jwiggins: Unless you're planning to fix the other `thisown2` usages, you shouldn't remove this comment.

jwiggins (approved):
Thanks for the update.

jvkersch:
@jwiggins @aaronayres35 @rahulporuri Thanks for the reviews. I restored the comment that was erroneously removed -- I don't plan on dig into the Swig code any more.

The next thing to do would be to remove the check for the version of Swig (to be clear, I think we should still check for Swig). I can do that in this PR or in a follow-up PR.

jwiggins:
I'd prefer a follow-up PR for removing the version check

rahulporuri (approved):
LGTM with one small question/comment.

rahulporuri:
> I'd prefer a follow-up PR for removing the version check

Same here. Let's get this merged so that I can revert the changes to the branch protections 😅 😬

kiva/agg/tests/test_affine_matrix.py
```diff
@@ -30,7 +30,8 @@ def test_init_from_array1(self):
         a = ones(6, "D")
         try:
             agg.AffineMatrix(a)
-        except NotImplementedError:
+        except (NotImplementedError, TypeError):
+            # NotImplementedError in Swig 3, TypeError in Swig 4 (see swig #1261)
```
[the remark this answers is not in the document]
jvkersch: @rahulporuri Did you want me to make a change here, or is your comment to provide additional clarification?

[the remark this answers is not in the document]
rahulporuri: > ... is your comment to provide additional clarification?

just additional clarification, for the human-lizard hybrid developers who will revisit this PR in the future for debugging purposes.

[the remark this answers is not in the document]
jwiggins: > for the human-lizard hybrid developers

Waitaminnit... _what do you know?_

PR #840, opened 2021-06-07
[Do not merge] Re-export layout helpers via enable.api

rahulporuri:
This PR re-exports the layout helpers `simple_container_do_layout`, `simple_container_get_preferred_size`, `stack_layout` and `stacked_preferred_size` via `enable.api` because they are now used by chaco. See PR https://github.com/enthought/chaco/pull/757. This PR also slightly improves the docstring of `stacked_preferred_size`.

### DIFF
--- a/enable/api.py
+++ b/enable/api.py
@@ -145,6 +145,12 @@ Enable Components
 - :class:`~.GraphicsContextEnable`
 - :class:`~.ImageGraphicsContextEnable`
 
+Layout Helpers
+--------------
+
+- :func:`~.stack_layout`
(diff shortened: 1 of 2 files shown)


rahulporuri:
> If these are not going to be needed in chaco when the refactor is complete, do they need to be in the api?

Yeah. Sorry. I wanted to turn this into a draft PR but I couldn't figure out how to. I'll just close the PR for now.
When the refactor is complete, this PR should be unnecessary and these helpers won't need to be re-exported via `enable.api`.

PR #787, opened 2021-04-20
Remove deprecated "str_to_font" and "points_in_polygon" functions.

rahulporuri:
This PR removes the deprecated `str_to_font` from `enable.api` and `points_in_polygon` from `kiva.agg`.

PR #843, opened 2021-06-09
add points_in_polygon back to kiva.agg

### LINKED ISSUE #842: Add points_in_polygon back to kiva.agg
Downstream users still import `points_in_polygon` from `kiva.agg`.  We should add it back for the 5.2 release to avoid breaking code

aaronayres35:
[targeting `maint/5.2`]

closes #842 

This was originally removed in #787, however we still have users of this code so we are re-adding it here to avoid breaks with the upcoming release.


rahulporuri:
here's what comes up when i search for `from kiva.agg import` in the enthought org on GitHub - of which only two instances of `from kiva.agg import ..., points_in_polygon, ...`  are relevant to us.

PR #847, opened 2021-06-11
use oldest-supported-numpy in pyproject.toml

### LINKED ISSUE #846: Built Enable/Kiva may be binary-incompatible with older NumPy versions
This is the sister issue to https://github.com/enthought/chaco/issues/778

See the above issue for the full saga.  In summary, enable will be built using the latest numpy, but an older version of numpy may be what exists in the environment after it is built.  This can lead to unexpected/confusing segfaults and compatibility issues. 

See comment: https://github.com/enthought/chaco/issues/778#issuecomment-859460130 for a promising approach to a resolution.

aaronayres35:
fixes #846 using the same approach as https://github.com/enthought/chaco/pull/779

We will want to backport this to `maint/5.2` prior to the 5.2.0 release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["cython", "numpy", "setuptools", "wheel"]
+requires = ["cython", "oldest-supported-numpy", "setuptools", "wheel"]
 build-backend = "setuptools.build_meta"


rahulporuri (approved):
LGTM

PR #848, opened 2021-06-14
Remove more code branches that supported Python 2

rahulporuri:
This PR removes 2 more code branches that supported Python 2. These code branches were hard to discover because they checked if a string is equal to a bytestring (`'' == b''`) - which is only true on Python 2 and not on Python 3. Note that these were discovered when looking at #839


aaronayres35 (approved):
LGTM

rahulporuri:
I don't think this PR needs to be backported to the 5.2 release.

PR #850, opened 2021-06-14
Set cython "language_level" explicitly to 3

rahulporuri:
This PR explicitly sets the Cython `language_level` to 3.

### DIFF
--- a/kiva/_cython_speedups.pyx
+++ b/kiva/_cython_speedups.pyx
@@ -7,6 +7,8 @@
 # is also available online at http://www.enthought.com/licenses/BSD.txt
 #
 # Thanks for using Enthought open source!
+
+# cython: language_level=3
 import numpy as np
 from numpy cimport uint8_t
 cimport _hit_test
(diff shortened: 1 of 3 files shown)


aaronayres35 (approved):
changes LGTM

On master / the maint branch I am not seeing the same `language_level` warning I was seeing with chaco... Nonetheless these changes make sense.  There are other `.pyx` files in the code base, e.g `kiva/quartz/ABCGI.pyx`, `kiva/quartz/CTFont.pyx`, etc.
Do these all need these header comments as well?

Enable does not explicitly call `cythonize` itself like chaco does.  Instead it uses `Cython.Distutils.build_ext` which calls it here https://github.com/cython/cython/blob/3e7aa76b8957e2347f3cb9dfc077897b65c5d194/Cython/Distutils/build_ext.py#L20-L21
Hence I do not know how we could cover all relevant files in one line like `extensions = cythonize(extensions, language_level = "3")` as mentioned here: https://stackoverflow.com/questions/34603628/how-to-specify-python-3-source-in-cythons-setup-py

PR #852, opened 2021-06-14
Remove is_qt4 checks

aaronayres35:
Targeting `stop-testing-pyqt4`.  This PR sits on top of #851 and removes `is_qt4` checks from the code base.

### DIFF
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -15,7 +15,7 @@ jobs:
   test-edm-linux:
     strategy:
       matrix:
-        toolkit: ['null', 'pyqt', 'pyqt5', 'pyside2', 'wx']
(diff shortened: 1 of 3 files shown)


rahulporuri:
We can't remove code that handles qt4 from the codebase yet. Removing it from the testsuite is still acceptable but not from the package itself. We'll have to close/shelve this (and PRs like these until later).

aaronayres35:
> We can't remove code that handles qt4 from the codebase yet. Removing it from the testsuite is still acceptable but not from the package itself. We'll have to close/shelve this (and PRs like these until later).

I sort of figured this was too much too soon, but opened the PR anyway as it was fresh on my mind.  We can shelve these until they are ready

corranwebster:
This is fixed by #1043

PR #855, opened 2021-06-15
Enable Release 5.2.0

aaronayres35:
The merge commit from this PR will be tagged as the 5.2.0 release


rahulporuri (approved):
LGTM.

I tested the `maint/5.2` branch before this PR today and it looked good. I tested against the null, qt toolkits and the agg, celiagg kiva backends and I didn't see any issues.

PR #858, opened 2021-06-17
Remove support for PyQt in ci helpers

rahulporuri:
PR #851 stops testing against PyQt4 on CI so this PR removes support for PyQt from the CI helpers i.e. `ci\edmtool.py`

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -53,7 +53,7 @@ using::
     python edmtool.py test_all
 
 Currently supported runtime value is `3.6``, and currently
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM

orthogonal, but it looks like enable uses pyside2 from edm although other ets packages do not.

rahulporuri:
> orthogonal, but it looks like enable uses pyside2 from edm although other ets packages do not.

PyFace and TraitsUI have valid concerns with the EDM PySide2 build - but I don't see a good reason why Enable and Chaco should care which is used.

PR #859, opened 2021-06-17
Update changelog with 5.1.1 and 5.2.0

aaronayres35:
Looks as though the changes for 5.1.1 were never pulled into the changelog on master either.
This PR simply copies the updated changelog from the `maint/5.2` branch back into master.


rahulporuri (approved):
LGTM

PR #864, opened 2021-06-25
Listen to screen DPI changes

### LINKED ISSUE #863: HighDPI/Retina drawing issues when moving window from high resolution to low
The following is part of a chaco plot when the window is opened on a hi-resolution display (at half the size of the screen-shot below):
image

This is what I get when I move the window to a lower-resolution display:
image

This looks like an issue with one of two things:
- the window not updating appropriately when the display resolution changes
- having the high-res image not scale smoothly to the smaller low-DPI image (if we can't detect the DPI change)

Or it might be a font-rendering issue.

Observed on Mac, PyQt5, and I think the standard Agg image backend.

jwiggins:
Closes #863

Note that I haven't tested this extensively. `base_pixel_scale` gets cached in a few places, so there a probably lingering corner cases.


rahulporuri:
CI is failing with the following error - 

```
ERROR: test_component_hidpi_size_stability (enable.tests.test_component_editor.TestComponentEditor)
  File "/home/runner/.edm/envs/enable-test-3.6-pyqt5/lib/python3.6/site-packages/enable/tests/test_component_editor.py", line 86, in test_component_hidpi_size_stability
    dialog.thing.window.base_pixel_scale = 2.0
  File "/home/runner/.edm/envs/enable-test-3.6-pyqt5/lib/python3.6/site-packages/traits/trait_type.py", line 67, in _read_only
    % (name, class_of(object))
traits.trait_errors.TraitError: The 'base_pixel_scale' trait of a Window instance is 'read only'.

```

corranwebster:
Revisiting this, it looks like the only place we are setting the `base_pixel_scale` is here:
https://github.com/enthought/enable/blob/718c863bdca67ddd1a9c314479f30bc03c9f69f7/enable/tests/test_component_editor.py#L86

corranwebster (approved):
This works really smoothly when moving between HiDPI laptop and regular resolution external monitor.  Manually tested using:
- agg
- celiagg
- qpainter
- quartz
- gl

The gl backend had issues with HiDPI display, but other than that everything worked as expected.

jwiggins:
Wait... How is the Wx test passing now?

corranwebster:
> Wait... How is the Wx test passing now?

Two tests - one for Wx (the old one), one for Qt

jwiggins:
But that doesn't explain why setting `base_pixel_scale` in the Wx test doesn't raise a `TraitError`. What am I missing?

corranwebster:
> But that doesn't explain why setting `base_pixel_scale` in the Wx test doesn't raise a `TraitError`. What am I missing?

That the original change was on the Qt-specific `Window` subclass and not the Wx one.  This PR doesn't attempt to handle the Wx case, I think.

jwiggins:
Right... Thanks for helping me finally understand the code that I wrote. 😞

And thanks for seeing this change through! LGTM.

PR #865, opened 2021-07-07
Use "ubuntu-latest" for GitHub Actions CI

rahulporuri:
This PR updates GitHub Actions CI to use `ubuntu-latest` instead of older `ubuntu-*` versions. This is possible now as #360 has been resolved and we no longer need to pin ubuntu to older versions.

### DIFF
--- a/.github/workflows/bleeding-edge.yml
+++ b/.github/workflows/bleeding-edge.yml
@@ -10,9 +10,7 @@ jobs:
   test-ets:
     strategy:
       matrix:
-        # Note that Ubuntu 20 cannot be used because its SWIG version is 4.0
-        # (enthought/enable#360)
-        os: [ubuntu-18.04]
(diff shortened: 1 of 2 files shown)


rahulporuri:
Most of the test failures end in this way - 

```
...

    import wx
  File "/home/runner/.edm/envs/enable-test-3.6-wx/lib/python3.6/site-packages/wx/__init__.py", line 12, in <module>
    __version__ = wx.__version__.VERSION_STRING
AttributeError: module 'wx' has no attribute '__version__'
```

This can be replicated offline

> ModuleNotFoundError: No module named 'wx.__version__.VERSION_STRING'; 'wx.__version__' is not a package

Note that the version I was using locally was 4.1.1 - whereas we pin the version to < 4.1 in traitsui and pyface.

PR #867, opened 2021-07-09
Make it possible to manually trigger cron job

rahulporuri:
This PR makes it possible to manually trigger the cron job - using `workflow_dispatch`. Ref https://github.com/enthought/ets/issues/68


aaronayres35 (approved):
LGTM

PR #871, opened 2021-07-14
Fix bug when no fonts are available

### LINKED ISSUE #870: Font machinery crashes hard when no fonts are available on a machine
Users have reported a hard crash in an application running on a headless machine. The subset of code that triggers the exception is

```
gc = GraphicsContext(temp_frame, pix_format="rgb24")
gc.set_font(Font(face_name="Verdana", size=self.size))
```

The full traceback is

```
  File "[edm environment]/lib/python3.6/site-packages/kiva/fonttools/font_manager.py", line 412, in _load_from_cache_or_rebuild
    fontManager = _pickle_load(cache_file)
  File "[edm environment]/lib/python3.6/site-packages/kiva/fonttools/font_manager.py", line 460, in _pickle_load
    fh = open(filename, "rb")
FileNotFoundError: [Errno 2] No such file or directory: '/home/centos/.enthought/kiva/fontList.cache'

[Traceback from private package]
    gc.set_font(Font(face_name="Verdana", size=self.size))
  File "[edm environment]/lib/python3.6/site-packages/kiva/agg/agg.py", line 1216, in set_font
    spec = font.findfont()
  File "[edm environment]/lib/python3.6/site-packages/kiva/fonttools/font.py", line 130, in findfont
    return default_font_manager().findfont(query)
  File "[edm environment]/lib/python3.6/site-packages/kiva/fonttools/font_manager.py", line 65, in default_font_manager
    fontManager = _load_from_cache_or_rebuild(_get_font_cache_path())
  File "[edm environment]/lib/python3.6/site-packages/kiva/fonttools/font_manager.py", line 419, in _load_from_cache_or_rebuild
    fontManager = _new_font_manager(cache_file)
  File "[edm environment]/lib/python3.6/site-packages/kiva/fonttools/font_manager.py", line 437, in _new_font_manager
    fontManager = FontManager()
  File "[edm environment]/lib/python3.6/site-packages/kiva/fonttools/font_manager.py", line 110, in __init__
    self.default_font["ttf"] = ttffiles[0]
IndexError: list index out of range
```

Note that the user also reported seeing the following warning on the machine.

```
In [2]: from kiva.image import GraphicsContext 
   ...:                                                                                                                                                                                                          
/home/msmarte/.edm/envs/kiva-test/lib/python3.6/site-packages/kiva/image.py:19: Warning: Error initializing Agg: No module named '_plat_support'
  from .agg import GraphicsContextArray as GraphicsContext
```

At the moment, the culprit seems to be 

https://github.com/enthought/enable/blob/7b9eff4d4185bc454b53d188afa1608371a5d3f7/kiva/fonttools/font_manager.py#L101-L110

aaronayres35:
This PR will fix #870 

~Currently it only includes a failing regression test. Current plan for a fix is to include a basic freely licensed font with kiva that can be used as a default if we can't find anything else.~ EDIT: I've added a quick fix of using a `TestTTF.ttf` font we already had included in the package for testing as the default if we can't find _any_ other fonts. This prevents the failure, but we may want to include a more legitimate default with the package?  TBH I don't know very much at all about fonts / ttf so I'm not sure what would make a good default here.


kiva/fonttools/font_manager.py
```diff
@@ -107,7 +107,15 @@ def __init__(self):
...
             # use anything
-            self.default_font["ttf"] = ttffiles[0]
+            if len(ttffiles) > 0:
+                self.default_font["ttf"] = ttffiles[0]
+            else:  # no available fonts, use TestTTF font
+                import pkg_resources
+                data_dir = pkg_resources.resource_filename(
+                    "kiva.fonttools.tests", "data"
+                )
+                path = os.path.join(data_dir, "TestTTF.ttf")
```
aaronayres35, line 117: I don't know very much about fonts, and I imagine this `TestTTF.ttf` is nott going to be useful as a default.  At a minimum though, this prevents the test from failing.

If this font isn't suitable for downstream needs, then one could use `add_application_font` (which previously would fail because it required initializing a font manager.

rahulporuri:
I've removed the WIP label.

kiva/fonttools/font_manager.py
jwiggins: I'm not sure this is a good solution. I think it might be better just to emit a warning and leave the default font unset. This `TestTTF.ttf` font isn't much more than a file header. It's extremely minimal.

jwiggins (commented):
I think we should raise a warning here as suggested by @rahulporuri in #870

jwiggins:
Celiagg includes the free [Montserrat font](https://github.com/JulietaUla/Montserrat).

kiva/fonttools/tests/test_font_manager.py
```diff
@@ -83,6 +83,15 @@ def test_no_import_side_effect(self):
...
                 modules[module_name] = original_module
 
+    # regression test for enthought/enable#870
+    def test_no_available_fonts(self):
+        # ensure we are not using an already existing global font manager or
+        # font cache and that we have no system fonts available
+        with patch_global_font_manager(None):
+            with patch_font_cache(self.temp_dir, []):
+                # should not fail
+                font_manager_module.default_font_manager()
```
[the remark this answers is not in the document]
aaronayres35: I realize `patch_font_cache` already causes `font_manager_module._new_font_manager(cache_file)` to be called which triggers an instantiation of a `FontManager`, so calling `default_font_manager()` is unnecessary.

kiva/fonttools/font_manager.py
```diff
@@ -107,7 +107,22 @@ def __init__(self):
...
+                    " Montserrat-Regular as a default which may not support"
+                    " all desired functionality. If needed please add a font"
+                    " with ``kiva.api.add_application_font``",
+                    stacklevel=2
+                )
+                import pkg_resources
+                data_dir = pkg_resources.resource_filename(
+                    "kiva.fonttools", "data"
+                )
+                path = os.path.join(data_dir, " Montserrat-Regular.ttf")
```
jwiggins, line 124: This doesn't exist yet, but maybe you realize that...

aaronayres35: Whoops, thank you! I had them added locally but didn't push them 🤦

kiva/fonttools/data/OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2011 The Montserrat Project Authors (https://github.com/JulietaUla/Montserrat)
```
aaronayres35, line 1: perhaps I should move this to `kiva/fonttools/LICENSES` instead...

jwiggins:
Due to the new file, you will also need to update:
* font_LICENSE.txt
* MANIFEST.in
* setup.py

aaronayres35:
Thanks @jwiggins!

font_LICENSE.txt
```diff
@@ -1,10 +1,13 @@
 Project                 License                   File
```
aaronayres35, line 1: I'm not certain I updated this file correctly

jwiggins (approved):
I would test the packaging, but otherwise LGTM

font_LICENSE.txt
jwiggins: It looks fine to me

aaronayres35:
I built an sdist and saw the font file was included as expected.  I then pip installed the sdist in a fresh env and ran the test added here which passed (ie the font was available).  Gonna go ahead and merge now

jwiggins:
Thanks for the quick fix here @aaronayres35

PR #888, opened 2021-11-08
Fix svg backend issue with pyparsing version >= 3

### LINKED ISSUE #887: Strange import failure on RHEL 7
Since updating my `enable` installation to `5.2.1`, via building from source, I'm getting the following strange import error, only on RHEL 7. I don't get this error when running under MacOS.

```
    import enable.savage.svg.css
  File "foo/lib/python3.8/site-packages/enable/savage/svg/css/__init__.py", line 10, in <module>
    from .transform import transformList
  File "foo/lib/python3.8/site-packages/enable/savage/svg/css/transform.py", line 18, in <module>
    from enable.savage.svg.pathdata import number, maybeComma
  File "foo/lib/python3.8/site-packages/enable/savage/svg/pathdata.py", line 117, in <module>
    lineTo = Group(Command("L") + Arguments(coordinatePairSequence))
  File "foo/lib/python3.8/site-packages/enable/savage/svg/pathdata.py", line 29, in Command
    return CaselessPreservingLiteral(char)
  File "foo/lib/python3.8/site-packages/enable/savage/svg/pathdata.py", line 43, in __init__
    self.name = "'%s'" % matchString
AttributeError: can't set attribute
```

rahulporuri:
fixes #887 

This PR fixes an issue because of an upstream change with the `pyparsing` `CaselessLiteral` class, where the `name` attribute became a read-only property in version >= 3. We now use the new public `set_name` method instead of directly setting the attribute.

Confusingly, this isn't documented in the API changes section of the documentation - https://pyparsing-docs.readthedocs.io/en/latest/whats_new_in_3_0_0.html#api-changes but it can be found by digging into the codebase https://github.com/pyparsing/pyparsing/blob/0352555d968f9952800f0728383de8e1f9526e1e/pyparsing/core.py#L1802-L1805. Note that `CaselessLiteral` inherits from `Token` which in turn inherits from `ParserElement`, which contains the property `name` and the new `set_name` public API.

Note to reviewer : I tested this branch locally in an environment with the latest pyparsing from PyPI (3.0.5) and the testsuite passed.

Todo
- [x] add a regression test

### DIFF
--- a/enable/savage/svg/pathdata.py
+++ b/enable/savage/svg/pathdata.py
@@ -16,9 +16,7 @@
             pass
 """
 
-from packaging.version import parse
 from pyparsing import (
-    __version__ as pyparsing_version_str,
     CaselessLiteral, Combine, Group, Literal, OneOrMore, Optional,
     ParseException, Word, ZeroOrMore, nums, oneOf
 )
@@ -43,13 +41,12 @@ class CaselessPreservingLiteral(CaselessLiteral):
     def __init__(self, matchString):
         super().__init__(matchString.upper())
 
-        pyparsing_version = parse(pyparsing_version_str)
-        if pyparsing_version.major == 2:
-            self.name = "'%s'" % matchString
-        elif pyparsing_version.major == 3:
-            self.set_name("'%s'" % matchString)
-
-        self.errmsg = "Expected " + self.name
+        quoted_name = f"'{matchString}'"
+        if hasattr(self, "set_name"):
+            # Only available in pyparsing >= 3
+            self.set_name(quoted_name)
+        else:
+            self.setName(quoted_name)
 
     def parseImpl(self, instring, loc, doActions=True):
         test = instring[loc:loc + self.matchLen]
--- a/enable/savage/svg/pathdata.py
+++ b/enable/savage/svg/pathdata.py
@@ -16,7 +16,9 @@
             pass
 """
 
+from packaging.version import parse
 from pyparsing import (
+    __version__ as pyparsing_version_str,
     CaselessLiteral, Combine, Group, Literal, OneOrMore, Optional,
     ParseException, Word, ZeroOrMore, nums, oneOf
 )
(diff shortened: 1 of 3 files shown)


enable/savage/svg/pathdata.py
```diff
@@ -40,7 +40,7 @@ class CaselessPreservingLiteral(CaselessLiteral):
 
     def __init__(self, matchString):
         super().__init__(matchString.upper())
-        self.name = "'%s'" % matchString
+        self.set_name("'%s'" % matchString)
```
rahulporuri, line 43: this will obviously not work for `pyparsing` v < 3 and we cannot drop support for `pyparsing` v < 3 yet so this needs to be written to handle the different versions of `pyparsing` differently.

rahulporuri:
Sigh. The linux jobs are waiting because they're looking for non-existent ubuntu-16.04 workers. See https://github.com/enthought/enable/pull/889

enable/savage/svg/pathdata.py
```diff
@@ -40,7 +42,13 @@ class CaselessPreservingLiteral(CaselessLiteral):
...
         super().__init__(matchString.upper())
-        self.name = "'%s'" % matchString
+
+        pyparsing_version = parse(pyparsing_version_str)
+        if pyparsing_version.major == 2:
+            self.name = "'%s'" % matchString
+        elif pyparsing_version.major == 3:
+            self.set_name("'%s'" % matchString)
+
         self.errmsg = "Expected " + self.name
```
jwiggins, line 52: Checking the library version every time seems a bit too much (and could be brittle). I think the better solution here is to use `setName` in the old version and `set_name` in the new version. Something like:
```python
quoted_name = f"'{matchString}'"
if hasattr(self, "set_name"):
    # New version 3+ API
    self.set_name(quoted_name)
else:
    # Old API
    self.setName(quoted_name)
```
Note that pyparsing sets `self.errmsg` in both versions, so this makes things slightly simpler here.

[Old version](https://github.com/pyparsing/pyparsing/blob/pyparsing_2.4.7/pyparsing.py#L1452-L1465)
[New version](https://github.com/pyparsing/pyparsing/blob/0352555d968f9952800f0728383de8e1f9526e1e/pyparsing/core.py#L1789-L1800)

rahulporuri: I wasn't sure about using the old `setName` API specifically because it was doing more than just updating the `name` attribute so I wasn't sure if it was fit for purpose.

Let's go with your solution if it is in fact fit for purpose.

jwiggins: Other than the
```python
if __diag__.enable_debug_on_named_expressions:
    self.setDebug()
return self
```
part, it's exactly the same as the code in our `__init__` method.

jwiggins (approved):
Thanks for the update. LGTM.

PR #889, opened 2021-11-08
Test on ubuntu-18.04, not 16.04

rahulporuri:
This PR updates the CI to use `ubuntu-18.04`, not `ubuntu-16.04` as it is no longer available on GitHub Actions CI. Accordingly, we also install the necessary `libsdl`.

### DIFF
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         toolkit: ['null', 'pyqt5', 'pyside2', 'wx']
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
       - name: Install Qt dependencies for Linux
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         toolkit: ['null', 'pyqt5', 'pyside2', 'wx']
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


.github/workflows/test-with-edm.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Install Wx dependencies for Linux
         run: |
           sudo apt-get update
-          sudo apt-get install libsdl2-2.0-0
+          sudo apt-get install libsdl1.2-dev
```
aaronayres35, line 30: I notice there was an old open and closed PR https://github.com/enthought/enable/pull/865 that I believe was closed because of wx failures.

Are those still present if we try to use ubuntu-latest - hence the use of 18.04 in this PR instead?  Does this wx dependency change have any impact on those failures?

aaronayres35: EDIT: looks like in the commit history on this PR you tried ubuntu-latest, then came back - so I'm assuming latest still has some issues and that is a can of worms for a future day

aaronayres35 (approved):
LGTM

test suite is happy

.github/workflows/test-with-edm.yml
rahulporuri: > Are those still present if we try to use ubuntu-latest

I haven't tested with ubuntu-latest (ubuntu-20.04 at the moment)

> Does this wx dependency change have any impact on those failures?

I'm not sure. I didn't investigate it further.

> so I'm assuming latest still has some issues

No. I realized that I made a mistake when using ubuntu-18.04, which is why i reverted the change to ubuntu-latest. I will investigate later what happens when we use ubuntu-latest.

PR #892, opened 2022-02-22
Add integration tests from TraitsUI

corranwebster:
These are integration tests from TraitsUI that use Enable and Kiva. Having them here makes more sense than in TraitsUI in terms of test dependencies.


jwiggins:
Is this a draft because there's no machinery to run this as an automated test?

corranwebster:
Yes.

PR #897, opened 2022-03-04
Replace QMidButton with QMiddleButton.

corranwebster:
Fixes #891.

PR #900, opened 2022-03-04
Use `QOpenGLWidget`

corranwebster:
The `QOpenGLWidget` replaces the deprecated `QGLWidget`.  The only change is that we no longer need to actively call `swapBuffer`.

This PR relies on #897 being merged first, since this can't be tested on Qt6 without that PR.

### DIFF
--- a/enable/qt4/base_window.py
+++ b/enable/qt4/base_window.py
@@ -26,7 +26,7 @@
 import warnings
 
 # Qt imports.
-from pyface.qt import QtCore, QtGui, QtOpenGL
+from pyface.qt import QtCore, QtGui, QtOpenGL, is_qt4, qt_api
 
 # Enthought library imports.
 from enable.abstract_window import AbstractWindow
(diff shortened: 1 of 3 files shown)


enable/qt4/base_window.py
```diff
@@ -26,7 +26,7 @@
 import warnings
 
 # Qt imports.
-from pyface.qt import QtCore, QtGui, QtOpenGL
+from pyface.qt import QtCore, QtGui, QtOpenGL, is_qt4, is_qt5, qt_api
```
jwiggins, line 29: It looks like `is_qt5` isn't used here.

kiva/gl/__init__.py
```diff
@@ -442,8 +446,11 @@ def GetLabel(text, pyglet_font):
         return label
 
 
-except ImportError:
+except ImportError as exc:
     # Pyglet is not available, so we forgo some features
+    logger.info(
```
jwiggins, line 451: Why not `logger.exception`?

jwiggins (approved):
LGTM w/ a couple comments

kiva/gl/__init__.py
corranwebster: It felt more natural to go from quietly carrying with no warning to giving an info worning which might be useful, vs a logged exception.

Possibly a good idea to at least include the exception info, though, since otherwise it may be hard to workout what is missing.

jwiggins: Sounds good

jwiggins (approved):

PR #901, opened 2022-03-07
Fix text locale test issue on Windows

### LINKED ISSUE #899: Regression with locale tests on Windows/Wx
We are now consistently getting this exception in Windows test with Wx:
```
ERROR: test_locale_independence (kiva.tests.agg.test_text.TestText)
  File "D:\a\enable\enable\.edm\envs\enable-test-3.6-wx\lib\site-packages\kiva\tests\agg\test_text.py", line 47, in test_locale_independence
    x0, _ = gc.get_text_position()
  File "D:\a\enable\enable\.edm\envs\enable-test-3.6-wx\lib\contextlib.py", line 88, in __exit__
    next(self.gen)
  File "D:\a\enable\enable\.edm\envs\enable-test-3.6-wx\lib\site-packages\kiva\tests\agg\test_text.py", line 30, in locale_context
    locale.setlocale(category, old)
  File "D:\a\enable\enable\.edm\envs\enable-test-3.6-wx\lib\locale.py", line 598, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting

Ran 381 tests in 101.942s

FAILED (errors=1, skipped=62, expected failures=1)
```

This looks like there is some issue in setting the locale back to its original value (which shouldn't ever be a problem.

corranwebster:
This attempts to set the locale to the current value before changing to the testing locale to ensure that a reset will work.  On failure, it skips which matches previous behaviour.

Fixes #899

### DIFF
--- a/kiva/tests/agg/test_text.py
+++ b/kiva/tests/agg/test_text.py
@@ -50,7 +50,7 @@ class TestText(unittest.TestCase):
                 gc.show_text(text)
                 x0, _ = gc.get_text_position()
 
-        with locale_context(locale.LC_CTYPE, ("en", "ASCII")):
+        with locale_context(locale.LC_CTYPE, ("C", "ASCII")):
             gc = agg.GraphicsContextArray((200, 200))
             f = Font("modern")
             with gc:
--- a/kiva/tests/agg/test_text.py
+++ b/kiva/tests/agg/test_text.py
@@ -9,6 +9,7 @@
 # Thanks for using Enthought open source!
 from contextlib import contextmanager
 import locale
+import sys
 import unittest
 
 from kiva import agg
@@ -21,9 +22,7 @@ def locale_context(category, new=None):
     """
     old = locale.getlocale(category)
     try:
-        # check that we can set the current locale to ensure finally will work
         # see Enable #899
-        locale.setlocale(category, old)
         locale.setlocale(category, new)
     except locale.Error as e:
         raise unittest.SkipTest(str(e))
@@ -34,6 +33,12 @@ def locale_context(category, new=None):
 
 
 class TestText(unittest.TestCase):
+
+    @unittest.skipIf(
+        sys.platform == "win32",
+        "Skipping on windows due to issues with setlocale. "
+        "See https://bugs.python.org/issue38324",
+    )
     def test_locale_independence(self):
         # Ensure that >ASCII Unicode text is decoded correctly regardless of
         # the locale.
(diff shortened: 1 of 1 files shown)


kiva/tests/agg/test_text.py
```diff
@@ -21,6 +21,9 @@ def locale_context(category, new=None):
     """
     old = locale.getlocale(category)
     try:
+        # check that we can set the current locale to ensure finally will work
```
rahulporuri, line 24: "to ensure finally" - it looks like something is missing here.

rahulporuri:
With the changes in this PR, the test is still being skipped. Looking into this further, it looks like this call is what is raising the `locale.Error`

https://github.com/enthought/enable/blob/754d3f05b5909fa2bc081335bcb0dd0c3627e40a/kiva/tests/agg/test_text.py#L52

and it looks like there is something fundamentally wrong with the test now - 

```
>>> import locale
>>> locale.getlocale(locale.LC_CTYPE)
(None, None)
>>> locale.setlocale(locale.LC_CTYPE, ('en', 'ASCII'))
  File "<stdin>", line 1, in <module>
  File "C:\Users\rporuri\.edm\envs\enable-test-3.6-wx\lib\locale.py", line 598, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting
>>> locale.setlocale(locale.LC_CTYPE, ('en', 'UTF-8'))
'en_US.UTF-8'
```

rahulporuri (commented):
It looks like there are issues with `locale.getlocale`/`locale.setlocale` on windows so let's just explicitly skip the test on windows instead of indirectly skipping them in this way.

kiva/tests/agg/test_text.py
corranwebster: This comment and the line it refers to have now been removed.

corranwebster:
Did the locale test work on anything other than mac?

rahulporuri:
> Did the locale test work on anything other than mac?

ha. looks like @rkern introduced the try/except because locale setting was failing on linux - https://github.com/enthought/enable/commit/4abc50026c0efc0003f4d0ab8581feb76659dfd9

corranwebster:
Possibly we should be using ("C", "ASCII")?  Still doesn't work on windows.

rahulporuri:
> Possibly we should be using ("C", "ASCII")? Still doesn't work on windows.

Hey that actually works on Python 3.8.10 -

```
>>> import locale
>>> locale.setlocale(locale.LC_CTYPE, ('C', 'ASCII'))
'C'
```

corranwebster:
Trying now with all skipping turned off - I think the `("en", "UTF-8")` will still fail on windows, but there may be an equivalent locale setting we can use.

I think it's the ASCII vs UTF-8 stuff that it's testing, rather than the language part of the setting, but it might be useful to have @rkern check that it's OK.

corranwebster:
Aha - `('en', 'utf8')` works on Windows.

PR #916, opened 2022-03-16
Fix Qt6 mouse event code

### LINKED ISSUE #915: QMouseWheelEvent constructor signature changed in Qt 5.12
As a result, MouseWheelEvent tests are failing in recent versions of Qt backends.

corranwebster:
This fixes #915 but additionally will fix code from other Qt mouse events, since the position API has changed - there are no longer `x` and `y` position getters available in Qt6.

This:
- uses the new API for position for Qt6 mouse events
- fixes the test code that generates synthetic mouse events for Qt6

### DIFF
--- a/enable/tests/qt4/test_mouse_wheel.py
+++ b/enable/tests/qt4/test_mouse_wheel.py
@@ -50,34 +50,34 @@ class MouseWheelTestCase(TestCase):
         # create and mock a mouse wheel event
         if is_qt4:
             qt_event = QtGui.QWheelEvent(
-                QtCore.QPoint(0, 0),
-                200,
-                QtCore.Qt.NoButton,
-                QtCore.Qt.NoModifier,
-                QtCore.Qt.Vertical,
+                QtCore.QPoint(0, 0),  # pos
+                200,  # delta
+                QtCore.Qt.NoButton,  # buttons
+                QtCore.Qt.NoModifier,  # modifiers
+                QtCore.Qt.Vertical,  # orient
             )
         elif is_qt5:
             qt_event = QtGui.QWheelEvent(
-                pos=QtCore.QPointF(0, 0),
-                globalPos=self.window.control.mapToGlobal(QtCore.QPoint(0, 0)),
-                pixelDelta=QtCore.QPoint(0, 200),
-                angleDelta=QtCore.QPoint(0, 200),
-                qt4Delta=200,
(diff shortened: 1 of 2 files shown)


enable/tests/qt4/test_mouse_wheel.py
```diff
@@ -92,18 +100,29 @@ def test_horizontal_mouse_wheel(self):
...
+            qt_event = QtGui.QWheelEvent(
+                QtCore.QPointF(0, 0),  #pos
+                self.window.control.mapToGlobal(QtCore.QPointF(0, 0)),  # global pos
+                QtCore.QPoint(200, 0),  # pixel delta
+                QtCore.QPoint(200, 0),  # angle delta
+                QtCore.Qt.NoButton,  # mouse buttons
+                QtCore.Qt.NoModifier,  # keyboard modifiers
+                QtCore.Qt.ScrollUpdate,  # phase
+                False,  #inverted
+            )
```
aaronayres35, line 125: maybe move these comments up to start at line 72

aaronayres35: also maybe would be nice to have similar comments for the qt5 case

enable/tests/qt4/test_mouse_wheel.py
```diff
@@ -57,7 +56,7 @@ def test_vertical_mouse_wheel(self):
                 QtCore.Qt.NoModifier,
                 QtCore.Qt.Vertical,
             )
-        else:
+        elif is_qt5:
             qt_event = QtGui.QWheelEvent(
                 QtCore.QPoint(0, 0),
                 self.window.control.mapToGlobal(QtCore.QPoint(0, 0)),
```
aaronayres35, line 62: nit: should these be QPointF as well as opposed to just QPoint?

ref: https://doc.qt.io/qt-5/qwheelevent-obsolete.html

aaronayres35 (approved):
LGTM just a couple minor comments

enable/tests/qt4/test_mouse_wheel.py
corranwebster: Actually, the right way improve readability is probably to use keyword arguments.

enable/tests/qt4/test_mouse_wheel.py
corranwebster: Apparently PySide2 doesn't like this.

PR #918, opened 2022-03-24
Improve font weight parsing

### LINKED ISSUE #908: Kiva fonttool weight constants don't match OpenType and CSS specs
The names of the weights are defined here: https://docs.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass

But these don't match the names defined here in all cases: https://github.com/enthought/enable/blob/718c863bdca67ddd1a9c314479f30bc03c9f69f7/kiva/fonttools/_constants.py#L90-L105

For example "light" is 300 in both specs, but 200 in the constants file.

This is likely to result in mis-matches in font queries because `FontEntry` objects have the wrong value.

Fixing this may require bumping the font database version and re-generating the font database.

There is a potential tension here because the constants are used for scoring closeness of match, but are also used for interpreting the contents of font files.

corranwebster:
This makes the following changes:

- corrects the font weight constants to match the OpenType values
- where possible, gets the font weight, stetch and slope from OS/2 data table
- adjusts the way FontEntry values are computed to account for these changes
- allows font entry weights which are not multiples of 100, but restricts them to range of 100 to 1000
- bumps the font database version number to force a recompute

This fixes #908.


kiva/fonttools/_scan_parse.py
```diff
@@ -224,13 +224,16 @@ def _ttf_font_property(fpath, font, face_index=0):
...
     #    lighter and bolder are also allowed.
-    weight = None
-    for w in weight_dict.keys():
-        if style_prop.find(w) >= 0:
-            weight = w
-            break
+    # We prefer weight values from the OS/2 table, if available, otherwise
+    # infer from the "name" table's style
+    weight = props.get("weight")
     if not weight:
```
rkern, line 230: I'm pretty sure this was meant to be `if weight is None:`, just with the bad habits of years past.

rkern (approved):

PR #919, opened 2022-03-25
Fix font weights

### LINKED ISSUE #905: Font queries ignore `Font.weight`
The `FontManager` and `FontQuery` system ignore the `Font.weight` attribute and look for a `style` which holds flags for `BOLD`, `ITALIC`, `BOLD_ITALIC`, etc.  See: https://github.com/enthought/enable/blob/62f390f8cf6c29536e1a70034db170197e489eef/kiva/fonttools/font.py#L149-L169

But the `str_to_font` function populates `weight`. See: https://github.com/enthought/enable/blob/62f390f8cf6c29536e1a70034db170197e489eef/kiva/fonttools/font.py#L54-L55

As does the trait validator: https://github.com/enthought/enable/blob/62f390f8cf6c29536e1a70034db170197e489eef/kiva/trait_defs/kiva_font_trait.py#L122-L123

As does the Wx editor:
https://github.com/enthought/enable/blob/62f390f8cf6c29536e1a70034db170197e489eef/kiva/trait_defs/ui/wx/kiva_font_editor.py#L98-L100

To reproduce, you can use `kvia_explorer.py` with this script and the agg backend:
```
# The graphics context is available as gc.

from kiva.api import (
    Font, BOLD_ITALIC, BOLD, ITALIC
)
from kiva.fonttools import str_to_font 

font_1 = Font(
    'Courier New',
    style=BOLD_ITALIC,
    size=24,
)
font_2 = str_to_font('24 pt italic bold Courier New')


with gc:
    gc.set_font(font_1)
    gc.show_text_at_point(f"font_1: weight={font_1.weight} style={font_1.style}", 20, 60)
    gc.set_font(font_2)
    gc.show_text_at_point(f"font_2: weight={font_2.weight} style={font_2.style}", 20, 20)
```

Which produces:
image

We would expect the second line to be bold.

Most backends suffer from this: Cairo does not, and a number of backends ignore bold/italic

corranwebster:
This is a reasonably comprehensive fix for #905 following the suggestions given there.  Together with #918 it gives good support for more refined font weights and styles on `qpainter` and both agg backends, as well as consistent behaviour on backends which only support bold and normal weights.  It is also backwards compatible with code that uses `style` for both italic and bold, raising a `DeprectationWarning` if the `style` is bold.

In making the changes, this PR also cleans up some of the early Python 2-ish idioms and uses more modern constructs where appropriate.

It includes smoke tests for rendering text in all combinations of font families, weights and styles on each backend, as well as adding tests for the `KivaFont` trait.  Documentation is also updated.

This PR doesn't attempt to update backends which can potentially support more font weights (eg. quartz) but which haven't exposed it in the C/C++ wrappers.  It also doesn't attempt to update or improve the font parsers to understand the new weight types - that should be a different PR.

Fixes #905.


rkern (approved):
Minor readability suggestion, but LGTM!

PR #931, opened 2022-04-12
FIX: Use correct path to the default font

### LINKED ISSUE #930: Calling `Font.findfont` results in `RecursionError` if no fonts are installed on the system
The fallback font for font-less systems is not quite working properly:

Steps to reproduce:
```
>>> from kiva.api import Font
>>> font = Font()
>>> font.findfont()
```

Looks like it may be due to an erroneous space in the font filename:
https://github.com/enthought/enable/blob/fa865d8b4aa055056de07aab0d88f54c644b05ce/kiva/fonttools/font_manager.py#L123

ffishburn:
Closes #930. Fixes a typo in the default font filename

### DIFF
--- a/kiva/fonttools/font_manager.py
+++ b/kiva/fonttools/font_manager.py
@@ -120,7 +120,7 @@ class FontManager:
                 data_dir = pkg_resources.resource_filename(
                     "kiva.fonttools", "data"
                 )
-                path = os.path.join(data_dir, " Montserrat-Regular.ttf")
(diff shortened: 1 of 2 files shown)


jwiggins (commented):
Good eye. Can you add a unit test which actually exercises this code path?

corranwebster:
I am adding tests for this - most of the machinery is already in the test suite so it's fairly straighforward.

jwiggins (approved):
LGTM. Thanks for the test coverage.

Issue #940, opened 2022-04-28
Release 5.3.0


corranwebster:
Pre-release
---
- [x] Backport PRs that have been merged to ``main`` to the maintenance branch. Use the "need backport ..." tag if there is one (but don't rely 100% on it)
- [ ] Verify that no other open issue needs to be addressed before the release.
- [ ] Test against other ETS packages and other ETS-using projects
- [ ] Check MANIFEST, requirements, changelog are still up to date.
- [ ] Test building the documentation

Major/Minor Release
-------------------
- [ ] Create branch `release/<release version number>` from `main` branch.
  - [ ] Set release to `<release version number>`, and set `IS_RELEASED` is true; commit
  - [ ] Install from source distribution and run tests again
  - [ ] Open a PR against `main` with this being the last commit so that CI is built on the release commit
  - [ ] merge the PR when it is approved
- [ ] Create a post-release PR branch from `main` branch.
  - [ ] Bump the minor version number i.e. `<release version number + 0.1.0>` and set `IS_RELEASED` to false; commit.
  - [ ] Open a PR against `main`
  - [ ] Merge the PR when it is approved
- [ ] From the commit at which `IS_RELEASED` is true and version is `<release version number>`, create a new maintenance branch `maint/<major version>.<minor version>` and push it to github
  - create a PR branch against `maint/<release version number>`
  - [ ] Bump the micro version number i.e. `<release version number + 0.0.1>` and set `IS_RELEASED` to false; commit.
  - [ ] open a PR against `maint/<release version number>`
  - [ ] Merge the PR when it is approved
- [ ] From the commit at which `IS_RELEASED` is true and version is `<release version number>`, tag (annotated!) `git tag -a -m "Release <release version number>" <release version number>`
  - [ ] Push the tag `git push origin <release version number>`
- [ ] Make PR targeting `gh-pages` branch: Generate documentation and copy the content to the branch. Verify that the resulting index.html looks good.
- [ ] Update GitHub Release pages https://github.com/enthought/pyface/releases
  A Github workflow should automatically publish the release to PyPI.
- [ ] Test the PyPI package

Maintenance Release
-------------------
- [ ] Create branch `release/<release version number>` from `maint/<release version number>` branch.
  - [ ] Set release to `<release version number>`, and set `IS_RELEASED` is true; commit
  - [ ] Install from source distribution and run tests again
  - [ ] Open a PR against `maint/<release version number>` with this being the last commit so that CI is built on the release commit
[...]

PR #941, opened 2022-04-28
Release 5.3.0

corranwebster:
This is the release PR

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ MAJOR = 5
 MINOR = 3
 MICRO = 0
 PRERELEASE = ""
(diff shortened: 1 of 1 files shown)


jwiggins:
This is targeted at the `main` branch. Where are you planning to make the release tag?

corranwebster:
On main.

Plan is merge this so we have a release on main with the release flag set, then a second PR which un-sets it.  Tag the release commit on master, and maintenance branch is created from that as well.  This is what we've been doing on other ETS projects recently.

See #940

jwiggins (approved):
Thanks for the pointer to #940. LGTM

PR #948, opened 2022-06-11
Use the "sphinx-copybutton" extension in documentation

rahulporuri:
This PR does the following: 
- Use the `sphinx_copybutton` extension when building the Sphinx documentation
- Create a `docs` extra for the `enable` package and add `sphinx-copybutton` as a `docs` dependency along with `enthought-sphinx-theme` and `sphinx`
- Install `sphinx-copybutton` from PyPI when building docs using the `ci/edmtool.py` utility.

Ref https://sphinx-copybutton.readthedocs.io/en/latest/


ci/edmtool.py
```diff
@@ -294,6 +294,9 @@ def docs(runtime, toolkit, environment):
     commands = [
         "edm install -y -e {environment} " + packages,
     ]
+    commands += [
+        "edm run -e {environment} -- python -m pip install sphinx-copybutton"
+    ]
```
jwiggins, line 299: Why not just:
```python
commands = [
    "edm install -y -e {environment} " + packages,
    "edm run -e {environment} -- python -m pip install sphinx-copybutton"
]
```
?

docs/source/conf.py
```diff
@@ -98,6 +99,12 @@
 # The name of the Pygments (syntax highlighting) style to use.
 pygments_style = 'sphinx'
 
+# Options for Sphinx copybutton extension
+# ---------------------------------------
+
+# Matches prompts - "$ ", ">>>" and "..."
+copybutton_prompt_text = r">>> |\.\.\. |\$ "
```
jwiggins, line 106: I feel like ellipses might be used in places which aren't code prompts. Or am I misunderstanding how this works?

jwiggins (approved):

ci/edmtool.py
rahulporuri: i am blind. resolved.

docs/source/conf.py
rahulporuri: Triple ellipses `...`?

The intention here is to support codeblocks that look like

```
>>> CustomClass(
...    var_a=1,
...    var_b=2,
... )
```

jwiggins: Ahh OK. That makes more sense.

rahulporuri: Thanks John! Merging!

PR #957, opened 2022-07-22
Remove generated C++ files

### LINKED ISSUE #956: Don't ship C/C++ files generated by Cython
Previously we included the generated C/C++ files for the various Cython extension modules in the repo and distributions.  The thought for this was that then people installing Enable didn't need to have Cython installed in their build environment, and the generated code didn't depend on the Cython version that people had.

Now we specify Cython as a build dependency in the `pyproject.toml` and if we need to we can even pin the particular Cython version there.  As a result we don't need to include the generated C/C++ code, particularly as it may be contributing to some issues such as #949.

corranwebster:
We now assume Cython is available in the build environment (specified by the `pyproject.toml`).

Fixes #956.


corranwebster:
Deleted a file that wasn't Cython-generated.

PR #963, opened 2022-07-24
More Benchmarks

corranwebster:
Currently there is almost no agreement over what behaviour should be in a number of backends.

This PR adds some benchmarks to the suite to explore:

- arc drawing
- font selection


jwiggins (approved):
LGTM

PR #966, opened 2022-07-25
Fix Kiva Quartz backend string encoding

### LINKED ISSUE #964: Quartz `CTFont` doesn't handle Python 3 bytes to str properly
See the following, which is the result of calling `str` on a bytes (Python 2 idiom) instead of explicitly using a codec.
```
In [2]: from kiva.quartz import CTFont                                          

In [3]: CTFont.default_font_info.names()                                        
Out[3]: 
["b'Kannada MN'",
 "b'Malayalam Sangam MN'",
 "b'Noto Sans Kannada'",
 "b'Beirut'",
 "b'DIN Alternate'",
 "b'PT Sans Caption'",
 "b'Devanagari MT'",
 "b'PingFang TC'",
 "b'Courier'",
 ... # etc
 "b'Gill Sans'"]
```

Culprit is this method:
https://github.com/enthought/enable/blob/bb2c358248ef72c03d62b2563985f01e1dad6977/kiva/quartz/CTFont.pyx#L17-L27

corranwebster:
Fixes #964.

Bytes seem to be assuming `mac-roman` encoding, so use that in decoder.

Also a drive-by fix to make quartz tests proper unittest tests rather than nose/pytest-style.

### DIFF
--- a/kiva/quartz/CTFont.pyx
+++ b/kiva/quartz/CTFont.pyx
@@ -17,14 +17,14 @@ include "CoreText.pxi"
 cdef object _cf_string_to_pystring(CFStringRef cf_string):
     cdef char* c_string
     cdef char buf[256]
-    c_string = CFStringGetCStringPtr(cf_string, kCFStringEncodingMacRoman)
+    c_string = CFStringGetCStringPtr(cf_string, kCFStringEncodingUTF8)
     if c_string == NULL:
         success = CFStringGetCString(cf_string, buf, 256,
-                    kCFStringEncodingMacRoman)
+                    kCFStringEncodingUTF8)
         retval = bytes(buf)
     else:
         retval = bytes(c_string)
-    return retval.decode('mac-roman')
+    return retval.decode('utf-8')
 
 cdef CFArrayRef _get_system_fonts():
         cdef CFIndex value = 1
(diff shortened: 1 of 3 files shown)


kiva/quartz/CTFont.pyx
```diff
@@ -21,10 +21,10 @@ cdef object _cf_string_to_pystring(CFStringRef cf_string):
     if c_string == NULL:
         success = CFStringGetCString(cf_string, buf, 256,
                     kCFStringEncodingMacRoman)
-        retval = str(buf)
+        retval = bytes(buf)
     else:
-        retval = str(c_string)
-    return retval
+        retval = bytes(c_string)
+    return retval.decode('mac-roman')
```
jwiggins, line 27: This is only being used for font names returned by the system. Should we perhaps look into using UTF-8 instead? I know some of the non-Latin fonts have non-Latin glyphs in their names.

kiva/tests/test_quartz.py
```diff
@@ -8,72 +8,85 @@
...
+        del ABCGI
+        del CTFont
+        del mac_context
+
+    @skip_if_not_wx
+    def test_macport(self):
+        if sys.platform != "darwin":
+            from unittest.case import SkipTest
+
+            raise SkipTest("macport is only built on OS X")
```
jwiggins, line 31: This skip should no longer be needed

kiva/quartz/CTFont.pyx
corranwebster: Possibly: I think this primarily for introspection of the system, rather than actively in font selection, but I've tried replacing with UTF-8 since it seems straightforward.

PR #969, opened 2022-07-31
PIL Image Kiva Backend

corranwebster:
This implements a basic Kiva `pil_image` backend using the PIL/Pillow `ImageDraw` routines (and related code).  This is intended to be a reasonably complete reference implementation of Kiva that can be used if there are problems building the Agg backends.

There are some gaps due to limitations of the `ImageDraw` library:
- no gradients, line dashes, line joins, line caps or choice between EOF/winding number fill
- some gaps in the API where basecore2d doesn't implement things (eg. `arc_to`, `stroke_rect`/`fill_rect`, etc.; some of these are straightforward to  implement, but it should be done in basecore2d)

But it does implement:
- path drawing
- image drawing
- full font drawing, including rotation
- rectangular clipping regions

And it includes some fixes/improvements to basecore2d that became apparent when writing the code.

There are likely corner cases, particularly in text rendering or when the expected behaviour is unclear.  These can be fixed as needed if they aren't found in the review of the PR.  It is about 50 times slower than the Agg renderers.

### DIFF
--- a/enable/gcbench/bench.py
+++ b/enable/gcbench/bench.py
@@ -26,6 +26,7 @@ _BACKENDS = {
         "cairo": "enable.null.cairo",
         "celiagg": "enable.null.celiagg",
         "opengl": "enable.gcbench.opengl",
+        "pil_image": "enable.null.pil_image",
         "qpainter": "enable.null.qpainter",
         "quartz": "enable.null.quartz",
     },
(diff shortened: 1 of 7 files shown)


corranwebster:
Apparently we're tickling a bug in Pillow on Python 3.8 - the operation which is failing is pasting a single color into the image, so _should_ be fine as it is very basic.

Issue #971, opened 2022-08-04
Script font family not supported in font queries


corranwebster:
If you do
```
gc.set_font(Font(family=SCRIPT, size=24))
```
in a backend that uses `kiva.fonttools` to find font files, then you get the following error:
```
/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/kiva/fonttools/font_manager.py:294: UserWarning: findfont: Font family ['script'] not found. Falling back to sans-serif
  % (query.get_family(), self.default_family)
```

This is because `script` is not listed in the options here: https://github.com/enthought/enable/blob/7079309609a0c1b23f9134234e1984f4b53d1178/kiva/fonttools/_constants.py#L11-L76

PR #973, opened 2022-08-04
Render font families better in QPainter backend

### LINKED ISSUE #972: QPainter font families not working correctly
When running the benchmark script, the QPainter backend renders as follows:
image

In particular, font family selection seems to be being ignored, but we also have one line where we are getting a very odd choice of some script-like font.

For comparison, Agg renders like this:
image

corranwebster:
Fall back to kiva.fonttools if Qt can't come up with a font name.  Fixes #972 

Benchmark looks like this on Python 3.8/PySide6/MacOS (appearance is likely to vary significantly based on OS and fonts installed):
image

Note that script and mono families are not rendered correctly due to #971 and #974.

### DIFF
--- a/kiva/qpainter.py
+++ b/kiva/qpainter.py
@@ -675,6 +675,7 @@ class GraphicsContext(object):
                 # otherwise use kiva.fonttools to suggest one
                 name = font.findfontname()
 
+        # setFamilies() was introduced in Qt 5.13
         if hasattr(qfont, 'setFamilies'):
             qfont.setFamilies([name])
         else:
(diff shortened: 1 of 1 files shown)


kiva/qpainter.py
```diff
@@ -649,13 +659,27 @@ def select_font(self, name, size, style="regular", encoding=None):
...
+        if font.face_name:
+            name = font.face_name
+        else:
+            # if we don't have a face name, see if Qt can suggest one
+            name = qfont.defaultFamily()
+            if not name:
+                # otherwise use kiva.fonttools to suggest one
+                name = font.findfontname()
+
+        if hasattr(qfont, 'setFamilies'):
```
jwiggins, line 678: Perhaps a note here that the method was introduced in Qt 5.13 would be helpful?

jwiggins (commented):
LGTM

jwiggins (approved):

PR #975, opened 2022-08-04
Fix `SCRIPT` font family in Kiva

### LINKED ISSUE #971: Script font family not supported in font queries

corranwebster:
Fix is a matter of having `Font.familymap` map to the right thing.  This is a _very_ old bug.

This is the benchmark rendering in the agg backend after the fix (note the "script" column):
image

Fixes #971.

### DIFF
--- a/kiva/fonttools/tests/test_font.py
+++ b/kiva/fonttools/tests/test_font.py
@@ -14,9 +14,10 @@ import os
 import unittest
 
 from kiva.constants import (
-    BOLD, BOLD_ITALIC, DEFAULT, ITALIC, MODERN, NORMAL, ROMAN, WEIGHT_BOLD,
+    BOLD, BOLD_ITALIC, DECORATIVE, DEFAULT, ITALIC, MODERN, NORMAL, ROMAN, SCRIPT, TELETYPE, WEIGHT_BOLD,
     WEIGHT_LIGHT, WEIGHT_NORMAL, SWISS,
 )
+from kiva.fonttools._constants import font_family_aliases, preferred_fonts
 from kiva.fonttools.tests._testing import patch_global_font_manager
 from kiva.fonttools.font import (
     DECORATIONS, FAMILIES, NOISE, STYLES, WEIGHTS, Font, str_to_font,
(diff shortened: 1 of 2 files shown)


kiva/fonttools/tests/test_font.py
```diff
@@ -193,6 +194,23 @@ def test_font_query_warnings(self):
...
+        ]
+        for family in families:
+            with self.subTest(family=family):
+                font = Font(family=family)
+                query = font._make_font_query()
+                query_family = query.get_family()[0]
+
+                self.assertEqual(query_family, Font.familymap.get(family))
+                self.assertIn(query_family, font_family_aliases)
+                self.assertIn(query_family, preferred_fonts)
```
jwiggins, line 212: There are items in `font_family_aliases` which are not keys in `preferred_fonts`. Perhaps you should only check against `preferred_fonts`?

jwiggins (approved):
Basically LGTM

kiva/fonttools/tests/test_font.py
corranwebster: We want to check it is in both, because of this line here: https://github.com/enthought/enable/blob/4d51db21a3cae2e316afa3c7329e35eba26ea589/kiva/fonttools/_score.py#L38

If it is not on `font_family_aliases` then `preferred_fonts` will never be accessed.

jwiggins: I'm saying that `query_family in font_family_aliases` can be `True` at the same time that `query_family in preferred_fonts` is `False`.

The more complete snippet from `score_family`...
https://github.com/enthought/enable/blob/4d51db21a3cae2e316afa3c7329e35eba26ea589/kiva/fonttools/_score.py#L38-L40

... is not really being tested here because converting from the integer Kiva family constants via `Font.familymap` will never result in the values `"sans"`, `"sans serif"`, or `"modern"`.

**Is the point of this test to keep `Font.familymap` values in sync with `font_family_aliases` and `preferred_fonts`?**

corranwebster: The entries in `familymap` need to be in both, yes (and that's what I care about for this PR).

`font_family_aliases` (and presumably `preferred_fonts`) can have extra values which come from the `face_name` (this is what derailed #977).  Eg. if someone names a font "sans" (ie. `Font("sans", 12)`) it will _also_ go through the above codepath.  That likely needs some re-engineering at some point, but it is outside the scope of this PR)

PR #977, opened 2022-08-04
Map `MODERN` to `"monospace"` and `TELETYPE` to `"typewriter"`

### LINKED ISSUE #974: Modern font family not supported properly in kiva.fonttools
The term "modern" in this context comes from Wx:
> `wx.FONTFAMILY_MODERN` - A fixed pitch font.
(see [`wx.FontFamily`](https://docs.wxpython.org/wx.FontFamily.enumeration.html#wx-fontfamily)

So what it should be producing is a monospaced font such as Monaco on Mac or Andale Mono (but _not_ Courier - that's Teletype/Typewriter).  Instead it is producing a default sans-serif font.

See "mono" and "type" columns here (Agg backend):
image

corranwebster:
This changes font handling to:
- map "modern" font family to "monospace" rather than "sans-serif" like other backends
- distinguish between "monospace" (sans-serif fixed pitch fonts) and "typewriter" (serif fixed pitch fonts).

Replaces uses of "modern" with "sans-serif" where appropriate (particularly in defaults) to keep appearances the same.

This now renders the benchmark like this (see the "mono" and "type" columns in particular):
image

This is not backwards compatible: users who use "modern" font family will get different fonts, but it is easily fixed by changing the font family.  So this should probably be part of a major release.

Fixes #974.


corranwebster:
Converting this to draft as there are other paths which get to `modern` in a `FontQuery` object and which are causing failures.

corranwebster:
This has been fixed up and is ready for review.

corranwebster:
Compare with current QPainter backend:
image

dpinte (approved):
LGTM

PR #978, opened 2022-08-05
Fix Quartz font rendering

### LINKED ISSUE #976: Quartz backend doesn't support font families
From running the benchmark scripts:
image

It also doesn't color them correctly: they should render as green-ish.

corranwebster:
This looks up font family names if no specific face name is given, and switches rendering of text to use the fill color rather than the stroke color, to match Agg, etc. rendering.

Font rendering benchmark now looks like this:
image

Fixes #976.

### DIFF
--- a/kiva/quartz/ABCGI.pyx
+++ b/kiva/quartz/ABCGI.pyx
@@ -809,7 +809,11 @@ cdef class CGContext:
             constants.ITALIC: 'italic',
             constants.BOLD_ITALIC: 'bold italic',
         }[font.is_bold() | font.style]
-        self.select_font(font.face_name, font.size, style=style)
+        if font.face_name:
+            name = font.face_name
+        else:
+            name = font.findfontname()
+        self.select_font(name, font.size, style=style)
 
     def set_font_size(self, float size):
         """ Change the size of the currently selected font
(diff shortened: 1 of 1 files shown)


kiva/quartz/ABCGI.pyx
```diff
@@ -809,7 +809,11 @@ cdef class CGContext:
             constants.ITALIC: 'italic',
             constants.BOLD_ITALIC: 'bold italic',
         }[font.is_bold() | font.style]
-        self.select_font(font.face_name, font.size, style=style)
+        if font.face_name:
+            name = font.face_name
+        else:
+            name = font.findfontname()
```
jwiggins, line 815: It feels like this is something the kiva `Font` class should be managing, but perhaps that's too fraught with all of the different backend implementations.

jwiggins (approved):
LGTM

kiva/quartz/ABCGI.pyx
corranwebster: Possibly.  I'd like to see `Font` to grow more like the CSS font specifiers where you can give a bunch of font face name suggestions, including generic ones, but that's a fairly big API change that would need to be managed.

PR #980, opened 2022-08-05
Take into account all font features when drawing SVG text

### LINKED ISSUE #979: SVG Font family selection doesn't work
We get this:
image

Instead of something like this:
image

Note that colors and font selection are wrong.

corranwebster:
Fixes #979.

Benchmark now looks like this:
image


jwiggins (approved):
LGTM

PR #990, opened 2022-08-10
Fix WEIGHT_EXTRAHEAVY on Qt6

### LINKED ISSUE #989: QPainter backend `WEIGHT_EXTRAHEAVY` fails on PySide6
It looks like Qt6 may have changed the implementation of weight constants, and no longer accepts integers, only `QFont.Weight` constants.

corranwebster:
Also some changes to tests to make sure all weights are tested correctly.

Benchmark render on PySide6:
image

Fixes #989.

### DIFF
--- a/kiva/qpainter.py
+++ b/kiva/qpainter.py
@@ -14,7 +14,7 @@ import numpy as np
 import warnings
 
 # Major package imports.
-from pyface.qt import QtCore, QtGui
+from pyface.qt import QtCore, QtGui, is_qt5
 
 # Local imports.
 from .abstract_graphics_context import AbstractGraphicsContext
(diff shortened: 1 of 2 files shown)


mdickinson (approved):
LGTM. I'm assuming it's intentional that `WEIGHT_HEAVY` and `WEIGHT_EXTRAHEAVY` map to the same thing with Qt 6?

corranwebster:
> I'm assuming it's intentional that `WEIGHT_HEAVY` and `WEIGHT_EXTRAHEAVY` map to the same thing with Qt 6?

Yes, there's nothing heavier than `QFont::Black` (weight 900) in Qt6.

PR #993, opened 2022-08-11
Replace uses of 0 and 1 with False and True, where appropriate.

corranwebster:
Note that in some cases there are more modern idioms that might be used, ~but fixing those is for another PR~ looking at the actual code, it might be worth the effort to make them more idiomatic (eg. `is_dashed` should be much clearer as a one-liner).


jwiggins (approved):
LGTM

PR #1001, opened 2022-08-15
Add long description content type

corranwebster:
This allows Twine to publish to PyPI without warning about long text content type.


mdickinson (approved):
LGTM. I've confirmed that a `python -m build --sdist` followed by `python -m twine check --strict dist/*` passes for me in a freshly-created Python 3.8 venv.

PR #1029, opened 2023-04-13
Remove OpenGL backend

### LINKED ISSUE #1010: Update Kiva for Pyglet 2
There are incompatible changes with Pyglet 2

corranwebster:
The OpenGL backend requires a lot of effort to maintain, is not complete and doesn't seem to be be being used in practice.

Fixes #1010


corranwebster:
> Missed a spot:

Searched for 'gl' and 'pyglet'; looks like I need to search for 'opengl' as well.

Also need to do basic fixes for tests (latest numpy in EDM isn't compatible with GitHub workers).

jwiggins:
Are we really doing this? 🥹

jwiggins (approved):

corranwebster:
It is done.

One day we might be able to do the same with the SWIG Agg backend.

PR #1033, opened 2023-04-17
MAINT: add compatibility with pyface 8.0.0

homosapien-lcy:
Recently, a lot of enthought open source packages such as enable, traitsui and pyface are changing the name of the qt folder name from qt4 to qt. This causes ModuleNotFoundErrors. Thus it would be necessary to add compatibility imports to handle these errors. This PR adds compatibility imports to solve the import error in enable/enable/examples/demo/enable/brush_draw.


corranwebster (approved):

corranwebster:
We need to turn off tests on EDM + MacOS - they will never succeed, and NumPy is not optional for Kiva.

PR #1034, opened 2023-04-17
Fix/update CI

corranwebster:
Currently this adds PR tests using pip on 3.8 and 3.10 and removes MacOS EDM tests which fail and will continue to fail due to test runner incompatibilities.

Eventually this should be expanded to run pip tests on Windows and Mac; and run on Python 3.11 as well.


corranwebster:
Closing in favour of #1036 - the pyproject.toml wrangling is a distraction in this PR

PR #1039, opened 2023-04-18
Add an action that builds wheels

corranwebster:
This will eventually be folded into the release action.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -453,45 +453,46 @@ if __name__ == "__main__":
     if sys.platform == 'darwin':
         ext_modules += macos_extensions()
 
-    setup(name='enable',
-          version=__version__,
-          author='Enthought, Inc',
-          author_email='info@enthought.com',
-          maintainer='ETS Developers',
-          maintainer_email='enthought-dev@enthought.com',
-          url='https://github.com/enthought/enable/',
-          # Note that this URL is only valid for tagged releases.
-          download_url=('https://github.com/enthought/enable/archive/'
-                        '{0}.tar.gz'.format(__version__)),
-          license='BSD',
-          classifiers=[c.strip() for c in """\
-              Development Status :: 5 - Production/Stable
-              Intended Audience :: Developers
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -429,26 +430,33 @@ def macos_extensions():
...
+                'fonttools/tests/data/*.ttf',
+                'fonttools/tests/data/*.txt',
+            ],
+            'kiva.examples': [
+                'kiva/*',
+                'kiva/*/*',
+            ],
+        },
+        zip_safe=False,
+        python_requires=">=3.7",
```
corranwebster, line 461: This is the only change here apart from formatting.

.github/workflows/publish-on-pypi.yml
```diff
@@ -6,17 +6,52 @@ on:
...
     - name: Check out the release commit
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - name: Set up Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
-        python-version: '3.8'
+        python-version: '3.11'
```
jwiggins, line 54: Any reason why the sdist should be created with the latest possible version of Python? I would assume an older version was better (and Cython source would be compiled by the user of the sdist)

corranwebster: No particular reason other than the assumption that most recent python would automatically get the most current packaging infrastructure.  Cython shouldn't be an issue because I don't think we ship the generated .cpp files in the sdist any more.

corranwebster: Looks like I was wrong about that: https://github.com/enthought/enable/blob/2230bb4487d292f91c724b34e5eb02abc193a60b/MANIFEST.in#L17

Probably should remove that - the isolated builds should work install Cython correctly from the sdist if needed.

And I also spot a stray gl reference in the manifest as well.

jwiggins (approved):
LGTM

PR #1041, opened 2023-04-18
Add explicit float -> int conversion for Qt

### LINKED ISSUE #1037: PyQt5 on Python 3.10 doesn't cast float to int
This is due to https://bugs.python.org/issue36048 (most likely).  We should probably be doing explicit casts to `int` in affected calls, since that was what was happening anyway.

corranwebster:
This was happening implicitly until https://bugs.python.org/issue36048 This makes it explicit, which means things should work with PyQt5 on Python 3.10.

Fixes #1037

### DIFF
--- a/.github/workflows/test-with-pip.yml
+++ b/.github/workflows/test-with-pip.yml
@@ -21,8 +21,6 @@ jobs:
             toolkit: 'wx'
           - os: 'macos-latest'
             toolkit: 'wx'
-          # PyQt5 API doesn't automatically cast float -> int, see #1037
-          - toolkit: 'pyqt5'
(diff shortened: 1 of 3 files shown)


jwiggins (approved):
Are these really the only call sites?

corranwebster:
> Are these really the only call sites?

Quite likely the only ones that have a mismatch between floats and ints.  The interface between the toolkit and Enable is fairly thin (here's a control, draw your stuff in it, listen for events; plus the native scrollbar stuff).

On an unrelated note, this seems to have turned the MacOS wx tests back on somehow...  I will fix before merging.

PR #1042, opened 2023-04-19
Python 3.11 support

corranwebster:
Run pip-based tests on 3.11.  We can't do Wx, PySide2 or PyQt5 easily since there are no wheels.


corranwebster:
To proceed with this we need either:
- a new celiagg release
- install from celiagg main

.github/workflows/bleeding-edge.yml
```diff
@@ -13,14 +13,13 @@ jobs:
...
-        toolkit: ['null', 'pyside6', 'wx']
-        python-version: ['3.10']
-        exclude:
-          # No Wx wheels available for ubuntu or macos
-          - os: 'ubuntu-latest'
-            toolkit: 'wx'
-          - os: 'macos-latest'
+        toolkit: ['null', 'pyside6']
+        python-version: ['3.11']
+        include:
```
ankit-gautam23, line 18: Why are we including this for this version? Is it needed?

ankit-gautam23 (changes requested):
If it can work with exclude let it be excluded unless there is some points that could be defended.

.github/workflows/bleeding-edge.yml
jwiggins: This ensures that Wx code is still tested by this workflow

corranwebster: @ankit-gautam23 This is a test suite which runs periodically and alerts Enthought devs if the tests fail due to upstream changes so we can take action.  We want some testing on WxPython, but it isn't easily installable for Python 3.11 yet.

corranwebster:
Tests should pass once https://github.com/celiagg/celiagg/pull/118 is merged

jwiggins (approved):

corranwebster:
Thanks!

PR #1043, opened 2023-04-20
Rename "qt4" to "qt"

corranwebster:
Rename `*.qt4.*` modules as `*.qt.*` and adapt imports to account for changes in Pyface 8.0 and TraitsUI 8.0 while keeping backwards compatibility.

It also removes some lingering PyQt4 support code.

This has some overlap with #1028 but doesn't fix everything that that PR does, so we still need that, or need to merge it into this.

As best I can tell, this change will break one test in an internal downstream project so we don't need to do any magic backwards compatibility mapping like we did in Pyface or TraitsUI.


enable/savage/trait_defs/qt/svg_button_editor.py
```diff
@@ -15,10 +15,13 @@
 # -----------------------------------------------------------------------------
 import os.path
 
+from pyface.qt import QtCore, QtGui
 from traits.api import Bool, Any, Str
-from traitsui.qt4.editor import Editor
+try:
+    from traitsui.qt.editor import Editor
+except ModuleNotFoundError:
+    from traitsui.qt4.editor import Editor
```
jwiggins, line 23: Doesn't this PR's change force us to use a `traitsui` which _doesn't_ have a `qt4` subpackage?

jwiggins (approved):

enable/savage/trait_defs/qt/svg_button_editor.py
corranwebster: No, it's currently backward compatible because it was low effort to do that (only 3-4 places needed try/excepts to achieve it) and it simplifies releasing.

PR #1051, opened 2023-05-05
Fix Quartz draw_path_at_points

corranwebster:
Also adds a test for all backends which implement the method.

This was causing failures in Chaco scatterplots with the Quartz backend.


jwiggins (approved):
LGTM

PR #1052, opened 2023-05-09
Test setting colors, and fixes based on tests.

corranwebster:
Test setting colors, and fixes based on tests.

This addresses some issues seen with the segment plot in Chaco.


dpinte (approved):
👍

Issue #1030, opened 2023-04-17
TypeError: unsupported operand type(s) for &: 'int' and 'KeyboardModifier'


homosapien-lcy:
**Problem Description**
When running many enable examples such as enable/enable/examples/demo/enable/basic_draw.py, the example will run successfully at first but the plot will disappear after clicking other parts of the screen. Then the following error message will show up in the terminal after closing the plot window:

```
(py311) (base) cyliu@aus552cyliu 3.11_test % python3.11 enable/enable/examples/demo/enable/basic_draw.py
2023-04-17 12:46:55.990 Python[5078:122981] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to /var/folders/2z/kylzj9s92y71cxscmljmpqrh0000gt/T/org.python.python.savedState
Traceback (most recent call last):
  File "/Users/cyliu/Documents/3.11_test/enable/enable/examples/demo/enable/basic_draw.py", line 42, in <module>
    demo = demo_main(Demo)
           ^^^^^^^^^^^^^^^
  File "/Users/cyliu/Documents/3.11_test/enable/enable/examples/_example_support.py", line 40, in demo_main
    demo_class().configure_traits()
  File "/Users/cyliu/Documents/3.11_test/traits/traits/has_traits.py", line 2163, in configure_traits
    rc = toolkit().view_application(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyliu/.venvs/py311/lib/python3.11/site-packages/traitsui/qt4/toolkit.py", line 237, in view_application
    return view_application.view_application(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyliu/.venvs/py311/lib/python3.11/site-packages/traitsui/qt4/view_application.py", line 92, in view_application
    return ViewApplication(
           ^^^^^^^^^^^^^^^^
  File "/Users/cyliu/.venvs/py311/lib/python3.11/site-packages/traitsui/qt4/view_application.py", line 138, in __init__
    start_event_loop_qt4()
  File "/Users/cyliu/.venvs/py311/lib/python3.11/site-packages/pyface/util/guisupport.py", line 156, in start_event_loop_qt4
    app.exec()
  File "/Users/cyliu/.venvs/py311/lib/python3.11/site-packages/traitsui/qt4/toolkit.py", line 129, in eventFilter
    if event.type() == QtCore.QEvent.Type.KeyPress:
[...]

PR #1053, opened 2023-05-09
Clean-up of Enable Qt toolkit code

corranwebster:
This PR:
- removes some remaining OpenGL code that is no longer used
- cleans up the logic around the creation of mouse events and drag events

This is a follow-up to some of the issues identified in #1030


dpinte (approved):
👍

Issue #1060, opened 2023-07-21
Cython 3.0 Breaks Kiva Build


corranwebster:
We are seeing C++ errors in the Kiva build - incorrect wrapper code is somehow being generated.  There look to be multiple different errors, so it may take a little time to unwind.

In the mean time, we may want to update the build requirements to specify earlier Cython.

drbsg:
This is affecting v5.3.1 as well on platforms which do not have binary wheels available (in particular Windows). I would very much appreciate a back-port of a fix for this.

corranwebster:
The work-around in the short-term is to specify "cython<3" in the pyproject.toml before building.  I'm not sure we have bandwidth to backport fixes in the near-term, but I will happily review PRs.

itziakos:
fixes by #1070

PR #1061, opened 2023-07-21
Insist on Cython prior to 3.0

corranwebster:
Hopefully this is a quick-fix for #1060 - but a more permanent fix will be needed at some future point.

Now includes a work-around for #1062 (hopefully)


jwiggins (approved):
I'm 👍 once CI is happy

PR #1063, opened 2023-08-08
Insist on Cython prior to 3.0

drbsg:
This is a back-port of pull request #1061 to the 5.3 maintenance branch.

### DIFF
--- a/enable/__init__.py
+++ b/enable/__init__.py
@@ -16,7 +16,7 @@ except ImportError:
     __version__ = "not-built"
 
 __requires__ = [
-    "numpy", "pillow", "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "fonttools"
+    "numpy", "pillow<10", "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "fonttools"
 ]
 
 __extras_require__ = {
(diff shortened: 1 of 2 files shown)


enable/__init__.py
```diff
@@ -16,7 +16,7 @@
     __version__ = "not-built"
 
 __requires__ = [
-    "numpy", "pillow", "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "fonttools"
```
jwiggins, line 19: Where did this change come from?

jwiggins (approved):
LGTM

@corranwebster Any objections? (you're a bit closer to this than me)

corranwebster:
PR looks good, not sure what's happening with CI

jwiggins:
Looks like the Wx environment needs `attrdict` installed before installing Wx

drbsg:
> Looks like the Wx environment needs `attrdict` installed before installing Wx

I am not familiar with EDM, so would adding this dependency to `dependencies` for both versions in `ci/edmtool.py` be a good way to achieve this?

corranwebster:
I think the right way to fix would either be to restrict the WxPython version or to add `attrdict` here: https://github.com/enthought/enable/blob/84b611dcc329c7558f5512b861fb8e002f90f39f/enable/__init__.py#L52

The problem seems to be this version of WxPython not declaring all dependencies.

itziakos:
#1070 updated code to work with Cython > 3.

